Issue #130, opened 2011-03-02
Feature Request: data_seek support


kyledrake:
Check out this code: https://github.com/oldmoe/mysqlplus/blob/master/ext/mysql.c#L1363

This is really useful for picking out random results on -huge- queries, and helps to get around the RAND() slowdown problems. You could implement a result[rand(result.count)] and not have to pull down an entire pile of records into ruby.

I'm working on a piece of code to allow people to utilize this easily using mysqlplus.. but I'd love to be able to do it in mysql2 instead :-)

sodabrew:
Revisiting this with the current architecture -- streaming (`:stream`) and prepared statements didn't exist in mysql2 when this was filed in 2011, and both change what's actually safe here. Confirmed feasible; here's the design.

## Feasibility, confirmed against the current codebase

Both C primitives exist and cover both protocols:
- `mysql_data_seek(MYSQL_RES*, offset)` -- plain queries (text protocol).
- `mysql_stmt_data_seek(MYSQL_STMT*, offset)` -- prepared statements (binary protocol).

Both reposition a pointer into an **already fully downloaded, client-side row buffer** -- pure memory arithmetic, no network I/O. This codebase already documents exactly that distinction for the existing row-fetch code (`result.c`, `nogvl_fetch_row`'s comment): *"Only a streaming result can hit the network from a row fetch. A non-streaming result has already been drained into client memory... fetching a row from it is pointer arithmetic over that buffer and cannot block."* A seek-based accessor on a buffered result inherits that exact property: **GVL-holding, no `rb_thread_call_without_gvl` needed**, unlike almost everything else in this file.

## Why streaming is a hard "no", not just "slow"

`:stream => true` never downloads the full result client-side at all:
- Plain queries: `mysql_use_result()`, which pulls rows from the wire one at a time as `#each` asks for them.
- Prepared statements: a server-side cursor (`mysql_stmt_attr_set(STMT_ATTR_CURSOR_TYPE, CURSOR_TYPE_READ_ONLY)`, `statement.c:699-703`), same idea, different mechanism.

Neither has a client-side row array `data_seek` could reposition within -- there's nothing to seek into. This has to raise a clear `Mysql2::Error` ("can't random-access a streaming result"), not attempt a seek and get undefined behavior from the C API. No amount of engineering makes this work for streaming; it's a permanent, documented restriction, not a bug to fix later.

## The real hazard: corrupting sequential iteration state

[...]

PR #345, opened 2012-12-30
Add element reference operator #[] to Mysql2::Result using mysql_data_seek

sodabrew:
For #130 mysql_data_seek is wrapped behind the element reference operator #[]. You can jump directly to a particular result row:

``` ruby
  result = @client.query "SELECT 1 UNION SELECT 2"
  p result[1]

  {"1" => 2}
```

Open for comment!

### DIFF
--- a/ext/mysql2/result.c
+++ b/ext/mysql2/result.c
@@ -66,6 +66,8 @@ typedef struct {
   VALUE block_given;
 } result_each_args;
 
+typedef VALUE (*fetch_row_func_t)(VALUE, MYSQL_FIELD *fields, const result_each_args *args);
+
 VALUE cBigDecimal, cDateTime, cDate;
 static VALUE cMysql2Result;
 static VALUE opt_decimal_zero, opt_float_zero, opt_time_year, opt_time_month, opt_utc_offset;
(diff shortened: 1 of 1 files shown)


sodabrew:
Two years later - rebased yo!

ext/mysql2/result.c
```diff
@@ -440,14 +444,121 @@ static VALUE rb_mysql_result_fetch_fields(VALUE self) {
   return wrapper->fields;
 }
 
+static void rb_mysql_row_query_options(VALUE opts, ID *db_timezone, ID *app_timezone, int *symbolizeKeys, int *asArray, int *castBool, int *cast, int *cacheRows) {
+  ID dbTz, appTz;
```
tamird: underindented by one?

ext/mysql2/result.c
```diff
@@ -755,6 +755,134 @@ static VALUE rb_mysql_result_fetch_fields(VALUE self) {
...
+
+  c_seek = NUM2LONG(seek);
+  if (!NIL_P(count)) {
+    c_count = NUM2LONG(count);
+    /* Special case: ary[x, 0] returns []*/
+    if (!c_count) return rb_ary_new();
+  }
+
+  if (!NIL_P(opts)) {
+    opts = rb_funcall(defaults, intern_merge, 1, opts);
```
tamird: this preamble could be implemented in ruby and would be much easier to read - why not do that?

sodabrew: I actually moved this further down the chain so that the same code is used by the variants of [] and each.

PR #403, opened 2013-07-17
Convert specs from "should" to "assert" syntax

brianmario:
This may seem like a useless waste of time, but the "should" syntax is deprecated anyway and I personally prefer using the assert syntax anyway. This also means it'll be a lot easier to move to minitest in the future if we want to.

Don't worry, I still love rspec's runner WAY more so we'll be sticking with it for at least a little while longer ;)

No huge rush on merging this, but it will get harder and harder to do if this branch and master get diverged too much. Especially if tests are being updated often in master.

### DIFF
--- a/spec/em/em_spec.rb
+++ b/spec/em/em_spec.rb
@@ -24,8 +24,8 @@ begin
         end
       end
 
-      assert_include results[0].keys, "second_query"
-      assert_include results[1].keys, "first_query"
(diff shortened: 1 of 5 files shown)


onebree:
@brianmario is there a difference between assert syntax and expect syntax? Throughout our app, we began using `expect()` when we upgraded to Rails 4.

sodabrew:
should -> assert migration completed in #595.

Issue #469, opened 2013-12-27
Detect and warn if libmysql.dll requires gendef.exe fix


sodabrew:
The failures seen on 64-bit Windows with 64-bit Ruby and 64-bit MySQL all appear to be due to `libmysql.dll` that is missing some form of linkage data.

What I'd like to do is find a way to detect and warn the user if their `libmysql.dll` needs to be regenerated with the additional linkage information. Compile time detection would be good, runtime detection would be great.

A fix is described at: https://bugs.ruby-lang.org/issues/8591

This supersedes our issues: #372, #430 (add more to this list as needed).

PR #473, opened 2013-12-29
Create libmysql.a using dlltool.exe on mingw64

sodabrew:
~~I have no idea if this will work.~~ It works! See #469.


sodabrew:
Currently blocked on rake-compiler/rake-compiler#100.

lib/mysql2.rb
```diff
@@ -3,6 +3,26 @@
 require 'bigdecimal'
 require 'rational' unless RUBY_VERSION >= '1.9.2'
 
+# Load libmysql.dll before requiring mysql2/mysql2.so
+# This gives a chance to be flexible about the load path
+# Or to bomb out with a clear error message instead of a linker crash
+if RUBY_PLATFORM =~ /mswin|mingw/
+  dll_search = [
+    File.expand_path('../vendor/libmysql.dll', File.dirname(__FILE__)).gsub('/', '\\'),
+    File.expand_path('../libmysql.dll', File.dirname(__FILE__)).gsub('/', '\\'),
```
sodabrew: Update: the fat binary mechanism is for multiple Ruby versions for the same platform, which means they would share the same `vendor/libmysql.dll` binary. I should still remove the second search location, though.

sodabrew:
At long last, I believe this is done. Anybody following the thread, please review code and test it out!

sodabrew:
In "Wrap vendor:mysql tasks with a platform argument" each of the platform Connector-C zip files are fetched and unzipped in `rake-compiler-dev-box`.

sodabrew:
I released mysql2 version 0.3.18 **with windows binaries** today!
If you find a problem with the 0.3.18 version, please open a new ticket.

PR #479, opened 2014-01-16
Refactor branching to assignment

juanxo:
Currently, the function `rb_mysql_result_each` in `result.c` sets some default values on C vars and change those values depending on the `options` ruby hash.

This PR removes the default C values and the branching with some simple assignments. This works because:

- The C vars default value are kept in sync with the `@query_options` ruby hash, and that hash is merged with the supplied options, so the merged hash has the same values than the default C ones, unless they are overriden.
- The code doesn't check for the actual values of the variables (1 or 0). It only bothers about their truthiness.

This improves a couple of minor things in my opinion:
- Changing an existing default value will only require changes in the ruby `@@default_query_options` attribute
- It removes N branches(one for each option). While this wont improve performance, it improves the readability a little bit.

### DIFF
--- a/ext/mysql2/result.c
+++ b/ext/mysql2/result.c
@@ -431,7 +431,7 @@ static VALUE rb_mysql_result_each(int argc, VALUE * argv, VALUE self) {
   mysql2_result_wrapper * wrapper;
   unsigned long i;
   const char * errstr;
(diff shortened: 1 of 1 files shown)


sodabrew:
Cherry-picked into #581

Issue #493, opened 2014-03-04
default_file option is ignored if added to Client.default_query_options


fcrimins:
This can be fixed by using the '@query_options' hash rather than the 'opts' hash in the options/symbols loop following 'initialize_ext' in Client#initialize.

Presumably this problem exists for other options if they are specified in Client.default_query_options as well.

sodabrew:
Confirmed: :default_file/:default_group are read once from the opts hash passed to Client.new (never from @query_options), then handed to libmysqlclient at connect time. There's no live value to read back afterward -- changing default_file after connecting can't have any effect regardless of where it's set, the same way switching between whole parcels of connection flags mid-connection just isn't something a live connection can do.

The actual fix: PR #1583 (originally for the same confusion with :database, #437) now also drops :default_file/:default_group from query_options after connecting, so Client.default_query_options.merge!(default_file: ...) no longer looks like it's configuring anything -- it just isn't there to find. The right way to reuse connection options across multiple clients remains what @sodabrew said in 2014: keep your own hash of options and pass it to each Client.new.

sodabrew:
Reopening — leaving open until PR #1583 actually merges, consistent with how #437/#507 (the same underlying fix) are being handled.

Issue #437, opened 2013-10-31
Changing database in query options hash has no effect


gstaubli:
Table:

```
prod.final_product
```

Re-creating bug:

``` ruby
client = Mysql2::Client.new(
{
  :host => "127.0.0.1", 
  :username => "username", 
  :password => "password", 
  :database => ""
}) # => #<Mysql2::Client ... :database => ""}, @read_timeout=nil>
client.query("SELECT * FROM final_product") # => Mysql2::Error: No database selected
client.query_options[:database] = "prod" # => "prod"
client # => #<Mysql2::Client ... @query_options = { ... :database => "prod"}, @read_timeout=nil>
client.query("SELECT * FROM final_product") # => Mysql2::Error: No database selected
```
### Using #select_db

``` ruby
client = Mysql2::Client.new(
{
  :host => "127.0.0.1", 
  :username => "username", 
  :password => "password", 
  :database => ""
}) # => #<Mysql2::Client ... :database => ""}, @read_timeout=nil>
client.select_db("prod") # => "prod"
client.query("SELECT * FROM final_product") # => #<Mysql2::Result ... :database => "prod"}>
```
### Using options hash on client initialization

``` ruby
client = Mysql2::Client.new(
{
  :host => "127.0.0.1", 
  :username => "username", 
  :password => "password", 
  :database => "prod"
}) # => #<Mysql2::Client ... :database => ""}, @read_timeout=nil>
client.query("SELECT * FROM final_product") # => #<Mysql2::Result ... :database => "prod"}>
```

PR #507, opened 2014-04-10
Separate out connect options from query options

sodabrew:
An idea to resolve the confusing situation seen in #493 and a few previous tickets.

First pass code has the bad effect of overriding the user's settings with defaults if the user changes the class hash values in `@@default_query_options` before connecting, but has not updated to know about `@@default_connect_options` introduced here.

Retains the bad situation of having both sets of options in the `@query_options` hash from there on out.

Needs some more thinking, putting this out for comment and ideas.


sodabrew:
Picking this up now for 0.4.0. It will also clarify the situation described in #437.

lib/mysql2/client.rb
```diff
@@ -1,71 +1,69 @@
 module Mysql2
   class Client
-    attr_reader :query_options, :read_timeout
+    attr_reader :connect_optins, :query_options, :read_timeout
```
sodabrew: typo

PR #563, opened 2014-11-22
Do not cast decimal to integer

### LINKED ISSUE #557: Column created as :decimal being treated as :integer on mysql with default: 0.0
Version information:
MYSQL DB version: 5.5.38-0ubuntu0.14.04.1
Rails version: 4.1.0
Ruby version: 2.1.0p0 (2013-12-25 revision 44422) [x86_64-linux]

I created a table with the following migration: 
create_table "my_table", force: true do |t|
t.decimal "my_column", default: 0.0
end

When attempting to assign values to the column, the values would be treated as integers. In the MYSQL console, the column was reported as DECIMAL(10,0) with a default of 0. In the rails console, the column was reported as type :integer and acted as one.

No idea what caused this, but setting precision: 10, scale: 2 caused rails to treat the column as a :decimal again.

gmile:
This reverts a099e55b678cec463b479c3589cb4dcfdcc837fc, where regression has been introduced.

Fixes #557.

### DIFF
--- a/ext/mysql2/result.c
+++ b/ext/mysql2/result.c
@@ -267,9 +267,7 @@ static VALUE rb_mysql_result_fetch_row(VALUE self, ID db_timezone, ID app_timezo
           break;
         case MYSQL_TYPE_DECIMAL:    /* DECIMAL or NUMERIC field */
         case MYSQL_TYPE_NEWDECIMAL: /* Precision math DECIMAL or NUMERIC field (MySQL 5.0.3 and up) */
-          if (fields[i].decimals == 0) {
-            val = rb_cstr2inum(row[i], 10);
-          } else if (strtod(row[i], NULL) == 0.000000){
+          if (strtod(row[i], NULL) == 0.000000){
             val = rb_funcall(cBigDecimal, intern_new, 1, opt_decimal_zero);
           }else{
             val = rb_funcall(cBigDecimal, intern_new, 1, rb_str_new(row[i], fieldLengths[i]));


sodabrew:
I'm not really sure how to proceed here. I don't have a good sense if people are now expecting the >=0.3.11 behavior, or if it's reasonable to choose the behavior as a query_option.

sodabrew:
Closing as no longer applicable: the behavior this reverts is now intentional, tested, current design, not an unreverted regression.

This PR reverts a099e55, which made `DECIMAL`/`NEWDECIMAL` columns with `decimals == 0` (e.g. `DECIMAL(10,0)`) cast to Ruby `Integer` instead of `BigDecimal`. That's still exactly what current master does:

https://github.com/brianmario/mysql2/blob/466a9ae0fbc81fbda1a6ac13c377faa0e8c94774/ext/mysql2/result.c#L1203-L1212

It's no longer an accidental leftover -- it's since been reinforced twice over:

- `mysql2_cast_integer` was built out specifically to handle this path safely for values beyond 64-bit range: https://github.com/brianmario/mysql2/blob/466a9ae0fbc81fbda1a6ac13c377faa0e8c94774/ext/mysql2/result.c#L710-L760
- There's a dedicated, currently-passing spec asserting this exact behavior for values that exceed 64-bit range: https://github.com/brianmario/mysql2/blob/466a9ae0fbc81fbda1a6ac13c377faa0e8c94774/spec/mysql2/result_spec.rb#L764-L770

This also lines up with the root cause discussed on #557: Rails' `ActiveRecord::Type` maps `decimal(p,0)` columns to `:integer` by convention, so mysql2 casting them to `Integer` matches Rails' own type system rather than fighting it. What looked like a regression in 2014 is the behavior mysql2 has since deliberately kept and built on.

Merging this now would revert that and break the DECIMAL(65,0) spec linked above.

PR #574, opened 2015-01-20
Update README with MacPorts + OS X installation

ain:


sodabrew:
`/opt/local/lib/mysql5` is in the default search path, is that no longer the installation path in macports?
https://github.com/brianmario/mysql2/blob/master/ext/mysql2/extconf.rb#L23

ain:
`mysql5` is 5.1, `mysql56` is 5.6, the latest on MacPorts.

I'd actually suggest extending the the default search to `~5.5`.

ain:
FMI please see [mysql5* packages on MacPorts](https://www.macports.org/ports.php?by=name&substr=mysql5).

sodabrew:
I also use MacPorts. Just noticed that I symlinked `/opt/local/lib/mysql5` to `/opt/local/mysql56` on my system. Derp.

It's a little bit annoying to chase version paths, but that's probably the best approach for now. I'd like to simplify your README update a bit, and/or split the Linux and OS X sections. I'll get back to this in the evening with some suggested text.

sodabrew:
... or a couple weeks later. Let me know if #577 looks reasonable to you in lieu of this PR.

ain:
All good. I think we've covered this topic and can close this PR in favour of #577.

PR #578, opened 2015-02-07
Only invoke vendor:mysql while cross-compiling for Windows

sodabrew:
Correct a mistake in #473, `vendor:mysql` is getting called and downloading Connector/C-{win32, -winx64}  on non-Windows platforms.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -7,7 +7,7 @@ install:
   - gem --version
   - gem install bundler --quiet --no-ri --no-rdoc
   - bundler --version
-  - bundle install --without benchmarks
+  - bundle install --without benchmarks --path vendor/bundle
 build_script:
   - bundle exec rake compile
 test_script:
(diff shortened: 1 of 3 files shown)


brianmario:
Don't we still need the win32/64 connector-c for cross-compilation (for headers and linking) on non-windows platforms?

sodabrew:
Yeah, this doesn't execute early enough. I'm working on an update.

sodabrew:
Something is way off now in the cross-compile job. It's passing the arguments for multiple variants to extconf.rb as it cycles through each cross build target. The obvious failure case is with the wrong bitness of Connector/C used by half of the variants. Need to dig down for a root cause.

sodabrew:
Also waiting on the AppVeyor build: https://ci.appveyor.com/project/sodabrew/mysql2/build/149

sodabrew:
And the final issue, which was `i386` vs. `x86` in the machine specs just got answered at rake-compiler/rake-compiler#105 -- `x86` is good.

PR #579, opened 2015-02-10
Fix msec is not passed

kamipo:


sodabrew:
Ouch good catch!

amatsuda:
:metal:

PR #509, opened 2014-04-16
Rebase of #489 "fix error in EM::Client::close"

sodabrew:
Rebase of #489 "fix error in EM::Client::close" by @mrms-dos.

PR #489, opened 2014-02-09
RuntimeError "invalid binding to detach" in EM::Client::close (2)

dmgtn:
Error happens when you close EM::Client with no query running.

```
RuntimeError invalid binding to detach
.../ruby-2.0.0-p353/gems/eventmachine-1.0.3/lib/em/connection.rb:272:in `detach_fd'
.../ruby-2.0.0-p353/gems/eventmachine-1.0.3/lib/em/connection.rb:272:in `detach'
.../ruby-2.0.0-p353/gems/mysql2-0.3.14/lib/mysql2/em.rb:29:in `close'
```

PR #580, opened 2015-02-10
Reimplement #509, add self-explaining specs

marshall-lee:
Related  to #489 and #509 

I reproduced the problem and added specs. It reproduces when trying to close a connection *outside* of deferred callbacks.

I also found similar workaround in ruby-em-pg-client: https://github.com/royaltm/ruby-em-pg-client/blob/v0.3.4/lib/pg/em.rb#L500 - It also detaches a watch only if it can.


sodabrew:
AWESOME!

Issue #406, opened 2013-07-19
Mysql2::Result#count always returns 0 when using stored procedures and MULTI_STATEMENTS


key88sf:
Gem versions after 0.3.11 all have this issue.

    db_config = Rails.application.config.database_configuration["development"].symbolize_keys
    client = Mysql2::Client.new(db_config.merge(:flags => Mysql2::Client::MULTI_STATEMENTS, :cache_rows => false))
    sql = "CALL my_proc();" # returns several rows
    result = client.query(sql, :stream => true)
    result.count # 0 (but should be non-zero)

sodabrew:
Ok, I can reproduce this. It appears to be an incompatibility between `:stream => true` and `:flags => Mysql2::Client::MULTI_STATEMENTS`.

sodabrew:
Reading through http://dev.mysql.com/doc/refman/5.0/en/mysql-use-result.html and looking at the code a bit, I am pretty sure we _don't_ correctly handle multi statements streaming correctly. I may need to kick this out by a milestone in order to put some time into it.

sodabrew:
@brianmario Hey, this looks weird. Off the cuff, would you expect that trying to iterate the rest of the results would kill the count?

``` ruby
        @multi_client = Mysql2::Client.new(DatabaseCredentials['root'].merge(:flags => Mysql2::Client::MULTI_STATEMENTS))
        result = @multi_client.query("SELECT 1 AS 'set_1'; SELECT 2 AS 'set_2'", :stream => true, :cache_rows => false)
        result.first.should eql({ 'set_1' => 1 })
        @multi_client.more_results?.should be_true

        result.count.should == 1
        result.each { |r| } # consume the end-of-results packet
        result.count.should == 1
```

Fails with (writing this as part of a new unit test)
```
  1) Mysql2::Client#query Multiple results sets should stream multiple result sets
     Failure/Error: result.count.should == 1
       expected: 1
            got: 0 (using ==)
     # ./spec/mysql2/client_spec.rb:640:in `block (4 levels) in <top (required)>'

```

sodabrew:
Merged #581, will be released in 0.4.0.

PR #581, opened 2015-02-14
Tests for streaming and stored procedures with multiple result sets

sodabrew:
Specs for #406. Generally the behavior there is correct and I do not think the OP's issue stands, however there is a weird thing I'd like to understand better:

``` ruby
        @multi_client = Mysql2::Client.new(DatabaseCredentials['root'].merge(:flags => Mysql2::Client::MULTI_STATEMENTS))
        result = @multi_client.query("SELECT 1 AS 'set_1'; SELECT 2 AS 'set_2'", :stream => true, :cache_rows => false)
        result.first.should eql({ 'set_1' => 1 })
        @multi_client.more_results?.should be_true

        result.count.should == 1
        result.each { |r| } # consume the end-of-results packet
        result.count.should == 1
```

Fails with (writing this as part of a new unit test) (it's the second result.count that's now 0)
```
  1) Mysql2::Client#query Multiple results sets should stream multiple result sets
     Failure/Error: result.count.should == 1
       expected: 1
            got: 0 (using ==)
     # ./spec/mysql2/client_spec.rb:640:in `block (4 levels) in <top (required)>'

```

### DIFF
--- a/ext/mysql2/result.c
+++ b/ext/mysql2/result.c
@@ -446,7 +446,7 @@ static VALUE rb_mysql_result_each(int argc, VALUE * argv, VALUE self) {
   mysql2_result_wrapper * wrapper;
   unsigned long i;
   const char * errstr;
-  int symbolizeKeys = 0, asArray = 0, castBool = 0, cacheRows = 1, cast = 1, streaming = 0;
+  int symbolizeKeys = 0, asArray = 0, castBool = 0, cacheRows = 1, cast = 1;
   MYSQL_FIELD * fields = NULL;
 
   GetMysql2Result(self, wrapper);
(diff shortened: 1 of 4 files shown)


sodabrew:
Ok, this is funky. Turns out if you call `results.count` before it has finished streaming, it will calls `mysql_num_rows`, but if you call `results.count` after streaming it finished, it will use `wrapper->numberOfRows`. 

It looks like this goes back to #273. I think the key fix in this PR is this:

``` diff
@@ -532,16 +528,15 @@ static VALUE rb_mysql_result_each(int argc, VALUE * argv, VALUE self) {

       do {
         row = rb_mysql_result_fetch_row(self, db_timezone, app_timezone, symbolizeKeys, asArray, castBool, cast, fields);
-
-        if (block != Qnil && row != Qnil) {
-          rb_yield(row);
-          wrapper->lastRowProcessed++;
+        if (row != Qnil) {
+          wrapper->numberOfRows++;
+          if (block != Qnil) {
+            rb_yield(row);
+          }
         }
       } while(row != Qnil);

-
-      wrapper->numberOfRows = wrapper->lastRowProcessed;
       wrapper->streamingComplete = 1;

       // Check for errors, the connection might have gone out from under us
```

sodabrew:
Moving this to 0.4.0, because it keeps looking like a refactor of longstanding code no matter how much I try to trim down the diff. The current behavior is definitely incorrect, I just can't zoom in on a concise changeset yet.

PR #586, opened 2015-02-21
Incomplete attempt to vary libmysql.dll per platform target

sodabrew:
See also rake-compiler/rake-compiler#106.


sodabrew:
The 0.3.19 windows gems were built with rake-compiler-dock - many thanks to @larskanis for [rake-compiler-dock](https://github.com/rake-compiler-dock/rake-compiler-dock) and for #637 (see 99c57e9). I updated it with d8bb4d1, working around the issue of the x86 vendor/libmysql.dll ending up in the x64 builds by simply calling `RakeCompilerDock.sh` twice and cleaning in between.

PR #558, opened 2014-11-09
[clean] prepared statements

justincase:
The master branch has diverged quite a bit since the original prepared statement commits by tenderlove, brianmario, and nyaxt. I've gone through most of the original commits and wrangled them to fit a clean branch. Tests pass with each commit.

APIs featured:

`Mysql2::Client#prepare(query) # => Mysql2::Statement`
`Mysql2::Statement#execute(*params) # => Mysql2::Statement`
`Mysql2::Statement#param_count # => Integer`
`Mysql2::Statement#field_count # => Integer`
`Mysql2::Statement#fields # => Array[String]`

This PR includes commits from nyaxt up to 50a7054299e08af1356ffae80fda13301a5a882f. Most notably, his work on returning `Mysql2::Result` from `Mysql2::Statement#execute` is missing.

This PR supersedes others and gives us a clean slate to work from.

PR #591, opened 2015-03-06
Prepared statements (rebase of #558)

justincase:
This is a rebase of https://github.com/brianmario/mysql2/pull/558 against the current master (5e21bf2).


sodabrew:
wow! and all tests pass! I will review asap.

justincase:
Many thanks, Aaron. There's quite a bit of code that could be refactored:

* deeply nested if and switch statements 
* args passing from `rb_mysql_result_each`  [result.c#L1008](https://github.com/justincase/mysql2/blob/8a2929c191b9c78d8982a6264a42f0ab2227ae23/ext/mysql2/result.c#L1008)
* `rb_mysql_result_each_nonstmt` and `rb_mysql_result_each_stmt ` are nearly identical

It works for now though.

ak47:
Hi there, just so happens this PR solves 100% of problems for my current task

Any idea if this PR will make it into the gem?

It would lead to much rejoicing.

thank you!

sodabrew:
Was just going through this for review, yes the `rb_mysql_result_each_stmt` and `rb_mysql_result_each_nonstmt` versions jumped out at me, and so did `rb_mysql_result_stmt_fetch_row` and `rb_mysql_result_fetch_row`. I will take a stab at factoring this out a bit.

sodabrew:
I just pushed a melding of rb_mysql_result_each_stmt and rb_mysql_result_each_nonstmt to a copy of this branch in my tree, would you review and pull the top commits from https://github.com/sodabrew/mysql2/tree/pr/591 into this PR?

ext/mysql2/result.c
```diff
@@ -191,6 +217,283 @@ static unsigned int msec_char_to_uint(char *msec_char, size_t len)
...
+      case MYSQL_TYPE_VARCHAR:      // char[]
+      case MYSQL_TYPE_TINY_BLOB:    // char[]
+      case MYSQL_TYPE_BLOB:         // char[]
+      case MYSQL_TYPE_MEDIUM_BLOB:  // char[]
+      case MYSQL_TYPE_LONG_BLOB:    // char[]
+      case MYSQL_TYPE_BIT:          // char[]
+      case MYSQL_TYPE_SET:          // char[]
+      case MYSQL_TYPE_ENUM:         // char[]
+      case MYSQL_TYPE_GEOMETRY:     // char[]
+        wrapper->result_buffers[i].buffer = malloc(fields[i].max_length);
```
sodabrew: Should be `xmalloc` or `xcalloc` for consistency?

ext/mysql2/statement.c
```diff
@@ -0,0 +1,442 @@
...
+      params_enc[i] = Qnil;
+
+      switch (TYPE(argv[i])) {
+        case T_NIL:
+          bind_buffers[i].buffer_type = MYSQL_TYPE_NULL;
+          break;
+        case T_FIXNUM:
+#if SIZEOF_INT < SIZEOF_LONG
+          bind_buffers[i].buffer_type = MYSQL_TYPE_LONGLONG;
+          bind_buffers[i].buffer = malloc(sizeof(long long int));
```
sodabrew: These are all `malloc`, too, but in result.c it's all `xmalloc`/`xcalloc`.

sodabrew: Actually this was explicitly chosen in 05163af780b3a0841020c58df52d6d937a9decb4.

ext/mysql2/result.c
sodabrew: Switched to xmalloc in a289066 for consistency, and to use xfree for all result_buffers to match.

justincase:
For future reference; why did we decide on xmalloc/xcalloc and what's the difference between those?

sodabrew:
Thanks for pulling my commits over! I'll do another read-through this week. It's looking good!

Regular malloc/calloc will grab memory from the system. xmalloc/xcalloc will get memory from Ruby's heap, which means it can be tracked from the GC system, and can also trigger GC thresholds (i.e. do a GC run before returning the allocation).

evanphx:
@sodabrew That is actually not true. `xmalloc` simply lets the ruby GC know that some C heap memory was allocated. This goes into it's calculate of when to run the GC, allowing the GC to run when there was only a small number of ruby objects created but a large amount of heap memory allocated, allowing `free` functions on the ruby objects to run and hopefully free the C heap memory.

sodabrew:
@justincase Here's another one for memory usage: c02cb6c. By the way, you can pull (fetch + merge --ff) the commits from my branch to your branch, rather than cherry-pick, and retain the same commit hashes, since my branch has the same history as your branch.

@evanphx I found my way to http://rxr.whitequark.org/mri/source/gc.c#4685 and then it got hairy! On the upside, it looks like using xmalloc means you get to use Ruby's memory information tools, which may be helpful for debugging memory leaks.

 I guess the bottom line is whether it is current best practice for extensions to use Ruby's xmalloc/xfree or not? Many articles on the topic are a few years old and reference the 1.8.x memory system...

evanphx:
@sodabrew Using `xmalloc` is generally recommended so that the ruby GC can run and call your free functions properly.

ext/mysql2/statement.c
sodabrew: Per https://github.com/brianmario/mysql2/pull/591#issuecomment-91651666 let's switch this to xmalloc unless @brianmario says he switched it for a strong reason?

justincase:
@sodabrew I merged the commits. However, I can't find c02cb6cf8f2fb73f7d3ef875b0e499db8eb44150 on any branch. Shall I apply it as a patch?

sodabrew:
I replaced c02cb6c with 2dbbcdc. I'll check if you've got them all.

sodabrew:
@rdp This is the current PR for prepared statements support. I have reviewed it and am working on a merge branch on my desktop. Sorry about going radio silent!

justincase:
@sodabrew Apologies for the delay, I had not noticed your edit. 

I'll attempt to change the dates back to their original state over the weekend. There are quite a few [suggestions on stackoverflow](http://stackoverflow.com/questions/454734/how-can-one-change-the-timestamp-of-an-old-commit-in-git) with regards to filter-branch.

sodabrew:
No objections to new SHAs, this branch has been through its share of rebases already :)

justincase:
@sodabrew The author dates have been restored. I did a rebase against current master as well.

sodabrew:
I've branched 0.3.x at https://github.com/brianmario/mysql2/tree/0.3.x and now we can take master down the rabbit hole.

justincase:
Thank you @tenderlove, @nyaxt, @sodabrew, @brianmario, @johncant. 

Tenderlove's first commit was almost 5 years ago and now we're close to a released version. 

Cheers.

PR #593, opened 2015-03-11
Friendly ext build error <3

kirs:
I always bundle this gem on my production servers, then it fails and I'm googling which Deian package should I install. After I find out it's `libmysqlclient-dev`, I can finally install it.

I think that suggesting this package right in the output will save few minutes for every Ruby developer.
The same [trick is used in sqlite3-ruby](https://github.com/sparklemotion/sqlite3-ruby/blob/master/ext/sqlite3/extconf.rb#L22)

:heart:

### DIFF
--- a/ext/mysql2/extconf.rb
+++ b/ext/mysql2/extconf.rb
@@ -2,7 +2,14 @@
 require 'mkmf'
 
 def asplode lib
-  abort "-----\n#{lib} is missing.  please check your installation of mysql and try again.\n-----"
+  if RUBY_PLATFORM =~ /mingw|mswin/
+    abort "-----\n#{lib} is missing. please check your installation of mysql and try again.\n-----"
+  elsif RUBY_PLATFORM =~ /darwin/
+    abort "-----\n#{lib} is missing. Try 'brew install mysql', check your installation of mysql and try again.\n-----"
+  else
+    abort "-----\n#{lib} is missing. Try 'apt-get install libmysqlclient-dev' or
+'yum install mysql-devel', check your installation of mysql and try again.\n-----"
+  end
 end
 
 # 2.0-only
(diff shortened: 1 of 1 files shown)


ext/mysql2/extconf.rb
```diff
@@ -2,7 +2,12 @@
 require 'mkmf'
 
 def asplode lib
-  abort "-----\n#{lib} is missing.  please check your installation of mysql and try again.\n-----"
+  if RUBY_PLATFORM =~ /mingw|mswin/
+    abort "-----\n#{lib} is missing. please check your installation of mysql and try again.\n-----"
+  else
+    abort "-----\n#{lib} is missing. Try 'apt-get install libmysqlclient-dev' or 
+'yum install mysql-devel', check your installation of mysql and try again.\n-----"
+  end
```
brianmario: if we're going this far we may want to include homebrew instructions for osx as well?

kirs: Sure

kirs:
@brianmario updated.

ext/mysql2/extconf.rb
```diff
@@ -69,7 +76,8 @@ def asplode lib
   libs = ['m', 'z', 'socket', 'nsl', 'mygcc']
   while not find_library('mysqlclient', 'mysql_query', lib, "#{lib}/mysql") do
     exit 1 if libs.empty?
-    have_library(libs.shift)
+    lib = libs.shift
+    asplode(lib) unless have_library(lib)
```
brianmario: Since this is in a loop, if `have_library` fails on the first iteration the rest won't be attempted since we're raising inside of `asplode` right?

kirs: Right, because `asplode` calls `abort`.

brianmario: I guess I meant - we *want* the loop to continue and only raise if none of the options were found.

kirs:
@brianmario fixed. It it the way you meant now?

brianmario:
Yeah that works thanks!

Issue #279, opened 2012-07-25
no ipv6 support


naox:
I've compiled mysql2 gem against newest mysql version
# ldd mysql2.so

```
    libmysqlclient.so.18 => /usr/lib64/libmysqlclient.so.18 (0x00007fba0eeb8000)
```

and still there is no ipv6 support. I've compiled php against same mysql version and there are no problems.
My mysql server is at ipv6 adress. I connect to it by hostname that is translated by /etc/hosts

Gem mysql2 only gives error

Can't connect to MySQL server on 'mysql5' (111) (Mysql::Error)
# cat /etc/hosts|grep mysql5

2a01:4f8:561:5165::c12          mysql5

Like I said there is zero problems with this configuration and php+mysql

naox:
Its 3 years later and I still can't connect to ipv6 mysql server. There are no problems (on same server) with php, no problems with python but gem mysql2 in ruby still don't support ipv6 connections to server.
All you get when using domain that has AAAA record only as mysql host is:
Unknown MySQL server host 'mysql-ipv6.domain.com' (1) (Mysql2::Error)

brianmario:
@naox interesting... We're just [handing the host string](https://github.com/brianmario/mysql2/blob/master/ext/mysql2/client.c#L340) passed in [directly to libmysqlclient](https://github.com/brianmario/mysql2/blob/master/ext/mysql2/client.c#L162-L165) to do with it what it will. And as @sodabrew and I have both tested, ipv6 connectivity has definitely been verified and working. Maybe there are some other connection options being passed that are somehow conflicting? Is there also an A record for that domain that libmysqlclient is maybe attempting to use first? Have you tried passing the ip address directly instead of a hostname?

naox:
```
Unknown MySQL server host '2a01:4f8:162:12e7::c1' (1) (Mysql2::Error)
  /home/naox/.gems/2.1/gems/mysql2-0.3.18/lib/mysql2/client.rb:70:in `connect'
  /home/naox/.gems/2.1/gems/mysql2-0.3.18/lib/mysql2/client.rb:70:in `initialize'
```

```
root@sv1 [/home/naox/.gems/2.1/gems/mysql2-0.3.18/lib/mysql2]# ldd mysql2.so
        linux-vdso.so.1 =>  (0x00007fff8f3d6000)
        libmysqlclient.so.18 => /var/mysql-5.5.29/lib/libmysqlclient.so.18 (0x00007f54f4592000)
[...]

PR #594, opened 2015-03-11
Move Mysql2::Client#info to a class method for easier debugging

brianmario:
Previously a successfully connected `Mysql2::Client` instance needed to be created in order to call it's `info` instance method. When needing to debug connection issues, that can be a pain. The C api calls we're using here don't require a connection to be passed, so there's no reason for us to require a connection. Although we were previously using the connections encoding to transcode the strings returned from the C api into `Encoding.default_internal` - the strings we get back from the C api are almost _always_ going to be fixed ASCII strings, so this changes that behavior to instead just tag the string with Ruby's `us-ascii` encoding and return them as-is without any transcoding.

I came across this while trying to help debug https://github.com/brianmario/mysql2/issues/279 with @naox

### DIFF
--- a/ext/mysql2/client.c
+++ b/ext/mysql2/client.c
@@ -16,7 +16,7 @@
 
 VALUE cMysql2Client;
 extern VALUE mMysql2, cMysql2Error;
-static VALUE sym_id, sym_version, sym_async, sym_symbolize_keys, sym_as, sym_array, sym_stream;
+static VALUE sym_id, sym_version, sym_header_version, sym_async, sym_symbolize_keys, sym_as, sym_array, sym_stream;
 static ID intern_merge, intern_merge_bang, intern_error_number_eql, intern_sql_state_eql;
 
 #ifndef HAVE_RB_HASH_DUP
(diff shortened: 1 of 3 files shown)


brianmario:
Oh yeah this also includes a new key in the hash returned: `header_version` which is the static version string we saw in the header during compile time.

naox:
I'm not sure that I should be commenting of changes I hardly grasp but when using this branch most methods seem not to work anymore:
```
irb(main):005:0> a = Mysql2::Client.new :host => "mysql5", :username => "naox_a", :password => "xxxx"
irb(main):006:0> a.info
NoMethodError: undefined method `info' for Mysql2::Client:Class
        from /home/naox/.gems/2.1/gems/mysql2-0.3.18/lib/mysql2/client.rb:86:in `info'
        from (irb):6
        from /usr/local/bin/irb-2.1:11:in `<main>'
irb(main):008:0> Mysql2::Client.close
NoMethodError: undefined method `close' for Mysql2::Client:Class
        from (irb):8
        from /usr/local/bin/irb-2.1:11:in `<main>'

```

this does not happen in version installed by "gem install mysql2" :

```
irb(main):002:0> require 'mysql2'
=> true
irb(main):003:0> a = Mysql2::Client.new :host => "mysql5", :username => "naox_a", :password => "xxxxxxxxxx"
=> #<Mysql2::Client:0x007ff0f3ed3d98 @read_timeout=nil, @query_options={:as=>:hash, :async=>false, :cast_booleans=>false, :symbolize_keys=>false, :database_timezone=>:local, :application_timezone=>nil, :cache_rows=>true, :connect_flags=>2147525125, :cast=>true, :default_file=>nil, :default_group=>nil, :host=>"mysql5", :username=>"naox_a", :password=>"a"}>
irb(main):004:0> a.info
=> {:id=>50077, :version=>"5.0.77"}
irb(main):005:0> a.close
=> nil
```

brianmario:
@naox it looks like you have the new ruby code but are running the old C code. You'll need to make sure the C extension is recompiled with the new C code from this pull request in order for that to work.

naox:
I don't think so. Whole gem has been deleted, recompiled, installed from a new code (this branch).

brianmario:
Note that `f22a2497fc5340529c83676cd817de5adc698c73` is the HEAD sha of this branch:

```
irb(main):001:0> require 'mysql2'
=> true
irb(main):002:0> `git rev-list -1 HEAD`.rstrip
=> "f22a2497fc5340529c83676cd817de5adc698c73"
irb(main):009:0> Mysql2::Client.info # class method version
=> {:id=>50520, :version=>"5.5.20", :header_version=>"5.5.20"}
irb(main):003:0> client = Mysql2::Client.new(:socket => "/<snip>/mysql/socket")
=> #<Mysql2::Client:0x007fa3f38ad510 @read_timeout=nil, @query_options={:as=>:hash, :async=>false, :cast_booleans=>false, :symbolize_keys=>false, :database_timezone=>:local, :application_timezone=>nil, :cache_rows=>true, :connect_flags=>2147525125, :cast=>true, :default_file=>nil, :default_group=>nil, :socket=>"/<snip>/data/mysql/socket"}>
irb(main):004:0> client.info # instance method version
=> {:id=>50520, :version=>"5.5.20", :header_version=>"5.5.20"}
```

sodabrew:
+1 - but I wonder what made Ruby 1.8.7 fail here.

tamird:
looks like the travis build succeeded (after a rebuild?) probably good to merge this

sodabrew:
Cherry-picked as cab7d2e

PR #595, opened 2015-03-12
Misc cleanup

tamird:
@sodabrew cleanup extracted from #592


tamird:
@sodabrew this no longer depends on #592 and has no breaking changes. How do you feel about merging it pre-0.4.0?

.travis.yml
```diff
@@ -42,11 +14,8 @@ rvm:
   - 2.1
   - 2.2
   - ree
+  - rbx-2
 matrix:
-  allow_failures:
-    - env: DB=mysql57
-    - rvm: rbx-2
-    - os: osx
```
sodabrew: OS X is a fully supported target; the Travis builds for OS X fail due to Travis issues more often.

.travis.yml
```diff
@@ -54,7 +23,5 @@ matrix:
       env: DB=mariadb10
     - rvm: 2.0.0
       env: DB=mysql57
-    - rvm: rbx-2
-      env: RBXOPT=-Xgc.honor_start=true
```
sodabrew: I needed this environment variable to get deterministic `GC.start` behavior on Rubinius. Is it no longer needed, or worked around in some other way?

.travis.yml
tamird: so you want to continue to allow those failures? I haven't seen any spurious failures during my work here...

.travis_setup.sh
```diff
@@ -0,0 +1,24 @@
...
+fi
+
+# Install MySQL if OS=darwin
+if [[ x$OSTYPE =~ ^xdarwin ]]; then
+  brew install mysql
+  mysql.server start
+else
+  sudo bash .travis_ssl.sh
+  sudo service mysql restart
+fi
```
sodabrew: These should be in separate if blocks. The issue is simply that I couldn't get SSL working on OS X in Travis; in this case I don't want to give the impression that these are mutually exclusive.

spec/mysql2/client_spec.rb
```diff
@@ -136,25 +131,34 @@ def connect *args
...
+    expect(results[1]['Variable_name']).to eql('Ssl_version')
+    expect(results[1]['Value']).not_to be_nil
+    expect(results[1]['Value']).to be_kind_of(String)
+    expect(results[1]['Value']).not_to be_empty
 
     ssl_client.close
   end
 
+  def run_gc
+    if defined?(Rubinius)
```
tamird: @sodabrew I think this is the rubinius workaround

README.md
```diff
@@ -437,14 +437,13 @@ As for field values themselves, I'm workin on it - but expect that soon.
...
 
- * Ruby MRI 1.8.7, 1.9.2, 1.9.3, 2.0.0, 2.1.x, 2.2.x (ongoing patch releases)
+ * Ruby MRI 1.8.7, 1.9.3, 2.0.0, 2.1.x, 2.2.x
  * Ruby Enterprise Edition (based on MRI 1.8.7)
  * Rubinius 2.x
 
 This gem is tested with the following MySQL and MariaDB versions:
 
- * MySQL 5.0, 5.1, 5.5, 5.6, 5.7
- * MySQL Connector/C 6.0 and 6.1 (primarily on Windows)
```
sodabrew: This remains true, don't remove.

ext/mysql2/mysql_enc_name_to_ruby.h
```diff
@@ -42,7 +42,7 @@ inline
 static unsigned int
 mysql2_mysql_enc_name_to_rb_hash (str, len)
      register const char *str;
-     register unsigned int len;
+     register const unsigned int len;
```
sodabrew: Did you re-run gperf to get this, or is it a hand edit?

spec/mysql2/client_spec.rb
tamird: yep, here's the logic from rubinius: https://github.com/rubinius/rubinius/blob/4b40c4ed0d36981c079754e88e603da50ad2cb37/vm/builtin/system.cpp#L848

ext/mysql2/mysql_enc_name_to_ruby.h
tamird: hand edit - clang complained

.travis_setup.sh
tamird: done

README.md
tamird: fixed

lib/mysql2/error.rb
```diff
@@ -53,27 +56,12 @@ def sql_state=(state)
...
-          message.chars.each do |char|
-            if char.valid_encoding?
-              new_message << char
-            else
-              new_message << REPLACEMENT_CHAR
-            end
-          end
-          new_message.encode(ENCODE_OPTS)
-        end
+        message.encode(Encoding::UTF_8, ENCODE_OPTS)
```
sodabrew: I recall that we discovered simply forcing the encoding does not work here, you can have invalid sequences in the string. What we're protecting against is someone's logging library emitting an invalid UTF-8 character warning instead of the actual error from MySQL (which itself might be an invalid character warning about the query!).

Gemfile
```diff
@@ -2,22 +2,29 @@ source 'https://rubygems.org'
 
 gemspec
 
-# benchmarks
+gem 'eventmachine' unless RUBY_PLATFORM =~ /mswin|mingw/
+
+group :test do
+  gem 'rake', '~> 10.4.2'
+  gem 'rake-compiler', '~> 0.9.5'
```
sodabrew: This is not a test dependency, it's for packaging on Windows.

mysql2.gemspec
```diff
@@ -13,9 +13,4 @@ Gem::Specification.new do |s|
 
   s.files = `git ls-files README.md CHANGELOG.md MIT-LICENSE ext lib support`.split
   s.test_files = `git ls-files spec examples`.split
-
-  # tests
-  s.add_development_dependency 'rake-compiler', '~> 0.9.5'
```
sodabrew: This is a dependency for packaging on Windows.

spec/mysql2/client_spec.rb
```diff
@@ -136,25 +131,34 @@ def connect *args
...
+      GC.run(true)
```
sodabrew: Oh, does this force Rubinius GC immediately even without the environment variable?

lib/mysql2/error.rb
tamird: Just going by the comment, the tests, and the documentation on `encode` and `scrub`, these implementations are exactly equivalent. =/

spec/mysql2/client_spec.rb
tamird: yeah, exactly

Gemfile
tamird: fixed

mysql2.gemspec
tamird: fixed

.travis.yml
tamird: see https://github.com/brianmario/mysql2/pull/595#discussion_r28534932

tamird:
thanks @sodabrew! I think I've addressed everything, and we have a green build :smile_cat:

lib/mysql2/error.rb
tamird: @sodabrew I've added two commits here to try and alleviate this concern. I think it's pretty well ironclad now, what do you think?

sodabrew: I'm... not sure how we missed this in 1.9? Or it didn't actually work. I don't recall. Your change looks right according to the docs though.
http://ruby-doc.org/core-1.9.3/String.html#method-i-encode

tamird:
@sodabrew is there anything else I should touch up?

tamird:
@sodabrew looks like travis has issues with mariadb5.5 now :(

tamird:
Ping.

sodabrew:
Would you do another rebase to master now that prepared statements have landed? If you wouldn't mind also running the new statements.c through rubocop, that'd be awesome.

tamird:
This PR doesn't include RuboCop, but I've rebased it.

mysql2.gemspec
```diff
@@ -13,9 +13,4 @@ Gem::Specification.new do |s|
 
   s.files = `git ls-files README.md CHANGELOG.md LICENSE ext lib support`.split
   s.test_files = `git ls-files spec examples`.split
-
-  # tests
-  s.add_development_dependency 'rake-compiler', '~> 0.9.5'
-  s.add_development_dependency 'rake', '~> 0.9.3'
-  s.add_development_dependency 'rspec', '~> 2.8.0'
```
sodabrew: For Rubygems.org to show its dependency info, these are better in the gemspec than in the Gemfile? I moved eventmachine because we're using a feature of EM that's not available on Windows and the gemspec cannot express a platform dependency.

sodabrew:
Review update: Just the one comment on the gemspec about development dependencies.

mysql2.gemspec
tamird: I suppose for rubygems.org that's a valid point. In my experience, I've never known anyone to audit that list (as given by rubygems) for any purpose. In fact, it seems that if someone really cares about the dependency list they'd have to audit the dependencies as given by bundler, rather than rubygems.

I moved these because the gemspec can't express a variety of conditions that the Gemfile can, and the use of both just adds cognitive overhead. Does that sound reasonable?

tamird:
Rebased

tamird:
Rubinius segfaulted; it's probably fine

spec/mysql2/client_spec.rb
```diff
@@ -345,127 +345,121 @@ def connect *args
...
-            trap(:USR1, 'DEFAULT')
+        client = Mysql2::Client.new(DatabaseCredentials['root'].merge(:read_timeout => 0))
+        expect {
+          client.query('SELECT SLEEP(0.1)')
+        }.to raise_error(Mysql2::Error)
+      end
+
+      # XXX this test is not deterministic (because Unix signal handling is not)
+      # and may fail on a loaded system
+      it "should run signal handlers while waiting for a response" do
```
sodabrew: This is the test that failed on Rbx on Travis. It was previously guarded not to run on Rubinius. I think that guard should be reinstated.

mysql2.gemspec
sodabrew: The gemspec is still the place where runtime dependencies are listed, including for Bundler. I'm on the fence for the internal dependencies, but leaning towards gemspec for convention.

The version bumps look fine to me - newer rake, rake-compiler, and of course newer rspec since you've ported the tests (awesome!).

spec/mysql2/client_spec.rb
tamird: Right, so that test actually passed on a previous version of rubinius (2.2.2?), which is why I removed the guard. Once the segfault is fixed (it might already be, current rbx is 2.2.10 but travis runs 2.2.5), the test will pass, so I think we should leave the guard out.

sodabrew: Travis was running Rbx 2.5.5, which is the most recent release at this time:
https://travis-ci.org/brianmario/mysql2/jobs/65574620#L111
http://rubini.us/downloads/

tamird: Ah, you're right. It turns out adding an empty call to `fork` fixes this (at least on my machine). I'm baffled.

sodabrew:
Thanks for sticking with this PR, let's do it!

PR #605, opened 2015-04-01
Refactor extconf.rb

lowjoel:
Hello there.

I've made a pretty major rewrite for extconf.rb. It was generating bad makefiles for Windows (since it is not commonly tested) and while trying to fix that I felt that the flow of logic could be much better.

I've split the path detection logic into the `detect_*` functions, and the configuration logic to `configure_*` functions. These are different depending on the platform.

I've tried to be as faithful to the original implementation as possible, however I have not tested this on platforms other than Windows. Comments much appreciated.

### DIFF
--- a/ext/mysql2/extconf.rb
+++ b/ext/mysql2/extconf.rb
@@ -2,6 +2,9 @@
 require 'mkmf'
 require 'English'
 
+# Disable these cops, because they make the build script _harder_ to read.
+# rubocop:disable Metrics/ClassLength, Style/GuardClause
+
 # For compatibility with Ruby 1.8 and Ruby EE, whose Rake breaks Object#rm_f
 class << self
    alias_method :rm_f_original, :rm_f
@@ -9,18 +12,21 @@ class << self
    alias_method :rm_f, :rm_f_original
 end
 
+# Represents the base configuration for all platforms.
 class Platform
+  attr_accessor :rpath_dir
+
   # Gets the proper platform object for the current platform.
   def self.current
     case RUBY_PLATFORM
-      when /mswin/
-        Windows.new
-      when /mingw/
-        WindowsMingw.new
-      when /darwin/
-        MacOS.new
-      else
-        Linux.new
+    when /mswin/
+      Windows.new
+    when /mingw/
+      WindowsMingw.new
+    when /darwin/
+      MacOS.new
+    else
+      Linux.new
     end
   end
 
@@ -34,19 +40,15 @@ class Platform
 
   protected
 
-  def rpath_dir
-    @rpath_dir
-  end
-
   # @return [(String, String)|nil] The include and library paths.
   def detect_paths
     detect_by_explicit_path || detect_by_mysql_config
   end
 
   def detect_headers
-    headers = %w{ errmsg.h mysqld_error.h }
-    prefix = ['', 'mysql/'].find do |prefix|
-      have_header("#{prefix}mysql.h")
+    headers = %w(errmsg.h mysqld_error.h)
+    prefix = ['', 'mysql/'].find do |candidate|
+      have_header("#{candidate}mysql.h")
     end
     asplode('mysql.h') unless prefix
 
@@ -57,9 +59,9 @@ class Platform
   end
 
   def detect_libraries
-    libraries = %w{ mysqlclient libmysql }
-    library = libraries.find do |library|
-      have_library(library, 'mysql_query')
+    libraries = %w(mysqlclient libmysql)
+    library = libraries.find do |candidate|
+      have_library(candidate, 'mysql_query')
     end
 
     asplode 'mysqlclient or libmysql' unless library
@@ -119,18 +121,16 @@ class Platform
         inc, lib = dir_config('mysql')
       end
 
-      error "Cannot find include dir(s) #{inc}" unless inc && inc.split(File::PATH_SEPARATOR).any?{|dir| File.directory?(dir)}
-      error "Cannot find library dir(s) #{lib}" unless lib && lib.split(File::PATH_SEPARATOR).any?{|dir| File.directory?(dir)}
+      error "Cannot find include dir(s) #{inc}" unless inc && inc.split(File::PATH_SEPARATOR).any? { |dir| File.directory?(dir) }
+      error "Cannot find library dir(s) #{lib}" unless lib && lib.split(File::PATH_SEPARATOR).any? { |dir| File.directory?(dir) }
       warn "Using --with-mysql-dir=#{File.dirname(inc)}"
       [inc, lib]
-    else
-      nil
     end
   end
 
   def detect_by_mysql_config
     if mysql_config_path
-      warn  "Using mysql_config at #{mysql_config_path}"
+      warn "Using mysql_config at #{mysql_config_path}"
 
       ver = `#{mysql_config_path} --version`.chomp.to_f
       includes = `#{mysql_config_path} --include`.chomp
@@ -171,10 +171,11 @@ class Platform
   end
 end
 
+# Configuration for generic Unix
 class Unix < Platform
   # borrowed from mysqlplus
   # http://github.com/oldmoe/mysqlplus/blob/master/ext/extconf.rb
-  DEFAULT_MYSQL_CONFIG_SEARCH_PATHS = %w[
+  DEFAULT_MYSQL_CONFIG_SEARCH_PATHS = %w(
     /opt
     /opt/local
     /opt/local/mysql
@@ -186,7 +187,7 @@ class Unix < Platform
     /usr/local/mysql-*
     /usr/local/lib/mysql5*
     /usr/local/opt/mysql5*
-  ].freeze
+  ).freeze
 
   def configure
     super
@@ -202,7 +203,7 @@ class Unix < Platform
       warn "Setting mysql rpath to #{explicit_rpath}"
       $LDFLAGS << rpath_flags
     else
-      if libdir = rpath_dir[%r{(-L)?(/[^ ]+)}, 2]
+      if (libdir = rpath_dir[%r{(-L)?(/[^ ]+)}, 2])
         rpath_flags = " -Wl,-rpath,#{libdir}"
         if RbConfig::CONFIG['RPATHFLAG'].to_s.empty? && try_link('int main() {return 0;}', rpath_flags)
           # Usually Ruby sets RPATHFLAG the right way for each system, but not on OS X.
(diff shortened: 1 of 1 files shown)


lowjoel:
Hmm, the breakage is on Ruby 1.9.1, 1.8.7, and Ruby EE because of `require 'rake'`. All are no longer supported. Should they be removed from the build matrix?

sodabrew:
Thanks for the effort on this! Will review it shortly. There's a ton of material in the merge queue at the moment.

I am surprised that the Makefiles aren't correct on Windows - I spent a lot of time updating the Windows support for the 0.3.18 release, and added AppVeyor tests. What version of Windows are you using?

lowjoel:
Ah yes, Windows. extconf.rb currently _should_ work for MingW, but I compile mine using VC++. There's a ton of conditionals in extconf that test for Unixy environments (that MingW is more than VC++), and also logic for generating libmysql.a from libmysql.dll (that VC++ doesn't need).

Also, because mysqlclient.lib is a static linkage version of MySQL, and that doesn't work too well with VC++. libmysql.lib works better because it's a DLL.

line 300 is `have_func('rb_thread_blocking_region')`... I think rake has overridden `rm_f`, I don't directly use it.

sodabrew:
Oh, that's great. I really appreciate that you're a VC++ user and dug into the build system to make it better across all platforms / Rubies (...gotta get those 1.8.7 and 1.9.3 Rubies working though...). Very odd error with `rm_f`.

lowjoel:
Hopefully with the platform code split up to various classes debugging the build system and improving it should be better in future.

I'm not sure how to start fixing the old Rubies though.

tamird:
@lowjoel looks like this patch fixes it:
```diff
diff --git a/ext/mysql2/extconf.rb b/ext/mysql2/extconf.rb
index 8320218..d6505e4 100644
--- a/ext/mysql2/extconf.rb
+++ b/ext/mysql2/extconf.rb
@@ -1,6 +1,12 @@
 # encoding: UTF-8
 require 'mkmf'
-require 'rake'
+
+# rake fucks with rm_f
+class << self
+  alias_method :rm_f_original, :rm_f
+  require 'rake'
+  alias_method :rm_f, :rm_f_original
+end
 
 class Platform
   # Gets the proper platform object for the current platform.

```

sodabrew:
Strikes me as either a) a bug in rake, or b) newer rake doesn't work with 1.8.7 anymore. Thanks for looking at this, @tamird. Let's get that patch into this branch and it should be mergeable in the 0.4.0 queue.

lowjoel:
@tamird Thanks for the patch, I've rebased and hope it passes! @sodabrew

sodabrew:
Thanks for updating the PR! Logged in my (slow) review queue.

lowjoel:
Thanks!

sodabrew:
Would you do another rebase to master now that prepared statements have landed?

lowjoel:
Okay, done!

I'm not sure if I'm expecting merge conflicts though? It appears that this shouldn't affect what happens in the library, unless I'm missing something...

sodabrew:
Mostly a matter of triggering the tests to ensure that the combined code runs tests cleanly.

lowjoel:
Okay, it seems to pass :)

ext/mysql2/extconf.rb
```diff
@@ -1,188 +1,312 @@
...
+    headers.each do |header|
+      header = "#{prefix}#{header}"
+      asplode(header) unless have_header(header)
+    end
+  end
+
+  def detect_libraries
+    libraries = %w{ mysqlclient libmysql }
+    library = libraries.find do |library|
+      have_library(library, 'mysql_query')
```
sodabrew: On my Mac, I have to remove the second argument for this to succeed.

That said, in the master extconf.rb, the output from `mysql_config --libs` is used without calling have_library. This is preferred to allow the user to provide an explicit --with-mysql-config=... that points at an non-default location.

ext/mysql2/extconf.rb
```diff
@@ -1,188 +1,312 @@
...
-      dlltool = RbConfig::CONFIG['DLLWRAP'].gsub('dllwrap', 'dlltool')
-      sh dlltool, '--kill-at',
-         '--dllname', 'libmysql.dll',
-         '--output-lib', 'libmysql.a',
-         '--input-def', deffile, libfile
+      $INCFLAGS += ' ' + includes
+      $libs = "#{libs} #{$libs}"
+      [includes, libs]
+    else
+      error 'Cannot find mysql_config' if with_config('mysql-config')
```
sodabrew: I think that a user supplied `--with-mysql-config` is getting lost here and isn't used?

ext/mysql2/extconf.rb
lowjoel: So I think the problem here is that --with-mysql-config isn't properly accepted? See the other comment you made below.

lowjoel: This method only handles the detection of libraries if there is no mysql-config path specified.

ext/mysql2/extconf.rb
```diff
@@ -1,188 +1,312 @@
...
+  def asplode(library)
+    error("#{library} is missing. #{asplode_suggestion}")
+  end
+
+  def asplode_suggestion
+    'Please check your installation of MySQL and try again.'
+  end
+
+  def mysql_config_path
+    nil
```
lowjoel: Would it work if this returns the default, i.e. `@mysql_config_path ||= with_config('mysql-config')`

sodabrew: I think this would be good - each subclass would have to call super first, or have different names for the subclass finder I suppose.

The philosophy is that if a user specifies a --with value, use it without trying other options.

lowjoel:
@sodabrew I've added the check and rebased. Have a look.

sodabrew:
Thanks for updating! Please rebase again when you get a chance, and I think there may be one more rebase next before this PR is up for final review.

lowjoel:
Done!

lowjoel:
Wait, @sodabrew: this was recently added:

```ruby
# This is our wishlist. We use whichever flags work on the host.
  # -Wall and -Wextra are included by default.
  # TODO: fix statement.c and remove -Wno-error=declaration-after-statement
  %w(
    -Werror
    -Weverything
    -fsanitize=address
    -fsanitize=integer
    -fsanitize=thread
    -fsanitize=memory
    -fsanitize=undefined
    -fsanitize=cfi
    -Wno-error=declaration-after-statement
  ).each do |flag|
    if try_link('int main() {return 0;}', flag)
      $CFLAGS << ' ' << flag
    end
  end
```

That only applies for GCC/clang? VC most definitely do not support those.

sodabrew:
Could you rebase once more? Sorry about the ever-changing whitespace in extconf.rb this week!

lowjoel:
@sodabrew no problem. I've added the compile flag wishlist back to the same one as current master. Also I saw that mariadb was added to the paths to check, my branch has been fixed to support that too.

tamird:
Apologies in advance; #633 is about to drop another conflict on this

lowjoel:
@tamird Ping me when that's merged. The only thing I'd preserve from #634 would be to add another directory to check for mysql's dev files (/usr/local/opt/mysql5*, which I think is for homebrew-like installs?)

tamird:
You should preserve the entire diff from #634 because the current code results in an infinite loop if it can't find the mysql client headers.

lowjoel:
@tamird OK -- I see how the loop could continue forever. Does b492763e5d4a595cb2e5bbb1d64da8f508e343a7 solve the problem?

tamird:
Why do you insist on keeping this cruft?

lowjoel:
Well, I'd rather not break things unnecessarily. If the decision is that we don't need this, then I'd be happy to omit it.

sodabrew:
I traced the history of that part of extconf.rb, and the extra library search block was borrowed from mysqlplus, going back to the original commit to mysqlplus, https://github.com/oldmoe/mysqlplus/commit/d7e47a3f2da222d1923bf7d1a6077e32621ac46d, which comes from an earlier gem called neverblock-mysql, https://rubygems.org/gems/oldmoe-neverblock-mysql/versions/0.1.0, the sources for which are no longer online. I think we can let it go :)

lowjoel:
@sodabrew rebased, and removed that block of code.

ext/mysql2/extconf.rb
```diff
@@ -1,192 +1,322 @@
...
-  if RUBY_PLATFORM =~ /mingw|mswin/
-    abort "-----\n#{lib} is missing. please check your installation of mysql and try again.\n-----"
-  elsif RUBY_PLATFORM =~ /darwin/
-    abort "-----\n#{lib} is missing. Try 'brew install mysql', check your installation of mysql and try again.\n-----"
-  else
-    abort "-----\n#{lib} is missing. Try 'apt-get install libmysqlclient-dev' or
-'yum install mysql-devel', check your installation of mysql and try again.\n-----"
-  end
+class << self
+   alias_method :rm_f_original, :rm_f
```
tamird: this needs a comment, no?

ext/mysql2/extconf.rb
```diff
@@ -1,192 +1,322 @@
...
-).select do |flag|
-  try_link('int main() {return 0;}', flag)
-end.each do |flag|
-  $CFLAGS << ' ' << flag
-end
+      # MySQL 5.5 and above already have re-entrant code in libmysqlclient (no _r).
+      if ver >= 5.5 || libs.empty?
+        libs = `#{mysql_config_path} --libs`.chomp
+      end
+      exit 1 if $? != 0
```
tamird: nit: `abort 'message' unless $CHILD_STATUS.success?`

tamird:
looks ok to me, but the commit history is a mess :(

lowjoel:
I'll squash it once this is ready to be merged. This PR has been going on for 3 months and rebased many times.

ext/mysql2/extconf.rb
lowjoel: Put that in, thanks!

sodabrew:
Sorry for the delay, I am going to put this off until after 0.4.0 - I'll be cherry-picking in the infinite loop fix from the 0.3.x branch in the mean time.

The blocker is that I'm not having luck out-of-the-box on OS X with MacPorts. It's finding my mysql_config binary, but doesn't seem to be picking up the paths from it. I don't have time to debug it just at the moment, and I don't want to delay 0.4 any further.

tamird:
0.4 is out and the rubocop PR landed earlier today. @lowjoel would you rebase this?

lowjoel:
Rebased, have a look.

I've suppressed 2 classes of Rubocop warnings, they are:
```
extconf.rb:13:1: C: Metrics/ClassLength: Class has too many lines. [120/100]
class Platform
^^^^^
extconf.rb:111:5: C: Style/GuardClause: Use a guard clause instead of wrapping the code inside a conditional expression.
    if inc && lib
    ^^
extconf.rb:330:5: C: Style/GuardClause: Use a guard clause instead of wrapping the code inside a conditional expression.
    unless RbConfig::CONFIG['host_os'] =~ /mswin|mingw/
    ^^^^^^
```
I think following the recommendation would make the build script even harder to read.

If you're okay with this, I'll squash it to one commit.

tamird:
Attempting to get this off my GH queue.

sodabrew:
Closing as superseded. This predates the current Windows build story entirely: extconf.rb now vendors libmariadb.dll from MSYS2's MariaDB Connector/C package via rake-compiler-dock for precompiled gems, and #1476 added real Windows CI to catch build regressions going forward. Neither existed when this was opened in 2015, and extconf.rb has been rewritten many times since -- there isn't a way to rebase this onto current master without discarding essentially the whole diff. Thanks for the work at the time.

PR #628, opened 2015-06-08
Fix more C warnings

tamird:
Turns out there was a bug in how CFLAGS were being added in extconf; fixing it made '-Weverything' actually turn on, and a bunch of issues were revealed.


sodabrew:
This fails to compile on OS X for me, due to a "Default label in switch which covers all enumeration values" warning. Even if we added a -Wno-error for that, I think -Weverything is too noisy to ship to end-users.

tamird:
I kind of like having to explicitly list which warnings we do not comply with and the reasons we do not.

sodabrew:
It's in two enumerations of MySQL return types in result.c:
```
ext/mysql2/result.c:298:7: error: default label in switch which covers all enumeration values [-Werror,-Wcovered-switch-default]
ext/mysql2/result.c:706:9: error: default label in switch which covers all enumeration values [-Werror,-Wcovered-switch-default]
```

tamird:
Oh, you must be compiling against a different version of the mysql header - on my system, those enumerations do not fully cover all values, so the warning is not issued. Which mysql are you compiling against?

sodabrew:
MySQL 5.5 -- of course we support compiling against a range of versions, so the default is an important fallback :)

tamird:
Ah, turns out we can kill the warnings entirely with `s/Wno-error=/Wno-/g`. PTAL

sodabrew:
The next issue is thornier. There are a number of shortenings ala `-Wshorten-64-to-32` and signedness conversion due to Ruby arrays being platform-signed-long in length[1] while MySQL result sets are an unsigned 64 on all platforms.

1 - http://rxr.whitequark.org/mri/source/include/ruby/ruby.h#927


```
ext/mysql2/result.c:749:35: error: comparison of integers of different signs: 'long' and 'unsigned int' [-Werror,-Wsign-compare]
  if (RARRAY_LEN(wrapper->fields) != wrapper->numberOfFields) {
      ~~~~~~~~~~~~~~~~~~~~~~~~~~~ ^  ~~~~~~~~~~~~~~~~~~~~~~~
ext/mysql2/result.c:905:45: error: implicit conversion loses integer precision: 'my_ulonglong' (aka 'unsigned long long') to 'unsigned long' [-Werror,-Wshorten-64-to-32]
    wrapper->numberOfRows = wrapper->stmt ? mysql_stmt_num_rows(wrapper->stmt) : mysql_num_rows(wrapper->result);
                          ~                 ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
ext/mysql2/result.c:905:82: error: implicit conversion loses integer precision: 'my_ulonglong' (aka 'unsigned long long') to 'unsigned long' [-Werror,-Wshorten-64-to-32]
    wrapper->numberOfRows = wrapper->stmt ? mysql_stmt_num_rows(wrapper->stmt) : mysql_num_rows(wrapper->result);
                          ~                                                      ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

tamird:
Neat. Where did you find a 32bit machine? :) PTAL

sodabrew:
Oddly, this is on my desktop Mac - it's 64-bit and compiled as such - so I'm confused, too! In result.h, numberOfRows is an unsigned long, which certainly should be 64-bit.

tamird:
These failures look spurious, can you retry the build? I've changed numberOfRows and numberOfFields to both be my_ulonglong, so perhaps the error you saw is cleared?

sodabrew:
I think #629 satisfies most of the goals of this PR but with less ambitious compiler warnings. Let me know if there are items in this branch you would still hope to see land, or consider it resolved and close the PR.

PR #630, opened 2015-06-10
Statement compile

### LINKED ISSUE #601: Should we be using StringValueCStr instead of StringValuePtr?
I saw this on the Ruby Weekly:

https://github.com/ged/ruby-pg/pull/5
http://stackoverflow.com/questions/28891447/weird-bytes-added-to-attribute-after-save-in-rails

### LINKED ISSUE #610: Calling count on a newly initalized result object causes segfault
Tested on ruby 1.8.7 and 2.2.0
With mysql2 gem version 0.3.18

Stumbled across this when trying to build up some unit tests.

irb(main):001:0> require 'rubygems'
=> true
irb(main):002:0> require 'mysql2'
=> true
irb(main):003:0> a = Mysql2::Result.new
=> #<Mysql2::Result:0x7f96e31e63c0>
irb(main):004:0> a.count
(irb):4: [BUG] Segmentation fault

sodabrew:
A few fixes in statement.c and resolves #601 and #610.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -29,6 +29,8 @@ environment:
     - ruby_version: "200-x64"
     - ruby_version: "21"
     - ruby_version: "21-x64"
+    - ruby_version: "22"
+    - ruby_version: "22-x64"
 cache:
   - vendor
 services:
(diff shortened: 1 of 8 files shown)


tamird:
LGTM

Issue #633, opened 2015-06-10
Tests fail with MySQL 5.5 on OSX


tamird:
Below output is with 1.9.3 but the same error happens with 2.2.2 and rbx-2.5.5 as well.

```
$ git clean -dxf && PATH=$(brew --prefix mysql55)/bin:$PATH bundle exec rake
Removing Gemfile.lock
Removing lib/mysql2/mysql2.bundle
Removing spec/configuration.yml
Removing spec/my.cnf
Removing tmp/
mkdir -p tmp/x86_64-darwin14.3.0/mysql2/1.9.3
cd tmp/x86_64-darwin14.3.0/mysql2/1.9.3
/Users/tamird/.rubies/ruby-1.9.3-p547/bin/ruby -I. ../../../../ext/mysql2/extconf.rb
checking for ruby/thread.h... no
checking for rb_thread_blocking_region()... yes
checking for rb_wait_for_single_fd()... yes
checking for rb_hash_dup()... yes
checking for rb_intern3()... yes
-----
Using mysql_config at /usr/local/opt/mysql55/bin/mysql_config
-----
checking for mysql.h... yes
checking for errmsg.h... yes
checking for mysqld_error.h... yes
-----
Setting rpath to /usr/local/opt/mysql55/lib
-----
creating Makefile
cd -
cd tmp/x86_64-darwin14.3.0/mysql2/1.9.3
make
compiling ../../../../ext/mysql2/client.c
compiling ../../../../ext/mysql2/infile.c
compiling ../../../../ext/mysql2/mysql2_ext.c
compiling ../../../../ext/mysql2/result.c
compiling ../../../../ext/mysql2/statement.c
linking shared-object mysql2/mysql2.bundle
cd -
mkdir -p tmp/x86_64-darwin14.3.0/stage/lib/mysql2
cp LICENSE tmp/x86_64-darwin14.3.0/stage/LICENSE
cp README.md tmp/x86_64-darwin14.3.0/stage/README.md
mkdir -p tmp/x86_64-darwin14.3.0/stage/examples
cp examples/eventmachine.rb tmp/x86_64-darwin14.3.0/stage/examples/eventmachine.rb
cp examples/threaded.rb tmp/x86_64-darwin14.3.0/stage/examples/threaded.rb
mkdir -p tmp/x86_64-darwin14.3.0/stage/ext/mysql2
cp ext/mysql2/client.c tmp/x86_64-darwin14.3.0/stage/ext/mysql2/client.c
[...]

PR #634, opened 2015-06-10
Test against mysql55 on OSX

tamird:
See #633.

### DIFF
--- a/ext/mysql2/extconf.rb
+++ b/ext/mysql2/extconf.rb
@@ -34,6 +34,7 @@ dirs = ENV['PATH'].split(File::PATH_SEPARATOR) + %w[
   /usr/local/mysql
   /usr/local/mysql-*
   /usr/local/lib/mysql5*
+  /usr/local/opt/mysql5*
 ].map{|dir| "#{dir}/bin" }
 
 GLOB = "{#{dirs.join(',')}}/{mysql_config,mysql_config5,mariadb_config}"
@@ -73,14 +74,8 @@ elsif mc = (with_config('mysql-config') || Dir[GLOB].first)
   rpath_dir = libs
 else
   inc, lib = dir_config('mysql', '/usr/local')
-  libs = ['m', 'z', 'socket', 'nsl', 'mygcc']
-  found = false
-  while not find_library('mysqlclient', 'mysql_query', lib, "#{lib}/mysql") do
-    exit 1 if libs.empty?
-    found ||= have_library(libs.shift)
-  end
 
-  asplode("mysql client") unless found
+  asplode("mysql client") unless find_library('mysqlclient', 'mysql_query', lib, "#{lib}/mysql")
 
   rpath_dir = lib
 end
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,8 +3,6 @@ bundler_args: --without benchmarks development
 before_install:
   - gem --version
   - bash .travis_setup.sh
-  - mysqld --version
-  - mysql -u $USER -e "CREATE DATABASE IF NOT EXISTS test"
 os:
   - linux
 rvm:
@@ -24,4 +22,8 @@ matrix:
     - rvm: 2.0.0
       env: DB=mysql57
     - rvm: 2.0.0
+      env: DB=mysql
+      os: osx
+    - rvm: 2.0.0
+      env: DB=mysql55
       os: osx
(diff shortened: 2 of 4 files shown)


sodabrew:
Thank you for debugging this!

tamird:
Whoa, hilarious. I've added a `brew update` which should hopefully include the bottle.

sodabrew:
Infinite loop of "checking for mysql_query() in -lmysqlclient... no" -- interesting failure in this block in extconf.rb
``` ruby
  while not find_library('mysqlclient', 'mysql_query', lib, "#{lib}/mysql") do
     exit 1 if libs.empty?
     found ||= have_library(libs.shift)
   end
```
https://travis-ci.org/brianmario/mysql2/jobs/66233625"

tamird:
I think I fixed that. Fingers crossed.

tamird:
Huzzah, this now repros in CI.

ext/mysql2/extconf.rb
```diff
@@ -73,14 +74,8 @@ def asplode lib
   rpath_dir = libs
 else
   inc, lib = dir_config('mysql', '/usr/local')
-  libs = ['m', 'z', 'socket', 'nsl', 'mygcc']
```
sodabrew: I wonder what systems may have required these extra link options. BSD? Solaris maybe?

tamird: ¯\_(ツ)_/¯

sodabrew:
Well isn't that a thing. Smells like a MySQL bug.

tamird:
Yeah, that's pretty unfortunate. I'd like to dig into it more but don't really know where to being. Any guidance? Also cool if you want to take it on.

Super strange that it's OSX only.

sodabrew:
So far I haven't found anything obvious, searching here: http://bugs.mysql.com/search.php

tamird:
OK, so this is interesting. This is a bug in libmysqlclient, not the running mysql server. I compiled mysql2 against 5.5 and then ran the test suite against a running 5.6 and **was** able to repro.

tamird:
Further confirmed: compiling against 5.6 and running against 5.5 does **not** repro.

tamird:
Looks like the problem is in `mysql_read_query_result`. Digging.

sodabrew:
http://bugs.mysql.com/bug.php?id=65956
THIS -> http://bugs.mysql.com/bug.php?id=62578

tamird:
Oh that's very interesting. http://bugs.mysql.com/bug.php?id=62578 claims that percona-server fixed it in 5.5. Gonna test it.

tamird:
Nope, still broken with percona-server 5.5.41-37.0

tamird:
I wonder what would happen if I compiled mysql with `-DTHREAD_SAFE_CLIENT`...

sodabrew:
Since we've got it narrowed down to MySQL 5.5 client, I'm comfortable making those tests pending:
``` ruby
  pending("libmysqlclient 5.5 does not reconnect after interrupt, see http://bugs.mysql.com/bug.php?id=62578") if Mysql::Client.info[:version] =~ /^5.5/
```

tamird:
Cool, I'll make that change shortly.

tamird:
Actually it still bothers me that this doesn't repo on linux. Any thoughts on that?

tamird:
Also, neither of those mysql bugs are reproducible on my machine, so I don't think that's what we're seeing here.

tamird:
Alright, I've whitelisted the tests, this should be good to merge.

ext/mysql2/extconf.rb
lowjoel: I've preserved this in my refactored extconf -- this only applies to Unix-like platforms (with good reason)

tamird: Sorry, what do you mean? When are these extra libraries needed?

lowjoel: I can't put my finger on this, but I believe this is for compatibility with _ancient_ compilers. Since we have tests running against Ruby 1.8, I'm not discounting that...

tamird: Maybe, but CI passed with this change.

tamird:
Bogus failures =/

tamird:
@sodabrew ready to merge this?

sodabrew:
Narrowed down to just OS X? Ok, I suppose since the CI's MySQL 5.5-ubuntu package isn't exhibiting the problem.

tamird:
Yep, the beauty of `pending` is that it asserts that the test definitely fails - given that we're green on Travis' ubuntu, this is an OSX thing.

PR #637, opened 2015-07-02
Add rake-compiler-dock for building Windows binary gems.

larskanis:
[rake-compiler-dock](https://github.com/larskanis/rake-compiler-dock) makes building Windows gems faster and with little to no setup compared to [rake-compiler-dev-box](https://github.com/tjschuck/rake-compiler-dev-box). A simple 'rake gem:windows' is usually enough.

This also removes warning about unused variable in the Windows branch, that otherwise breaks the build.

The build of the mysql2 gem is successfully tested on Linux and Windows. I tested all resulting gems on RubyInstaller versions 2.1.4-x64, 2.2.1-x86 and 2.2.1-x64 on Windows-7.

The cross ruby versions used by gem:windows are not passed to the container, so the rake-compiler-dock's defaults are used, which are [defined here](https://github.com/larskanis/rake-compiler-dock/blob/master/Dockerfile#L113).

### DIFF
--- a/ext/mysql2/client.c
+++ b/ext/mysql2/client.c
@@ -880,7 +880,7 @@ static VALUE rb_mysql_client_server_info(VALUE self) {
  *
  * Return the file descriptor number for this client.
  */
-static VALUE rb_mysql_client_socket(VALUE self) {
+static VALUE rb_mysql_client_socket(RB_MYSQL_UNUSED VALUE self) {
 #ifndef _WIN32
   GET_CLIENT(self);
   REQUIRE_CONNECTED(wrapper);
--- a/Gemfile
+++ b/Gemfile
@@ -21,6 +21,7 @@ end
 
 group :development do
   gem 'pry'
+  gem 'rake-compiler-dock', '~> 0.4.2'
 end
 
 platforms :rbx do
--- a/tasks/compile.rake
+++ b/tasks/compile.rake
@@ -86,3 +86,12 @@ else
     Rake::Task['cross'].prerequisites.unshift 'vendor:mysql:cross'
   end
 end
+
+desc "Build the windows binary gems per rake-compiler-dock"
+task 'gem:windows' do
+  require 'rake_compiler_dock'
+  RakeCompilerDock.sh <<-EOT
+    bundle install "--without=test benchmarks development rbx" &&
+    rake cross native gem
+  EOT
+end


larskanis:
As requested in https://github.com/tjschuck/rake-compiler-dev-box/pull/22 , I added unzip to the docker image. This will be part of rake-compiler-dock-0.4.2, so the apt-get part in the PR can be removed eventually.

larskanis:
The PR is now updated to rake-compiler-dock-0.4.2, which has unzip included.

sodabrew:
Thank you! This looks really neat! I will try making some test builds to sanity check before merge.

Could I ask to split the `RB_MYSQL_UNUSED` macro and the Gemfile/Rakefile changes into separate commits?

larskanis:
@sodabrew I's updated into to two commits, now.

ext/mysql2/client.c
```diff
@@ -880,7 +880,7 @@ static VALUE rb_mysql_client_server_info(VALUE self) {
  *
  * Return the file descriptor number for this client.
  */
-static VALUE rb_mysql_client_socket(VALUE self) {
+static VALUE rb_mysql_client_socket(RB_MYSQL_UNUSED VALUE self) {
```
sodabrew: I don't want to mark this as unused in all cases. It may make sense to move the ifdef outwards, and have two separate function definitions for windows and non-windows.

sodabrew:
By the way, are you familiar enough with rake-compiler to help with #586?

ext/mysql2/client.c
larskanis: In pg.gem we have
```C
#define UNUSED(x) ((void)(x))
```
and we mark the variable as
```C
UNUSED(self);
```
in the branch, that doesn't use this variable.
This was my first thought, but then I noticed `RB_MYSQL_UNUSED` .

larskanis:
Yes, I'm familiar enough with it and I already noticed https://github.com/brianmario/mysql2/pull/586, but had no time to fix this, so far. However it should be possible to pick the right DLL based on the gemspec that is passed to `cross_compiling` [here](https://github.com/brianmario/mysql2/blob/master/tasks/compile.rake#L27) .

ext/mysql2/client.c
sodabrew: It's a hint to the compiler that the argument won't be used - I think it is only to suppress the warning - I don't know if there is an assembly output or optimization difference (I would be interested to find out! but not the hard way ;))

larskanis: The above `UNUSED(self);` would work and to define two different versions of this function would work equally. I'll update the pull request based on your desire.

As a side-note: In pg.gem I solved the socket issue on Windows years ago: https://github.com/larskanis/ruby-pg/commit/84a1bb7fa40b59001

sodabrew: Do I see a way to get the socket handle on Windows with Ruby 2.0 in there!?

larskanis: Of course, this is what 'rb_w32_wrap_io_handle' is made for.

sodabrew:
Cherry-picked rake-compiler-dock as 99c57e9b5fe140620dea833b0a648fa54b6d4fec, but I'd like to do something different for the unused variable warning.

PR #640, opened 2015-07-12
Allow query options to be set on each Statement object

sodabrew:
The original Prepared Statements API didn't provide a way to pass query options into the prepare or execute phase of a statement. Rather, you'd have to modify the query options on the Client object prior to each Statement#execute.

In this PR, I've added a query options argument to Client#prepare, analogous to Client#query, and save a copy of these options into the Statement object, to be used for each Statement#execute.

I sort of wish the options were on Statement#execute, but I can't think of any way to differentiate between "argument that is supposed to be a ? replacement" and "argument that is the options hash".

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -23,7 +23,7 @@ Metrics/BlockNesting:
 # Offense count: 1
 # Configuration parameters: CountComments.
 Metrics/ClassLength:
-  Max: 125
+  Max: 128
 
 # Offense count: 3
 Metrics/CyclomaticComplexity:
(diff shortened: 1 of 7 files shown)


sodabrew:
@jeremyevans Thanks for bringing this issue up in #774. I've hesitated on this PR because I wasn't sure if it was actually needed, or if it even made sense to store different query options into each statement. It's also not obvious to me whether a change to query options on the Client itself should override the options on a Statement that was created previously. Your feedback as a user greatly appreciated!

sodabrew:
Closing in favor of #912

PR #642, opened 2015-07-14
Fix encoding if Mysql2#escape was nothing escaped

takatoshiono:
If `Encoding.default_internal` and the mysql encoding is different, `Encoding::CompatibilityError` is raised when concatenating a escaped string and a not escaped string. Because of `Mysql2#escape` returns a different encoding according to the string is escaped or not.

This pull request changes this. `Mysql2#escape` returns the same encoding regardless of whether it has been escaped.

### DIFF
--- a/ext/mysql2/client.c
+++ b/ext/mysql2/client.c
@@ -720,6 +720,11 @@ static VALUE rb_mysql_client_real_escape(VALUE self, VALUE str) {
   newLen = mysql_real_escape_string(wrapper->client, (char *)newStr, RSTRING_PTR(str), oldLen);
   if (newLen == oldLen) {
     /* no need to return a new ruby string if nothing changed */
+#ifdef HAVE_RUBY_ENCODING_H
+    if (default_internal_enc) {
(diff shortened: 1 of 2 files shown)


sodabrew:
The returned string must be in the connection encoding, not the default encoding. That error is correct, you cannot concatenate material that is not encoded for the database to material that is encoded for the database and then ship the result to the database.

takatoshiono:
In the master branch, the returned string encoding is depend on the string is changed by `mysql_real_escape_string` or not. see [this](https://github.com/brianmario/mysql2/blob/b3cfa21edf503a9513435ee8551271571d22d11d/ext/mysql2/client.c#L720-735). If nothing changed by `mysql_real_escape_string`, the returned string is the connection encoding but if it changed, the returned string is the default internal encoding by the following code.

```ruby
    if (default_internal_enc) {
      rb_str = rb_str_export_to_enc(rb_str, default_internal_enc);
    }
```

sodabrew:
Oh, you're right.

takatoshiono:
Thank you :smile:

PR #643, opened 2015-07-14
Fix more C warnings

tamird:
Revival of #628 (sorry, it fell off my radar while you were investigating the statement issues).

I think this should be good to merge.

### DIFF
--- a/ext/mysql2/extconf.rb
+++ b/ext/mysql2/extconf.rb
@@ -93,18 +93,35 @@ end
 end
 
 # This is our wishlist. We use whichever flags work on the host.
-# TODO: fix statement.c and remove -Wno-declaration-after-statement
-# TODO: fix gperf mysql_enc_name_to_ruby.h and remove -Wno-missing-field-initializers
-%w(
-  -Wall
-  -Wextra
-  -Werror
-  -Wno-unused-function
-  -Wno-declaration-after-statement
-  -Wno-missing-field-initializers
-).each do |flag|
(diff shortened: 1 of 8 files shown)


sodabrew:
Needs a rebase since the rubocop PR included the gperf changes, and other comments.

ext/mysql2/extconf.rb
```diff
@@ -94,18 +94,32 @@ def asplode lib
...
-  try_link('int main() {return 0;}', flag)
-end.each do |flag|
+# -Wall and -Wextra are included by default.
+[
+  '-fsanitize=address',
+  '-fsanitize=cfi',
+  '-fsanitize=integer',
+  '-fsanitize=memory',
+  '-fsanitize=thread',
+  '-fsanitize=undefined',
```
sodabrew: I don't think people would want to go to production with these?

ext/mysql2/extconf.rb
```diff
@@ -94,18 +94,32 @@ def asplode lib
...
+  '-Werror',
+  '-Weverything',
```
sodabrew: I'm concerned that -Weverything is an open-ended list, so future compilers will add in more checks that we might not pass. I would use -Weverything without -Werror, but given the number of allowances below, I don't want to use -Weverything.

tamird: We could use `-Werror` only in CI?

sodabrew:
I really like the idea of -Werror only in CI. Would help in cases like #674

tamird:
hurray!

PR #656, opened 2015-08-24
Travis addon mariadb

grooverdan:
travis has a mariadb addon that covers version 5.5, 10.0 and 10.1 so this patch replaces the existing install mechanisms and uses the plugin.

libmariadbd-dev is still needed but I'll see if I can get that into the travis mariadb plugin.

create database if not exists isn't in mariadb-5.5 so I use mariadb specific tests to avoid the if exists for the 5.5 version.


grooverdan:
build failure is as travis-ci/travis-core/pull/469 hasn't been pushed to production.

sodabrew:
Why is libmariadbd-dev needed? That's for the embedded database...

.travis_mariadb.sh
```diff
@@ -1,16 +0,0 @@
...
-apt-key adv --recv-keys --keyserver hkp://keyserver.ubuntu.com:80 0xcbcb082a1bb943db
-
-if [[ x$1 = xmariadb55 ]]; then
-  add-apt-repository 'deb http://ftp.osuosl.org/pub/mariadb/repo/5.5/ubuntu precise main'
-elif [[ x$1 = xmariadb10 ]]; then
-  add-apt-repository 'deb http://ftp.osuosl.org/pub/mariadb/repo/10.0/ubuntu precise main'
-fi
-
-apt-get update
-apt-get -o Dpkg::Options::=--force-confdef -o Dpkg::Options::=--force-confold -y install mariadb-server libmariadbd-dev
```
sodabrew: Oh, I didn't notice I had libmariadbd-dev in the old config too. I wonder if there isn't a better client library to use?

grooverdan: libmariadb-dev and libmysql-dev would be pretty close. Linking against both is probably a reasonable test.

libmariadbd-dev is the development client libraries not the embedded db.

sodabrew:
Hi! Any news on the libmariadbd-dev package?

sodabrew:
Merged this work in #699.

Issue #485, opened 2014-02-04
Could you consider maintaining a Changelog?


jlecour:
Hi,

Mysql2 is such an important gem, and it's mainly not written in Ruby, which makes it difficult to follow along.

It would be very useful if you could maintain a Changelog.

Thanks.

jaredbeck:
>> Please add a description to the 0.3.17 entry ..
>>
>  I have a tab open with change notes but haven't submitted it yet. I'll get it posted soon.
>

Three months later, do you still have that tab open? :)  I know some people who keep tabs open that long.

Anyway, we could use a description for 0.3.18 also.  Thanks.

sodabrew:
You know what, I had a Chrome crash a few weeks ago that took out all of my tabs - and that was one of them! I'll go update the releases now. Thanks for pinging.

sodabrew:
Done with 0.3.18, 0.3.17. I will try to go back and do 0.3.{13, 14, 15} as well.

jaredbeck:
Thanks :heart:

On Tuesday, February 17, 2015, Aaron Stone <notifications@github.com> wrote:

> Done with 0.3.18, 0.3.17. I will try to go back and do 0.3.{13, 14, 15} as
> well.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/brianmario/mysql2/issues/485#issuecomment-74774646>.
>


-- 
Jared Beck • (607) 216-5373 • jared@jaredbeck.com

PR #659, opened 2015-08-31
Add CHANGELOG.md

marknadig:
addresses #485 

I think it would be helpful to have an explicit pointer to the release notes. I know I check for a CHANGELOG first, and then the README. I went searching for changes from 3.19 and didn't find either a changelog or a reference in README. It took a while before I found this issue.


sodabrew:
Thanks!

PR #665, opened 2015-09-06
Test on ruby-head, allow failures

sodabrew:
@tamird @cbandy The query timeout workaround is failing on the uninitialized constant Timeout::ExitException in ruby-head. I wonder if they renamed this upstream!? I'll be sorting through https://github.com/ruby/ruby/blob/trunk/lib/timeout.rb soon...

https://travis-ci.org/sodabrew/mysql2/jobs/79014190

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,7 @@ rvm:
   - 2.2
   - ree
   - rbx-2
+  - ruby-head
 matrix:
   include:
     - rvm: 2.0.0
@@ -30,3 +31,5 @@ matrix:
     - rvm: 2.0.0
       env: DB=mysql55
       os: osx
+  allow_failures:
+    - rvm: ruby-head


sodabrew:
Yep: https://github.com/ruby/ruby/commit/0f663b244944c5eb2b6c7d0862c15c377747df05

tamird:
Sigh. Easy enough to solve by defining an alias at load-time.

cbandy:
Maybe it's because I wasn't paying attention to the initial implementation...

Aren't there other places where interrupts can affect the protocol state? Are there some defines that help us do this around specific C calls in the extension?

sodabrew:
There are other places where the interruption would cause a problem, but Tamir zoomed in on the most "obvious" and common point.

This issue is certainly not a blocker at the moment, I am just getting ready for the upcoming Ruby release in December.

jeremy:
Just hit this upgrading Active Record to support 0.4.0 (we test on Ruby trunk, too).

jeremy:
Note that the `ExitException` removal was backported, too, but with an alias for compatibility: https://github.com/ruby/ruby/commit/40270906824a981652719c5e614c2cf4d1c234b0#diff-32f7163092f0e8e64a3936fba1a3d59dR47

tamird:
@jeremy sadly that won't help here as this library still supports all the way back to 1.8.7 (though the relevant versions are those that support `Thread.handle_interrupt` and don't have the backport: 2.0 and 2.1).

jeremy:
@tamird Right, the backports only demonstrate that aliasing `ExitException = Error` is the way to go. See #671 😁

PR #666, opened 2015-09-06
cleanup: use rspec's type matcher

tamird:


sodabrew:
In a couple places, you switch from `be_kind_of Foo` to `be_instance_of Foo`. What's the difference?

tamird:
`instance_of` is more strict; it won't match subclasses.

sodabrew:
Ah, nice summary at https://www.relishapp.com/rspec/rspec-expectations/v/2-0/docs/matchers/type-check-matchers - ok this makes sense. Would you rebase?

tamird:
rebased

sodabrew:
Love it when tests are green!

PR #667, opened 2015-09-07
Get MySQL 5.7 SSL specs to pass with pre-generated SSL certs 

sodabrew:
To get MySQL 5.7 to fully pass tests:
* Updated the installer a bit
* Create the self-signed cert for "localhost" so that it can validate
* Add new `:protocol` option so that we can force TCP connections to localhost which allows the localhost SSL cert to be validated.


sodabrew:
Trying something new - just generating the cert for the current hostname.

Alternatively, because cert generation is pretty slow, pre-generate certs for a dummy hostname and add then as localhost aliases: http://docs.travis-ci.com/user/customizing-the-build/#Custom-Hostnames

spec/ssl/cert.cnf
```diff
@@ -0,0 +1,22 @@
...
+
+countryName_default            = US
+stateOrProvinceName_default    = CA
+localityName_default           = San Francisco
+0.organizationName_default     = mysql2_gem
+organizationalUnitName_default = Mysql2Gem
+emailAddress_default           = mysql2gem@example.com
+
+
+commonName_default             = mysql2gem.examle.com
```
sodabrew: typo.

PR #669, opened 2015-09-07
Fix test that implicitly creates a passwordless user

jeremy:
When the MySQL server has `NO_AUTO_CREATE_USER`, a GRANT can't implicitly
create a user unless a password is specified. To fix, just provide a
dummy password in the example db config and update the GRANT.

### DIFF
--- a/spec/configuration.yml.example
+++ b/spec/configuration.yml.example
@@ -7,11 +7,11 @@ root:
 user:
   host: localhost
   username: LOCALUSERNAME
-  password:
+  password: test
   database: mysql2_test
 
 numericuser:
   host: localhost
   username: LOCALUSERNAME
-  password:
+  password: test
   database: 12345
--- a/spec/configuration.yml.example
+++ b/spec/configuration.yml.example
@@ -3,15 +3,3 @@ root:
   username: root
   password:
   database: test
-
-user:
-  host: localhost
-  username: LOCALUSERNAME
-  password: test
-  database: mysql2_test
-
-numericuser:
-  host: localhost
-  username: LOCALUSERNAME
-  password: test
-  database: 12345
--- a/spec/mysql2/client_spec.rb
+++ b/spec/mysql2/client_spec.rb
@@ -211,13 +211,13 @@ RSpec.describe Mysql2::Client do
   end
 
   it "should be able to connect to database with numeric-only name" do
-    creds = DatabaseCredentials['numericuser']
-    @client.query "CREATE DATABASE IF NOT EXISTS `#{creds['database']}`"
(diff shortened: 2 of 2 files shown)


spec/mysql2/client_spec.rb
```diff
@@ -200,7 +200,7 @@ def run_gc
   it "should be able to connect to database with numeric-only name" do
     creds = DatabaseCredentials['numericuser']
     @client.query "CREATE DATABASE IF NOT EXISTS `#{creds['database']}`"
-    @client.query "GRANT ALL ON `#{creds['database']}`.* TO #{creds['username']}@`#{creds['host']}`"
+    @client.query "GRANT ALL ON `#{creds['database']}`.* TO #{creds['username']}@`#{creds['host']}` IDENTIFIED BY '#{creds['password']}'"
```
sodabrew: This looks good

spec/configuration.yml.example
```diff
@@ -7,11 +7,11 @@ root:
...
   username: LOCALUSERNAME
-  password:
+  password: test
   database: mysql2_test
 
 numericuser:
   host: localhost
   username: LOCALUSERNAME
-  password:
+  password: test
```
sodabrew: I need to check where the user is created; IIRC the assumption is that a $USER user already exists?

jeremy: The user is implicitly created by this `GRANT`: https://github.com/brianmario/mysql2/pull/669/files#diff-f50f287044ea76fe2dfe616c8d01c194R203

We could explicitly `CREATE USER …` in spec helper's `config.before(:all)`, but that was added in MySQL 5.0.2.

spec/configuration.yml.example
```diff
@@ -7,11 +7,11 @@ root:
 user:
   host: localhost
   username: LOCALUSERNAME
-  password:
+  password: test
```
sodabrew: The default `travis` user has no password: http://docs.travis-ci.com/user/database-setup/#MySQL

I'm not sure how the tests pass with this change?

jeremy: These config are only for specs. We only connect as `root`.

In fact, the `user` config is unused and, according to git history, has never been used!

IMO the `numericuser` config is inappropriate here too. The spec should use the same config and just change the db name to a numeric one.

jeremy: Updated to drop the unused toplevel db configs. The numeric-db spec should just create the numeric db using the main test db creds. Much simpler!

sodabrew:
Resolved by 39901daa90caf51e44cd85aedbeccd10f763fd4a

PR #670, opened 2015-09-07
Fix missing 'end' and hidden failing specs

jeremy:
Not sure how this worked before. Perhaps masked by rspec runner?

Also fixes the failing spec it exposed.


sodabrew:
Thanks, yes this was a merge error. I fixed it locally while building the gem packages and pushed to master just a moment ago cdc9321b34a82cc58fc9901c3a1bfba4f1db7a47

PR #671, opened 2015-09-07
Fix Timeout interrupt handling on Ruby 2.3

jeremy:
Timeout::ExitException was removed in Ruby 2.3.0, 2.2.3, and 2.1.8
in favor of Timeout::Error, but only 2.1.x and 2.2.x have backward-
compatibility aliases.

Along with #669 and #670, this makes tests green for me on latest Ruby trunk.

### DIFF
--- a/lib/mysql2/client.rb
+++ b/lib/mysql2/client.rb
@@ -82,8 +82,18 @@ module Mysql2
     if Thread.respond_to?(:handle_interrupt)
       require 'timeout'
 
+      # Timeout::ExitException was removed in Ruby 2.3.0, 2.2.3, and 2.1.8,
+      # but is present in earlier 2.1.x and 2.2.x, so we provide a shim.
+      #
+      # Set our own constant instead of modifying core ::Timeout.
+      TimeoutError = if defined?(::Timeout::ExitException)
+        ::Timeout::ExitException
+      else
+        ::Timeout::Error
+      end
+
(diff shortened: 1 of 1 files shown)


lib/mysql2/client.rb
```diff
@@ -82,8 +82,18 @@ def self.default_query_options
     if Thread.respond_to?(:handle_interrupt)
       require 'timeout'
 
+      # Timeout::ExitException was removed in Ruby 2.3.0, 2.2.3, and 2.1.8,
+      # but is present in earlier 2.1.x and 2.2.x, so we provide a shim.
+      #
+      # Set our own constant instead of modifying core ::Timeout.
+      TimeoutError = if defined?(::Timeout::ExitException)
```
cbandy: Is it better or worse to use a variable instead of a constant?

jeremy: Idiomatic to use a constant when it won't be reassigned and we want lexical lookup from an instance method. If this is a value we needed to configure, using a class or instance attribute would make sense.

sodabrew:
It occurs to me that we need this for Mysql2::Statement#execute as well. To avoid duplicating code, maybe we should move this up to lib/mysql2.rb in the Mysql2 or Mysql2::Util modules?

sodabrew:
Let me know if #679 looks solid.

jeremy:
:+1: !

PR #673, opened 2015-09-08
Support prepared_statement.close

jeremy:
Useful when caching prepared statements. Evicted statements can release
server resources immediately rather than waiting for Ruby GC.

### DIFF
--- a/ext/mysql2/statement.c
+++ b/ext/mysql2/statement.c
@@ -17,6 +17,15 @@ static void rb_mysql_stmt_mark(void * ptr) {
   rb_gc_mark(stmt_wrapper->client);
 }
 
+static void *nogvl_stmt_close(void *ptr) {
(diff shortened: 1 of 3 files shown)


sodabrew:
Cool! Can be rebased now since that spec is fixed.

sodabrew:
Ping for rebase :)

jeremy:
Done!

sodabrew:
Looks like closing the statement handle is causing some knock-on memory allocation effects, I'm digging into this.

sodabrew:
Based on the comments on https://dev.mysql.com/doc/refman/5.0/en/mysql-stmt-execute.html it appears that attempting to use a statement handle after you've freed it will segfault the client library instead of returning a clean error. Ideas:

* Add checks before all other mysql_stmt_* methods that the `stmt_wrapper->closed` is zero.
* Set `stmt_wrapper->stmt = NULL` right after calling mysql_stmt_close on it.

PR #677, opened 2015-09-08
Replace `Timeout::ExitException` with `Timeout::Error`

yui-knk:
ruby 2.3.0 will not support `Timeout::ExitException`,
so replace it.

https://github.com/ruby/ruby/commit/0f663b244944c5eb2b6c7d0862c15c377747df05


sodabrew:
Does this work in earlier versions of Ruby? See also #671

yui-knk:
Sorry, I missed #671 , I will close this and discuss on #671.

PR #679, opened 2015-09-09
Fix Timeout interrupt handling on Ruby 2.3 and protect Mysql2::Statement#execute

sodabrew:
Timeout::ExitException was removed in Ruby 2.3.0, 2.2.3, and 2.1.8, in favor of Timeout::Error. Backwards compatible aliases are provided for Ruby 2.1.x and 2.2.x, but not earlier verions.

With thanks to @jeremy for PR #671 and @yui-knk for PR #677, this commit also protects prepared statements from being interrupted, so the compat shim is in Mysql2::Util.

### DIFF
--- a/ext/mysql2/statement.c
+++ b/ext/mysql2/statement.c
@@ -470,7 +470,7 @@ void init_mysql2_statement() {
 
   rb_define_method(cMysql2Statement, "param_count", param_count, 0);
   rb_define_method(cMysql2Statement, "field_count", field_count, 0);
-  rb_define_method(cMysql2Statement, "execute", execute, -1);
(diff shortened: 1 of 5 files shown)


sodabrew:
This is causing Ruby 1.9.3 to sometimes segfault right after the execute is timed out, which is probably because 1.9.3 isn't actually protecting against the timeout (no Thread#handle_interrupt) and once the statement handle is invalid mysqlclient is fragile (as seen while debugging #673). I would solve this by actually skipping that test if Thread#handle_interrupt is not available and commenting about it.

This is also causing Rubinius to segfault at the _end_ of all of the tests, and I'm less sure of what's going on there.

PR #680, opened 2015-09-10
fix c-warnings; fix CI

tamird:
@sodabrew this should fix travis

### DIFF
--- a/ext/mysql2/statement.h
+++ b/ext/mysql2/statement.h
@@ -13,6 +13,6 @@ void init_mysql2_statement(void);
 void decr_mysql2_stmt(mysql_stmt_wrapper *stmt_wrapper);
 
 VALUE rb_mysql_stmt_new(VALUE rb_client, VALUE sql);
-void rb_raise_mysql2_stmt_error(mysql_stmt_wrapper *stmt_wrapper);
+void rb_raise_mysql2_stmt_error(mysql_stmt_wrapper *stmt_wrapper) RB_MYSQL_NORETURN;
 
 #endif
--- a/ext/mysql2/client.c
+++ b/ext/mysql2/client.c
@@ -129,7 +129,6 @@ static VALUE rb_raise_mysql2_error(mysql_client_wrapper *wrapper) {
                  UINT2NUM(mysql_errno(wrapper->client)),
                  rb_sql_state);
   rb_exc_raise(e);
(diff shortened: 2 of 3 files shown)


tamird:
travis failure is our old friend `dlopen(e, 9): image not found - e1`

ext/mysql2/statement.h
```diff
@@ -13,6 +13,6 @@ void init_mysql2_statement(void);
 void decr_mysql2_stmt(mysql_stmt_wrapper *stmt_wrapper);
 
 VALUE rb_mysql_stmt_new(VALUE rb_client, VALUE sql);
-void rb_raise_mysql2_stmt_error(mysql_stmt_wrapper *stmt_wrapper);
+void rb_raise_mysql2_stmt_error(mysql_stmt_wrapper *stmt_wrapper) __attribute__ ((noreturn));
```
sodabrew: Should be wrapped in a "GCC-ism" macro?

tamird: aren't unknown attributes ignored by compilers that don't support them? what did the build do on appveyor?

sodabrew:
Both Travis and AppVeyor built master fine today. Where did you see these warnings for build failures?

tamird:
The travis badge on the readme is red, and here's the build for current HEAD https://travis-ci.org/brianmario/mysql2/builds/79690310

sodabrew:
Right - that's what I mean. The failures are due to specs failing in the already-known ways, not compile failures.

tamird:
Looks like a compile error to me:
https://travis-ci.org/brianmario/mysql2/jobs/79690325
https://travis-ci.org/brianmario/mysql2/jobs/79690328

sodabrew:
I need to enroll in school.
cantreadgood

ext/mysql2/statement.h
sodabrew: Unknown attributes are no longer ignored in GCC https://gcc.gnu.org/bugzilla/show_bug.cgi?id=55768  -- could be true for other compilers that don't ignore and don't know about this one. There's something similar for the UNUSED macro: https://github.com/brianmario/mysql2/blob/master/ext/mysql2/mysql2_ext.h#L31-L35

tamird: done

PR #683, opened 2015-09-16
Implement easy flags array

### LINKED ISSUE #649: Make it easier to pass flags from database.yml
I was just thinking about #526 and how it should be easier to set the connection flags. I found a guide on the web that shows how to find and set the integer value of the flags: http://ursdiganta.blogspot.com/2011/07/mysql2-and-stored-procedures-in-rails-3.html

> Create mysql2_adapter.rb file in config/initializers with the following content:
``` ruby
module ActiveRecord
  class Base
    # Overriding ActiveRecord::Base.mysql2_connection
    # method to allow passing options from database.yml
    #
    # Example of database.yml
    #
    #   login: &login
    #     socket: /tmp/mysql.sock
    #     adapter: mysql2
    #     host: localhost
    #     encoding: utf8
    #     flags: 131072
    #
    # @param [Hash] config hash that you define in your
    #   database.yml
    # @return [Mysql2Adapter] new MySQL adapter object
    #
    def self.mysql2_connection(config)
      config[:username] = 'root' if config[:username].nil?
      if Mysql2::Client.const_defined? :FOUND_ROWS
        config[:flags] = config[:flags] ? config[:flags] | Mysql2::Client::FOUND_ROWS : Mysql2::Client::FOUND_ROWS
      end
      client = Mysql2::Client.new(config.symbolize_keys)
      options = [config[:host], config[:username], config[:password], config[:database], config[:port], config[:socket], 0]
      ConnectionAdapters::Mysql2Adapter.new(client, logger, options, config)
    end
  end
end
```
> So now you can pass any additional options from your database.yml. See that 131072 ? This is the value of CLIENT_MULTI_RESULTS constant. 

Well that sucks! It should be easier! I found this gem: https://github.com/zerobearing2/activerecord-mysql-adapter-flags

>config/database.yml
``` yaml
  development:
    adapter: mysql
    username: dbuser
    password: "secert"
    host: localhost
    database: test
    flags: "CLIENT_MULTI_RESULTS, CLIENT_COMPRESS"
```

> Note: the flags get OR'd together before passing onto the connect method. If the flag const doesn't exist in the Mysql class, it will be ignored.

I think a "mini-DSL" like this would be pretty nice to provide.

sodabrew:
Resolves #649

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -8,15 +8,19 @@
 
 # Offense count: 2
 Metrics/AbcSize:
-  Max: 68
+  Max: 80
 
 # Offense count: 1
 Metrics/BlockNesting:
   Max: 5
 
+# Offense count: 1
+Metrics/ClassLength:
+  Max: 125
+
 # Offense count: 2
 Metrics/CyclomaticComplexity:
-  Max: 23
+  Max: 25
 
 # Offense count: 290
 # Configuration parameters: AllowURI, URISchemes.
@@ -26,7 +30,7 @@ Metrics/LineLength:
 # Offense count: 5
 # Configuration parameters: CountComments.
 Metrics/MethodLength:
-  Max: 43
+  Max: 50
 
 # Offense count: 1
 Metrics/PerceivedComplexity:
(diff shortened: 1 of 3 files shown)


sodabrew:
Helpful debugging warn will go away in the next rebase:
```
  should parse flags array
Adding MySQL flag: FOUND_ROWS
Removing MySQL flag: !PROTOCOL_41
```

Gotta convince RuboCop not to complain about 107 lines in the class. Bleh.

cbandy:
> Gotta convince RuboCop not to complain about 107 lines in the class. Bleh.

Could disable RuboCop. :wink:

Issue #606, opened 2015-04-02
mysql2 gem improperly closing MySQL connections


dtest:
The mysql2 gem doesn't properly close the connection, even when explicitly calling the 'client.close' method, resulting in increase of 'Aborted_clients' counter indicating an improper closure of the connection. My setup is MySQL 5.6.23 installed from MySQL's ubuntu repo on an Ubuntu Trusty EC2 instance, mysql2 gem v0.3.18, ruby 2.0.0p353 (embedded sensu version), libmysqlclient18

```
dtest@ip-10-1-1-73:~# /opt/sensu/embedded/bin/ruby --version
ruby 2.0.0p353 (2013-11-22 revision 43784) [x86_64-linux]
dtest@ip-10-1-1-73:~# /opt/sensu/embedded/bin/gem list | grep mysql
mysql (2.9.1)
mysql2 (0.3.18)
dtest@ip-10-1-1-73:~# ldd /opt/sensu/embedded/lib/ruby/gems/2.0.0/gems/mysql2-0.3.18/lib/mysql2/mysql2.so
  linux-vdso.so.1 =>  (0x00007fffa07fe000)
  libruby.so.2.0 => /opt/sensu/embedded/lib/libruby.so.2.0 (0x00007f8bde4ec000)
  libmysqlclient.so.18 => /usr/lib/x86_64-linux-gnu/libmysqlclient.so.18 (0x00007f8bddf35000)
  libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007f8bddd10000)
  libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f8bdd94b000)
  librt.so.1 => /lib/x86_64-linux-gnu/librt.so.1 (0x00007f8bdd743000)
  libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007f8bdd53e000)
  libcrypt.so.1 => /lib/x86_64-linux-gnu/libcrypt.so.1 (0x00007f8bdd305000)
  libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007f8bdcfff000)
  /lib64/ld-linux-x86-64.so.2 (0x00007f8bdeba9000)
  libz.so.1 => /opt/sensu/embedded/lib/libz.so.1 (0x00007f8bdcde1000)
[...]

PR #684, opened 2015-09-25
Close connection during garbage collection

cbandy:
Restores the behavior of v0.3.15 which sends `QUIT` during garbage collection. Keep the behavior in #463 behind a new connection option, `automatic_close`.

Relates to #606.

### DIFF
--- a/ext/mysql2/client.c
+++ b/ext/mysql2/client.c
@@ -260,11 +260,7 @@ static VALUE allocate(VALUE klass) {
   obj = Data_Make_Struct(klass, mysql_client_wrapper, rb_mysql_client_mark, rb_mysql_client_free, wrapper);
   wrapper->encoding = Qnil;
   MARK_CONN_INACTIVE(self);
-#ifndef _WIN32
-  wrapper->automatic_close = 0;
-#else
   wrapper->automatic_close = 1;
-#endif
   wrapper->server_version = 0;
   wrapper->reconnect_enabled = 0;
   wrapper->connect_timeout = 0;
@@ -386,8 +382,10 @@ static VALUE rb_connect(VALUE self, VALUE user, VALUE pass, VALUE host, VALUE po
 }
 
 /*
- * Terminate the connection; call this when the connection is no longer needed.
- * To have the garbage collector close the connection, enable +automatic_close+.
+ * Immediately disconnect from the server; normally the garbage collector
+ * will disconnect automatically when a connection is no longer needed.
+ * Explicitly closing this will free up server resources sooner than waiting
+ * for the garbage collector.
  *
  * @return [nil]
  */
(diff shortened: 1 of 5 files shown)


lib/mysql2/client.rb
```diff
@@ -30,7 +30,7 @@ def initialize(opts = {})
       opts[:connect_timeout] = 120 unless opts.key?(:connect_timeout)
 
       # TODO: stricter validation rather than silent massaging
-      [:reconnect, :connect_timeout, :local_infile, :read_timeout, :write_timeout, :default_file, :default_group, :secure_auth, :init_command].each do |key|
+      [:reconnect, :connect_timeout, :local_infile, :read_timeout, :write_timeout, :default_file, :default_group, :secure_auth, :init_command, :automatic_close].each do |key|
         next unless opts.key?(key)
         case key
         when :reconnect, :local_infile, :secure_auth
```
sodabrew: Add `:automatic_close` to the list of booleans.

cbandy: Will do. `automatic_close=` already does this using `RTEST`.

cbandy: Done.

ext/mysql2/client.c
```diff
@@ -1086,6 +1086,41 @@ static VALUE rb_mysql_client_encoding(VALUE self) {
...
+ * Set this to +false+ to leave the connection open after it is garbage
+ * collected. To avoid "Aborted connection" errors on the server, explicitly
+ * call +close+ when the connection is no longer needed.
+ *
+ * @see http://dev.mysql.com/doc/en/communication-errors.html
+ */
+static VALUE set_automatic_close(VALUE self, VALUE value) {
+  GET_CLIENT(self);
+  if (RTEST(value)) {
+    value = Qtrue;
```
sodabrew: I would avoid re-using the variable `value`, since we're not modifying the argument and shouldn't give the impression that we might be trying to do that.

ext/mysql2/client.c
```diff
@@ -1086,6 +1086,41 @@ static VALUE rb_mysql_client_encoding(VALUE self) {
...
+    wrapper->automatic_close = 1;
+  } else {
+#ifndef _WIN32
+    value = Qfalse;
+    wrapper->automatic_close = 0;
+#else
+    rb_raise(cMysql2Error, "Connections are always closed by garbage collector on Windows");
```
sodabrew: I would rb_warn here, rather than rb_raise. I'm imagining an app that comes with this setting baked in, and then someone wants to deploy on Windows, but ends up blocked until the code is updated.

sodabrew:
Thanks for your patience on this! I'm collecting the todo list for 0.4.3 and I'm keen to include this PR!

ext/mysql2/client.c
cbandy: Since the return value from assignment is usually ignored, I'll change this to return the input unmodified value.

ext/mysql2/client.c
cbandy: It looks like testing warnings is a little tricky. Any idea how to change L199 in client_spec?

ext/mysql2/client.c
cbandy: Done.

ext/mysql2/client.c
cbandy: Done.

cbandy:
The testing of `$stderr` works in Linux, but I don't have a Windows machine or build server to try it on. Do you?

sodabrew:
I do have a windows vm, I'll give it a spin!

PR #692, opened 2015-10-18
Support result.free

### LINKED ISSUE #690: stmt.close cause segv
* ruby 2.2.3p173 (2015-08-18 revision 51636) [x86_64-darwin14]
* OS X 10.10.5

```ruby
require 'mysql2'

client = Mysql2::Client.new(
  host: 'localhost',
  username: 'root',
  database: 'test',
)

client.query("DROP TABLE IF EXISTS t")
client.query("CREATE TABLE t (a int)")

stmt = client.prepare('SELECT * FROM t WHERE a = ?')
stmt.execute(1)

client.query("ALTER TABLE t ADD COLUMN b int")

# begin
#   stmt.execute(1)
# rescue
#   # <Mysql2::Error: The number of columns in the result set differs from the number of bound buffers. You must reset the statement, rebind the result set columns, and execute the statement again>
#   p $!
# end

stmt.close
# Segmentation fault
GC.start
```

kamipo:
Fixes #690.

When calling `mysql_stmt_execute`, we should not call `mysql_stmt_close`
before calling `mysql_stmt_free_result`.

Example:

```ruby
require 'mysql2'

client = Mysql2::Client.new(
  host: 'localhost',
  username: 'root',
  database: 'mysql',
)

stmt = client.prepare('SELECT 1 FROM dual WHERE 0 = ?')
stmt.execute(1).tap { |result|
  # should call result.free before calling stmt.close
  result.free if result.respond_to?(:free)
}

stmt.close
# Segmentation fault
GC.start
```

### DIFF
--- a/ext/mysql2/result.c
+++ b/ext/mysql2/result.c
@@ -48,7 +48,7 @@ static rb_encoding *binaryEncoding;
 #define MYSQL2_MIN_TIME 62171150401ULL
 #endif
 
-#define GET_RESULT(obj) \
+#define GET_RESULT(self) \
   mysql2_result_wrapper *wrapper; \
   Data_Get_Struct(self, mysql2_result_wrapper, wrapper);
 
@@ -86,7 +86,7 @@ static void rb_mysql_result_mark(void * wrapper) {
 }
 
 /* this may be called manually or during GC */
-static void rb_mysql_result_free_result(mysql2_result_wrapper * wrapper) {
+static void rb_mysql_result_free_result_(mysql2_result_wrapper * wrapper) {
   if (!wrapper) return;
 
   if (wrapper->resultFreed != 1) {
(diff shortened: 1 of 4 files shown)


sodabrew:
I don't want to put the user in a position that a failure to call `result.free` will crash. How about keeping track of the most recent Result in the Statement object, and if a new statement.execute or Statement object is GCd, then it'll free the last Result at that time.

ext/mysql2/result.c
```diff
@@ -48,7 +48,7 @@ static rb_encoding *binaryEncoding;
 #define MYSQL2_MIN_TIME 62171150401ULL
 #endif
 
-#define GET_RESULT(obj) \
+#define GET_RESULT(self) \
```
sodabrew: Oh, nice catch on this.

sodabrew:
Closing in favor of #697 - but if Result#free turns out to be required for the ActiveRecord adapter, then we'll follow up on this.

sodabrew:
I was wrong, I think we need this. Could you rebase the commits up to current master?

PR #693, opened 2015-10-18
Support JSON with prepared statement

kamipo:


ext/mysql2/result.c
```diff
@@ -709,7 +715,10 @@ static VALUE rb_mysql_result_fetch_row(VALUE self, MYSQL_FIELD * fields, const r
         case MYSQL_TYPE_STRING:     /* CHAR or BINARY field */
         case MYSQL_TYPE_SET:        /* SET field */
         case MYSQL_TYPE_ENUM:       /* ENUM field */
-        case MYSQL_TYPE_GEOMETRY:   /* Spatial fielda */
+        case MYSQL_TYPE_GEOMETRY:   /* Spatial field */
+#ifdef HAVE_MYSQL_TYPE_JSON
+        case MYSQL_TYPE_JSON:       /* JSON field */
+#endif
         default:
```
sodabrew: If the type is unknown, it will fall through to the default, so this isn't strictly necessary

ext/mysql2/result.c
```diff
@@ -491,6 +494,9 @@ static VALUE rb_mysql_result_fetch_row_stmt(VALUE self, MYSQL_FIELD * fields, co
         case MYSQL_TYPE_SET:          // char[]
         case MYSQL_TYPE_ENUM:         // char[]
         case MYSQL_TYPE_GEOMETRY:     // char[]
+#ifdef HAVE_MYSQL_TYPE_JSON
+        case MYSQL_TYPE_JSON:         // char[]
+#endif
```
sodabrew: I notice there's no default here. Maybe there should be?

ext/mysql2/mysql2_ext.h
```diff
@@ -12,15 +12,21 @@ void Init_mysql2(void);
...
 #include <mysql/mysql.h>
 #include <mysql/mysql_com.h>
+#include <mysql/mysql_version.h>
 #include <mysql/errmsg.h>
 #include <mysql/mysqld_error.h>
 #endif
 
+#if 50708 <= MYSQL_VERSION_ID && MYSQL_VERSION_ID < 100000
+#define HAVE_MYSQL_TYPE_JSON
+#endif
```
sodabrew: Rather than keying off of a version check, we can either check if the symbol is defined in extconf.rb, or even simpler, `#ifdef MYSQL_TYPE_JSON`

kamipo:
I fixed to simpler implementation. I think that there is no need to be an error unsupported buffer type in result_buffers. It is because if a result includes unsupported buffer type, `mysql_stmt_store_result` causes an error "Using unsupported buffer type: 245" (245 is `MYSQL_TYPE_JSON`). 

https://github.com/mysql/mysql-server/blob/5.7/libbinlogevents/export/binary_log_types.h#L51

https://github.com/kamipo/mysql2/blob/support_type_json/ext/mysql2/statement.c#L363-L366

```ruby
    if (rb_thread_call_without_gvl(nogvl_stmt_store_result, stmt, RUBY_UBF_IO, 0) == Qfalse) {
      mysql_free_result(metadata);
      rb_raise_mysql2_stmt_error(stmt_wrapper);
    }
```

PR #695, opened 2015-11-05
extconf: pass `-lasan` when using address sanitizer

### LINKED ISSUE #685: __asan_option_detect_stack_use_after_return on docker ci build since 0.4.1
I'm trying to use version 0.4.1 in my docker ci instance (ubuntu:vivid image). It builds fine but doesn't work because it ends with this error:

```
/cache/gems/stellenticket/ubuntu-ci_latest/ruby/2.1.0/gems/activesupport-4.2.4/lib/active_support/dependencies.rb:274:in `require': /cache/gems/stellenticket/ubuntu-ci_latest/ruby/2.1.0/extensions/x86_64-linux/2.1.0/mysql2-0.4.1/mysql2/mysql2.so: undefined symbol: __asan_option_detect_stack_use_after_return - /cache/gems/stellenticket/ubuntu-ci_latest/ruby/2.1.0/extensions/x86_64-linux/2.1.0/mysql2-0.4.1/mysql2/mysql2.so (LoadError)
```

Looks like `libasan` was not linked correctly?

The error does not happen with version 0.3.20.

tamird:
Fixes #685. See google/sanitizers#380.

[Review on Reviewable](https://reviewable.io/reviews/brianmario/mysql2/695)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,7 @@ os:
 addons:
   hosts:
     - mysql2gem.example.com
+env: MYSQL2_CI=1
 rvm:
   - 1.8.7
   - 1.9.3
@@ -21,24 +22,24 @@ rvm:
 matrix:
   include:
     - rvm: 2.0.0
-      env: DB=mariadb55
+      env: MYSQL2_CI=1 DB=mariadb55
       addons:
          mariadb: 5.5
     - rvm: 2.0.0
-      env: DB=mariadb10.0
+      env: MYSQL2_CI=1 DB=mariadb10.0
       addons:
          mariadb: 10.0
     - rvm: 2.0.0
-      env: DB=mariadb10.1
+      env: MYSQL2_CI=1 DB=mariadb10.1
       addons:
          mariadb: 10.1
     - rvm: 2.0.0
-      env: DB=mysql57
+      env: MYSQL2_CI=1 DB=mysql57
     - rvm: 2.0.0
-      env: DB=mysql
+      env: MYSQL2_CI=1 DB=mysql
       os: osx
     - rvm: 2.0.0
-      env: DB=mysql55
+      env: MYSQL2_CI=1 DB=mysql55
       os: osx
   allow_failures:
     - rvm: ruby-head
--- a/ext/mysql2/extconf.rb
+++ b/ext/mysql2/extconf.rb
@@ -117,7 +117,7 @@ wishlist = [
 if ENV['MYSQL2_CI']
   wishlist += [
     '-Werror',
-    '-fsanitize=address',
+    '-fsanitize=address -lasan',
     '-fsanitize=cfi',
     '-fsanitize=integer',
     '-fsanitize=memory',
--- a/ext/mysql2/extconf.rb
+++ b/ext/mysql2/extconf.rb
@@ -127,7 +127,7 @@ if ENV['MYSQL2_CI']
 end
 
 usable_flags = wishlist.select do |flag|
-  try_link('int main() {return 0;}', flag)
+  try_link('int main() {return 0;}', "-Werror -Wunknown-warning-option #{flag}")
 end
 
 $CFLAGS << ' ' << usable_flags.join(' ')
--- a/ext/mysql2/extconf.rb
+++ b/ext/mysql2/extconf.rb
@@ -105,14 +105,16 @@ wishlist = [
   '-Wno-missing-field-initializers', # gperf generates bad code
   '-Wno-missing-variable-declarations', # missing symbols due to ruby native ext initialization
   '-Wno-padded', # mysql :(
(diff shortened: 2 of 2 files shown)


ext/mysql2/extconf.rb
```diff
@@ -105,17 +105,19 @@ def asplode(lib)
...
   '-Wno-undef', # rubinius :(
+  '-Wno-unreachable-code', # rubby :(
   '-Wno-used-but-marked-unused', # rubby :(
 ]
 
 if ENV['CI']
   wishlist += [
     '-Werror',
-    '-fsanitize=address',
+    '-fsanitize=address -lasan',
```
brianmario: This could be needed because it's a "plugin" to the compiler chain on some (all?) systems. Though, I've definitely used `-fsanitize=address` on OSX and not had to link to `libasan` myself so I'm not really sure...

tamird: It seems like an issue with newer versions of the clang toolchain. Either way, it doesn't appear to be documented explicitly anywhere.

ext/mysql2/extconf.rb
```diff
@@ -105,17 +105,19 @@ def asplode(lib)
   '-Wno-missing-field-initializers', # gperf generates bad code
   '-Wno-missing-variable-declarations', # missing symbols due to ruby native ext initialization
   '-Wno-padded', # mysql :(
+  '-Wno-reserved-id-macro', # rubby :(
```
sodabrew: This causes the OS X build to fail

tamird: Yes, pushed a commit.

ext/mysql2/extconf.rb
```diff
@@ -125,7 +127,7 @@ def asplode(lib)
 end
 
 usable_flags = wishlist.select do |flag|
-  try_link('int main() {return 0;}', flag)
+  try_link('int main() {return 0;}', "-Werror -Wunknown-warning-option #{flag}")
```
sodabrew: Clever! In the event that `-Wunknown-warning-option` itself is unsupported, it should reject all of the flags.

tamird: Yeah, that's true. It'd be good to surface that, but I think this is reasonable for now. It seems that there's some divergence between clang and gcc on this, in that gcc's version of this flag is or was `-Wunknown-warning` - I don't have gcc installed, so I can't check, but something to be aware of.

tamird:
More failures in travis, but they appear to be old known flakes and not related to the changes in this PR.

sodabrew:
I re-ran the tests, we're good. Would you like to squash to 1 or 2 commits?

tamird:
I'd leave it as is; each commit is addressing a separate problem. I can squash if you feel otherwise?

sodabrew:
And also, thank you @tamird for jumping right onto the thread in #685 with this PR! Really appreciate your quick response to help sort this out.

tamird:
@sodabrew no worries! I've updated cc12d23 as you suggested. Let me know if you'd like to make additional changes before landing this.

.travis.yml
```diff
@@ -8,6 +8,10 @@ os:
 addons:
   hosts:
     - mysql2gem.example.com
+
+env:
+  - MYSQL2_CI=1
+
```
sodabrew: Extra whitespace lines not needed (the rest of the file is pretty dense).

tamird: Fixed

.travis.yml
```diff
@@ -8,6 +8,8 @@ os:
 addons:
   hosts:
     - mysql2gem.example.com
+env:
+  - MYSQL2_CI=1
```
sodabrew: Looks like the alternate-DB envs in the matrix below are removing this. How about:
```
env:
  global:
    - MYSQL2_CI=1
```

tamird: alright, here it goes

sodabrew:
Adding `-lasan` to CFLAGS doesn't work - On Ubuntu Trusty, adding `-lasan` to `$LDFLAGS` does the trick. On Mac OS X 10.10, adding `-lasan` to `$LDFLAGS` did not work.

Some extra environment variables are also needed to make this really useful (line numbers) http://tsdgeos.blogspot.com/2014/03/asan-and-gcc-how-to-get-line-numbers-in.html

However, I am getting address sanitizer crashes in places that appear correct on inspection. I'm not convinced that this tool is going to pan out here :/ I am inclined to remove this block rather than trying to fix it right now. We can always revisit at a later date.

tamird:
@sodabrew your call. This change already tightens the guarantee that this won't break outside of CI.

Can you share some of the details of how you got those crashes?

sodabrew:
Closing in favor of #703.

PR #697, opened 2015-11-18
Avoid crashing when Statement#close is called before a Result is collected

### LINKED ISSUE #690: stmt.close cause segv

sodabrew:
Alternative to #692 / Resolves #690 

I found that you can call `mysql_stmt_close` first, without calling `mysql_stmt_result_free`, however you must not later call `mysql_stmt_result_free`.

### DIFF
--- a/ext/mysql2/client.c
+++ b/ext/mysql2/client.c
@@ -133,7 +133,7 @@ static VALUE rb_raise_mysql2_error(mysql_client_wrapper *wrapper) {
 
 static void *nogvl_init(void *ptr) {
   MYSQL *client;
-  mysql_client_wrapper *wrapper = (mysql_client_wrapper *)ptr;
+  mysql_client_wrapper *wrapper = ptr;
 
   /* may initialize embedded server and read /etc/services off disk */
   client = mysql_init(wrapper->client);
@@ -224,7 +224,7 @@ static void *nogvl_close(void *ptr) {
 
 /* this is called during GC */
 static void rb_mysql_client_free(void *ptr) {
-  mysql_client_wrapper *wrapper = (mysql_client_wrapper *)ptr;
+  mysql_client_wrapper *wrapper = ptr;
   decr_mysql2_client(wrapper);
 }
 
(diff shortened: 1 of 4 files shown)


sodabrew:
@kamipo What do you think of doing it this way, rather than providing an explicit Result#free that the user must call?

kamipo:
I think it is better to call immediately `mysql_stmt_free_result` to a result set is produced by `mysql_stmt_store_result`. But actually this is Rubyish way, it looks good to me.
I PRed https://github.com/rails/rails/pull/22352 working on this branch. It looks like works fine. Thanks!

sodabrew:
@kamipo I asked on that Rails PR, and should probably ask here for continuity, if you require the Mysql2::Result#free method in order to hook this into the ActiveRecord adapter, or if allowing Ruby GC to handle the result free is sufficient?

kamipo:
I realized that when `if (!is_streaming)`, `Mysql2::Result#each` is called in `Statement#execute`, then `rb_mysql_result_free_result` is called immediately. It seems good to me. `Mysql2::Result#free` is not needed for now.

https://github.com/sodabrew/mysql2/blob/ed5613d26b1f54e5eb0629a06cb5136de4ae47cf/ext/mysql2/statement.c#L374-L377
```ruby
  if (!is_streaming) {
    // cache all result
    rb_funcall(resultObj, intern_each, 0);
  }
```

https://github.com/sodabrew/mysql2/blob/ed5613d26b1f54e5eb0629a06cb5136de4ae47cf/ext/mysql2/result.c#L840-L843
```ruby
      if (wrapper->lastRowProcessed == wrapper->numberOfRows) {
        /* we don't need the mysql C dataset around anymore, peace it */
      }
```

PR #663, opened 2015-09-05
Terminate connections correctly when calling close

cbandy:
Fixes #606

PR #702, opened 2015-11-20
fix memory leak on connection close

### LINKED ISSUE #701: memory leak
Hi, it seems like there is a memory leak in 0.4.1.

Here is a script to reproduce it:

```ruby
require 'mysql2'
require 'objspace'
require 'tempfile'

def check
  client = Mysql2::Client.new(host: 'localhost', port: 3306, username: 'root', password: nil)
  results = client.query("SHOW STATUS LIKE 'wsrep_local_state'")
  ro = client.query("SHOW VARIABLES LIKE 'read_only'").first

  results.count
  client.close
end

count = ARGV[0]

if count
  count.to_i.times { check }
  GC.start
  GC.start
  if ARGV[1]
    puts "ready: " + $$.to_s
    sleep
  end

  dump = ObjectSpace.dump_all
  dump.close
  puts dump.path
else
  loop { check }
end
```

If I run the program like this, it will grow indefinitely:

```
$ ruby -I lib test.rb
```

If I provide a count and dump the heap, I don't see any changes in Ruby object counts:

```
[aaron@TC mysql2 (master)]$ ruby -Ilib:test test.rb 1000 | xargs cat | cut -d' ' -f2 | sort | uniq -c | sort -n
   1 "root":"global_list",
   1 "root":"global_tbl",
   1 "root":"machine_context",
   1 "root":"vm",
   1 "type":"COMPLEX",
   1 "type":"RATIONAL",
   2 "type":"BIGNUM",
   3 "type":"STRUCT",
   4 "type":"FILE",
   6 "type":"FLOAT",
  39 "type":"MODULE",
  60 "type":"ICLASS",
  78 "type":"REGEXP",
  98 "type":"NODE",
 494 "type":"DATA",
 587 "type":"CLASS",
 854 "type":"HASH",
2648 "type":"OBJECT",
5051 "type":"ARRAY",
5303 "type":"IMEMO",
16957 "type":"STRING",
[aaron@TC mysql2 (master)]$ ruby -Ilib:test test.rb 2000 | xargs cat | cut -d' ' -f2 | sort | uniq -c | sort -n
   1 "root":"global_list",
   1 "root":"global_tbl",
   1 "root":"machine_context",
   1 "root":"vm",
   1 "type":"COMPLEX",
   1 "type":"RATIONAL",
   2 "type":"BIGNUM",
   3 "type":"STRUCT",
   4 "type":"FILE",
   6 "type":"FLOAT",
  39 "type":"MODULE",
  60 "type":"ICLASS",
  78 "type":"REGEXP",
  98 "type":"NODE",
 494 "type":"DATA",
 587 "type":"CLASS",
 854 "type":"HASH",
2648 "type":"OBJECT",
5051 "type":"ARRAY",
5303 "type":"IMEMO",
16958 "type":"STRING",
```

If this were a Ruby object leak, I would expect that the heap would change when I change the number of iterations.  But it doesn't, which leads me to believe that this is a Genuine Memory Leak™.

If I put the program to sleep and use `leaks` on OS X, it looks like all the memory being leaked is allocated [here](https://github.com/brianmario/mysql2/blob/72c1074d3ec243fb8a23346dca84ca8f06886576/ext/mysql2/client.c#L269).  I suspect that the reference counting in the free function isn't working correctly.

To get the leak information, I ran the test program like this (this is OS X specific):

```
$ MallocStackLogging=1 ruby -Ilib:test test.rb 2000 sleep
```

Then when the program said "ready", in a different terminal I did this (with $PID replaced with the PID):

```
sudo leaks $PID > leaks.txt
```

`leaks.txt` contained on the memory that has been allocated and is no longer referenced, along with the stack trace for the allocation.  Here is an example:

```
        Call stack: [thread 0x7fff7b506000]: | start | main | ruby_run_node | ruby_exec_node | ruby_exec_internal | rb_iseq_eval_main | vm_exec | vm_exec_core | vm_call_general | vm_call_method | vm_call_method_each_type | vm_call_cfunc | vm_call_cfunc_with_frame | call_cfunc_0 | int_dotimes | rb_yield_1 | rb_yield_0 | vm_yield | invoke_block_from_c_splattable | invoke_block_from_c_0 | invoke_block | vm_exec | vm_exec_core | vm_call_cfunc | vm_call_cfunc_with_frame | call_cfunc_m1 | rb_class_new_instance | rb_obj_alloc | allocate client.c:269 | ruby_xmalloc | objspace_xmalloc | malloc | malloc_zone_malloc 
Leak: 0x7fa054001800  size=1536  zone: DefaultMallocZone_0x110516000
        0x00000000 0x00000000 0x00000000 0x00000000     ................
        0x54072000 0x00007fa0 0x54070000 0x00007fa0     . .T.......T....
        0x54070000 0x00007fa0 0x0000000f 0x00000000     ...T............
        0x00000000 0x00000000 0x00000000 0x00000000     ................
        0x00000000 0x00000000 0x00000000 0x00000000     ................
        0x00002000 0x00000000 0x40000000 0x00000000     . .........@....
        0x00000001 0x00000000 0x01e13380 0x01e13380     .........3...3..
        0x00000001 0x00000000 0x00000000 0x00000000     ................
        ...
```

The leak report says it found about 2k leaks which matches up with the number I passed in.

tenderlove:
It looks like when you call `close` on a connection, it closes the
connection but doesn't free the memory allocated for the connection.
Since the pointed gets nulled out, the GC free function doesn't have a
pointer to free.

Fixes #701 

I suspect there is still a possible race condition with the refcount thing, but I can't prove it yet.  However, this seems to plug the hole for #701.

### DIFF
--- a/ext/mysql2/client.c
+++ b/ext/mysql2/client.c
@@ -213,6 +213,7 @@ static void *nogvl_close(void *ptr) {
 
   if (wrapper->client) {
     mysql_close(wrapper->client);
+    xfree(wrapper->client);
     wrapper->client = NULL;
     wrapper->connected = 0;
     wrapper->active_thread = Qnil;


sodabrew:
Thank you for debugging this! And the `xfree(wrapper->client)` at line 252 can be removed, since it will always have be NULL after `nogvl_close` returns. I missed this in reviewing #663.

brianmario:
Thanks <3

PR #703, opened 2015-11-24
Add new option --with-sanitize[=list,of,clang,sanitizers]

### LINKED ISSUE #698: [0.4.1] The gem is not correctly compiled with GCC 4.8.4
The installation of the gem is fine, however when it's put to use by some DB task, the following error is shown.

```bash
rake aborted!
LoadError: ./vendor/bundle/ruby/2.1.0/gems/mysql2-0.4.1/lib/mysql2/mysql2.so: undefined symbol: __asan_unregister_globals - /home/runner/productive/vendor/bundle/ruby/2.1.0/gems/mysql2-0.4.1/lib/mysql2/mysql2.so
```

## Reproducing the error
```bash
~ $ gcc --version
gcc (Ubuntu 4.8.4-2ubuntu1~14.04) 4.8.4
# clone a sample Rails 3 project
~ $ git clone https://github.com/thinreports/thinreports-rails3-example.git && cd thinreports-rails3-example
# add the mysql2 (0.4.1) gem to the Gemfile
# create ./config/database.yml and configure it to use the mysql2 gem
# install the gems
thinreports-rails3-example $ bundle install --path vendor/bundle
# run any DB task with rake
thinreports-rails3-example $ bundle exec rake db:drop
# !!! the error is shown here
```

## Mitigating the issue
Older versions of the gem don't have this issue. To mitigate it, an older GCC has to be used.

```
$ sudo apt-get install -y gcc-4.7 g++-4.7
$ sudo update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-4.7 100 --slave /usr/bin/g++ g++ /usr/bin/g++-4.7
$ sudo update-alternatives --auto gcc
$ gcc --version
gcc (Ubuntu/Linaro 4.7.3-12ubuntu1) 4.7.3
# remove the installed gems from the sample project,
# of course removing only the mysql2 gem works too
thinreports-rails3-example $ rm -rf vendor
# install the gems again
thinreports-rails3-example $ bundle install --path vendor/bundle
# run a DB task with rake
thinreports-rails3-example $ bundle exec rake db:drop
# the command shouldn't error out
```

The OS is Ubuntu 14.04 x64.

sodabrew:
Resolves #698, #695, #685.

/cc @tamird

### DIFF
--- a/ext/mysql2/extconf.rb
+++ b/ext/mysql2/extconf.rb
@@ -105,30 +105,51 @@ wishlist = [
   '-Wno-missing-field-initializers', # gperf generates bad code
   '-Wno-missing-variable-declarations', # missing symbols due to ruby native ext initialization
   '-Wno-padded', # mysql :(
+  '-Wno-reserved-id-macro', # rubby :(
   '-Wno-sign-conversion', # gperf generates bad code
   '-Wno-static-in-inline', # gperf generates bad code
   '-Wno-switch-enum', # result.c -- enum_field_types (when not fully covered, e.g. mysql 5.6+)
   '-Wno-undef', # rubinius :(
+  '-Wno-unreachable-code', # rubby :(
   '-Wno-used-but-marked-unused', # rubby :(
 ]
 
-if ENV['CI']
-  wishlist += [
-    '-Werror',
-    '-fsanitize=address',
-    '-fsanitize=cfi',
-    '-fsanitize=integer',
-    '-fsanitize=memory',
-    '-fsanitize=thread',
-    '-fsanitize=undefined',
-  ]
-end
-
 usable_flags = wishlist.select do |flag|
-  try_link('int main() {return 0;}', flag)
+  try_link('int main() {return 0;}',  "-Werror -Wunknown-warning-option #{flag}")
 end
 
 $CFLAGS << ' ' << usable_flags.join(' ')
+
+enabled_sanitizers = disabled_sanitizers = []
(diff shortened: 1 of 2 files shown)


ext/mysql2/extconf.rb
```diff
@@ -105,31 +105,52 @@ def asplode(lib)
...
 $CFLAGS << ' ' << usable_flags.join(' ')
 
+enabled_sanitizers = disabled_sanitizers = []
+# Specify a commna-separated list of sanitizers, or try them all by default
+sanitizers = with_config('sanitize')
+case sanitizers
+when true
+  # Try them all, turn on whatever we can
+  sanitizers = 'address,cfi,integer,memory,thread,undefined'
+  enabled_sanitizers, _ = sanitizers.split(',').partition do |s|
```
tamird: s/partition/select/

tamird:
LGTM. Do you want these sanitizers enabled in CI?

ext/mysql2/extconf.rb
sodabrew: Yep since I'm ignoring the false array anyhow.

sodabrew:
At the moment, only the Travis OS X instances are compiling with new enough clang or GCC, and for some reason I haven't figured out yet, mysql2 compiled with address sanitizer hangs on OS X. So for now I'll leave these off in CI. When Travis moves to Ubuntu Trusty, its gcc does include address sanitizer, and I'll revisit.

PR #706, opened 2015-12-01
Allow nil for timeouts instead of casting to zero

carsonreinke:
The only problem I can see is that the default `connect_timeout` would not be used if set to nil in the `opts`.

### DIFF
--- a/lib/mysql2/client.rb
+++ b/lib/mysql2/client.rb
@@ -36,7 +36,7 @@ module Mysql2
         when :reconnect, :local_infile, :secure_auth
           send(:"#{key}=", !!opts[key]) # rubocop:disable Style/DoubleNegation
         when :connect_timeout, :read_timeout, :write_timeout
-          send(:"#{key}=", opts[key].to_i)
+          send(:"#{key}=", opts[key]) unless opts[key].nil?
         else
           send(:"#{key}=", opts[key])
         end
--- a/spec/mysql2/client_spec.rb
+++ b/spec/mysql2/client_spec.rb
@@ -352,6 +352,12 @@ RSpec.describe Mysql2::Client do
     }.to raise_error(Mysql2::Error)
   end
 
+  it "should allow nil read_timeout" do
+    client = Mysql2::Client.new(:read_timeout => nil)
+
+    expect(client.read_timeout).to be_nil
+  end
+
   context "#query" do
     it "should let you query again if iterating is finished when streaming" do
       @client.query("SELECT 1 UNION SELECT 2", :stream => true, :cache_rows => false).each.to_a


lib/mysql2/client.rb
```diff
@@ -36,7 +36,7 @@ def initialize(opts = {})
         when :reconnect, :local_infile, :secure_auth
           send(:"#{key}=", !!opts[key]) # rubocop:disable Style/DoubleNegation
         when :connect_timeout, :read_timeout, :write_timeout
-          send(:"#{key}=", opts[key].to_i)
+          send(:"#{key}=", opts[key].nil?() ? nil : opts[key].to_i)
```
sodabrew: The trailing parens on `nil?` are not needed.

sodabrew:
Yes, I think you're right. Nice catch!

lib/mysql2/client.rb
carsonreinke: `.nil? ? nil :` looks weird to me, I will change it though.

carsonreinke:
@sodabrew this should be good to go, let me know otherwise.

lib/mysql2/client.rb
```diff
@@ -36,7 +36,7 @@ def initialize(opts = {})
         when :reconnect, :local_infile, :secure_auth
           send(:"#{key}=", !!opts[key]) # rubocop:disable Style/DoubleNegation
         when :connect_timeout, :read_timeout, :write_timeout
-          send(:"#{key}=", opts[key].to_i)
+          send(:"#{key}=", opts[key].nil? ? nil : opts[key].to_i)
```
jeremy: Perhaps we should drop the `to_i` entirely.

jeremy: Aha – it's due to accepting String values from URI params without explicit conversion. #313

sodabrew:
Unit tests are failing at:
```
  1) Mysql2::Client should allow nil read_timeout
     Failure/Error: send(:"#{key}=", opts[key].nil? ? nil : opts[key].to_i)
     
     TypeError:
       wrong argument type nil (expected Fixnum)
     # ./lib/mysql2/client.rb:39:in `read_timeout='
     # ./lib/mysql2/client.rb:39:in `block in initialize'
     # ./lib/mysql2/client.rb:33:in `each'
     # ./lib/mysql2/client.rb:33:in `initialize'
     # ./spec/mysql2/client_spec.rb:356:in `new'
     # ./spec/mysql2/client_spec.rb:356:in `block (2 levels) in <top (required)>'
```

Which is in code here:
``` C
static VALUE set_read_timeout(VALUE self, VALUE value) {
  long int sec;
  Check_Type(value, T_FIXNUM);
  sec = FIX2INT(value);
  if (sec < 0) {
    rb_raise(cMysql2Error, "read_timeout must be a positive integer, you passed %ld", sec);
```

Since the default timeout value is nil, it means that the underlying code should also allow assigning nil. But I don't see that MySQL can actually accept a non-positive-Integer value for this setting: https://dev.mysql.com/doc/refman/5.5/en/mysql-options.html

This suggests to me that, for this option, nil means "just leave this at the default" and so we shouldn't even pass the value through from the Client#initialize if the value is nil.

carsonreinke:
Not sure how I missed my own test failing.  Maybe just a simple `unless nil?` will avoid the cast?  Just instead of having the timeout as 0.

sodabrew:
I think this would be OK. I do feel a little weird that once you set a non-nil value, you can't set it back to nil, but it's definitely better than the error case you're seeing.
``` ruby
  send(:"#{key}=", opts[key].to_i) unless opts[key].nil?
```

carsonreinke:
Fixed plus rebase.  I can't see anything from Travis though, it oddly just says "There was an error while loading data"

sodabrew:
Travis had a service failure earlier today, and is still stuck/behind on
the queue. Seems like it will be a few more hours.

sodabrew:
Travis running again and tests are passing. RuboCop picked up this lint warning:
```
spec/mysql2/client_spec.rb:354:1: C: Style/TrailingWhitespace: Trailing whitespace detected.
spec/mysql2/client_spec.rb:357:1: C: Style/TrailingWhitespace: Trailing whitespace detected.
```

carsonreinke:
How can I get Travis fixed on this?  I still see it as pending.

PR #722, opened 2016-01-14
Fix: BigDecimal statement parameter is ignored.

mperrando:
Added code to treat a parameter of Ruby type BigDecimal.


ext/mysql2/statement.c
```diff
@@ -204,6 +204,15 @@ static void *nogvl_stmt_store_result(void *ptr) {
     xfree(length_buffers);                                  \
   }
 
+/* Prepare a string parameter
+ */
+#define SET_BUFFER_FOR_STRING                                  \
```
sodabrew: I think this should be a function rather than a macro.

mperrando:
@sodabrew I strongly agree.
I followed the FREE_BINDS idea, but I prefer a function too.
I will change the PR ASAP.

mperrando:
I have updated the PR.

Issue #738, opened 2016-03-01
failed to allocate memory for an empty result set


dreyks:
I'm not sure if this is related to #736 

This only happens when query returns empty result set

```console
2.3.0:001> require 'mysql2'
 => true
2.3.0:002> client = Mysql2::Client.new(username: :root, password: :password)
 => #<Mysql2::Client:0x000000028c3dc0 @read_timeout=nil, @query_options={:as=>:hash, :async=>false, :cast_booleans=>false, :symbolize_keys=>false, :database_timezone=>:local, :application_timezone=>nil, :cache_rows=>true, :connect_flags=>2147525125, :cast=>true, :default_file=>nil, :default_group=>nil, :username=>:root, :password=>:password}>
2.3.0:003> result = client.query('SELECT NULL LIMIT 0')
 => #<Mysql2::Result:0x000000028b3c68 @query_options={:as=>:hash, :async=>false, :cast_booleans=>false, :symbolize_keys=>false, :database_timezone=>:local, :application_timezone=>nil, :cache_rows=>true, :connect_flags=>2147525125, :cast=>true, :default_file=>nil, :default_group=>nil, :username=>:root, :password=>:password}>
2.3.0:004> result.to_a
 => []
2.3.0:005> result.to_a
NoMemoryError: failed to allocate memory

```

I think 5189cfee45872a10cf54afbfcd721de12b2ae36e is responsible for this

sodabrew:
Resolved by #741

PR #741, opened 2016-03-06
Fix segv when num of rows is zero

### LINKED ISSUE #736: Results buffer allocation fails with prepared statements
In `rb_mysql_result_alloc_result_buffers` of result.c, the line 
[`wrapper->result_buffers[i].buffer = xmalloc(fields[i].max_length);`](https://github.com/brianmario/mysql2/blob/e8001bb0a78dd8a88504b2884131bd4cc3be68d5/ext/mysql2/result.c#L321)
is resulting in an error: `failed to allocate memory (NoMemoryError)`

While debugging this issue, I ran a `printf( "%lu\n", fields[i].max_length );` and got varying values:
139857292607408, 140190085463984, etc. (That corresponds to ~140 TB which is clearly incorrect. Possibly uninitialized memory?)

Per the MySQL docs: (ctrl+f max_length) http://dev.mysql.com/doc/refman/5.7/en/c-api-data-structures.html `STMT_ATTR_UPDATE_MAX_LENGTH` needs to be set for `max_length` to be set.
Oddly, it [does appear to be being set](https://github.com/brianmario/mysql2/blob/e8001bb0a78dd8a88504b2884131bd4cc3be68d5/ext/mysql2/statement.c#L123-L128) (and I verified that this code was being run).

I'm a little lost here and would appreciate help in debugging this issue. Is anyone able to reproduce it?

Update: To clarify, sometimes it succeeds with normal values for max_length (e.g. 34, 54, 60)

kamipo:
Fixes #736, #738.

Sorry for my wrong commit. It caused by 5189cfe when num of rows is zero.

Current code has 2 problems.

First one, `numberOfRows ==  0` and `lastRowProcessed == 0` are valid result when result rows is zero.
We should use the condition `wrapper->rows == Qnil` instead.

Second one, we should initialize `fields` before `rb_mysql_result_free_result` even if result rows is zero.
Otherwise the following code causes segv:

```ruby
require 'mysql2'

client = Mysql2::Client.new(
  host: 'localhost',
  username: 'root',
  database: 'mysql',
)

5000.times do |n|
  puts n
  result = client.query('SELECT * FROM user WHERE 1=0 LIMIT 0')
  result.to_a
  result.fields
end
```

### DIFF
--- a/ext/mysql2/result.c
+++ b/ext/mysql2/result.c
@@ -913,7 +913,7 @@ static VALUE rb_mysql_result_each(int argc, VALUE * argv, VALUE self) {
     app_timezone = Qnil;
   }
 
-  if (wrapper->lastRowProcessed == 0 && !wrapper->is_streaming) {
+  if (wrapper->rows == Qnil && !wrapper->is_streaming) {
     wrapper->numberOfRows = wrapper->stmt_wrapper ? mysql_stmt_num_rows(wrapper->stmt_wrapper->stmt) : mysql_num_rows(wrapper->result);
     if (wrapper->numberOfRows == 0) {
       rb_mysql_result_free_result(wrapper);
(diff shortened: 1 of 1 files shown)


sodabrew:
Thank you for fixing and further improving these corner cases!

PR #742, opened 2016-03-08
Fix 0.3.x segfault for :cache_rows => false

### LINKED ISSUE #691: segfault when using cache_rows true
Noticing new segfaults when using cache_rows true

example here:

https://gist.github.com/dandv/9d260ba2bf1113fd0b4e

This used to work perfect feels like a recent regression.

ajaska:
This fixes #691.

I managed to reproduce the segfault with the following code (note: the `SELECT` just returns the numbers 1..10000).

```ruby
require 'mysql2'                                                                   
                                                                                   
client = Mysql2::Client.new(:host => "localhost", :username => "vagrant", :database => "test")
                                                                                   
results = client.query("SELECT @row := @row + 1 as row FROM " \                    
"(select 0 union all select 1 union all select 3 union all select 4 union all select 5 union all select 6 union all select 6 union all select 7 union all select 8 union all select 9) t, "\
"(select 0 union all select 1 union all select 3 union all select 4 union all select 5 union all select 6 union all select 6 union all select 7 union all select 8 union all select 9) t2, "\
"(select 0 union all select 1 union all select 3 union all select 4 union all select 5 union all select 6 union all select 6 union all select 7 union all select 8 union all select 9) t3, "\
"(select 0 union all select 1 union all select 3 union all select 4 union all select 5 union all select 6 union all select 6 union all select 7 union all select 8 union all select 9) t4, "\
"(SELECT @row:=0) as f", :cache_rows => false)                                     
                                                                                   
puts results.to_a.length                                                           
puts results.to_a.length                                                           
puts results.to_a.length                                                           
puts results.to_a.length                                                           
puts results.to_a.length                                                           
puts results.to_a.length                                                           
puts results.to_a.length                                                           
puts results.to_a.length                                                           
puts results.to_a.length                                                           
puts results.to_a.length                                                           
puts results.to_a.length  
```

(Yes, the repeated `to_a` and large `SELECT` appear to be necessary, I'm not entirely sure why.)

The segfault is related to `wrapper->lastRowProcessed` not being set back to 0 when we free the mysql result. (Thus causing issues with future calls of `#each`.) I'm not sure what the direct cause is, (and it's very late at night and I can't reason about code very well at the moment). This seems to fix it, and I'll take another look tomorrow.

### DIFF
--- a/ext/mysql2/result.c
+++ b/ext/mysql2/result.c
@@ -531,6 +531,10 @@ static VALUE rb_mysql_result_each(int argc, VALUE * argv, VALUE self) {
         return wrapper->rows;
       }
       wrapper->rows = rb_ary_new2(wrapper->numberOfRows);
+    } else if (!cacheRows && wrapper->lastRowProcessed == wrapper->numberOfRows) {
+      mysql_data_seek(wrapper->result, 0);
+      wrapper->lastRowProcessed = 0;
+      wrapper->rows = rb_ary_new2(wrapper->numberOfRows);
     }
 
     if (cacheRows && wrapper->lastRowProcessed == wrapper->numberOfRows) {
@@ -558,7 +562,9 @@ static VALUE rb_mysql_result_each(int argc, VALUE * argv, VALUE self) {
 
         if (row == Qnil) {
           /* we don't need the mysql C dataset around anymore, peace it */
-          rb_mysql_result_free_result(wrapper);
+          if (cacheRows) {
+            rb_mysql_result_free_result(wrapper);
+          }
           return Qnil;
         }
 
@@ -566,7 +572,7 @@ static VALUE rb_mysql_result_each(int argc, VALUE * argv, VALUE self) {
           rb_yield(row);
         }
       }
-      if (wrapper->lastRowProcessed == wrapper->numberOfRows) {
+      if (wrapper->lastRowProcessed == wrapper->numberOfRows && cacheRows) {
         /* we don't need the mysql C dataset around anymore, peace it */
         rb_mysql_result_free_result(wrapper);
       }
--- a/spec/mysql2/result_spec.rb
+++ b/spec/mysql2/result_spec.rb
@@ -78,6 +78,11 @@ describe Mysql2::Result do
       result.first.object_id.should_not eql(result.first.object_id)
     end
 
+    it "should be able to iterate a second time even if cache_rows is disabled" do
+      result = @client.query "SELECT 1 UNION SELECT 2", :cache_rows => false
+      result.to_a.should eql(result.to_a)
+    end
+
     it "should yield different value for #first if streaming" do
       result = @client.query "SELECT 1 UNION SELECT 2", :stream => true, :cache_rows => false
       result.first.should_not eql(result.first)


kamipo:
We can't call `rb_mysql_result_fetch_row` after `rb_mysql_result_free_result`.
I think this means that we can't call `to_a` (`rb_mysql_result_each`) twice when `:cache_rows => false`.

I propose the following two methods if fix this problem:

* Avoid call `rb_mysql_result_free_result` implicitly
* Raise an error when `wrapper->resultFreed` is true

ext/mysql2/result.c
```diff
@@ -78,6 +78,7 @@ static void rb_mysql_result_free_result(mysql2_result_wrapper * wrapper) {
   if (wrapper && wrapper->resultFreed != 1) {
     /* FIXME: this may call flush_use_result, which can hit the socket */
     mysql_free_result(wrapper->result);
+    wrapper->lastRowProcessed = 0;
```
kamipo: Looks like that it will break the `:cache_rows => true` behavior.
https://github.com/ajaska/mysql2/blob/0.3.x/ext/mysql2/result.c#L528-L535

sodabrew:
I think you're right - calling .to_a will iterate the entire result set. When cacheRows is false, we do not allow iterating the result set again.

ext/mysql2/result.c
ajaska: Yes, you are correct.

ajaska:
Hm, wait that might not be right approach. Should an extra variable be declared on the `wrapper` struct?

ajaska:
e.g.

```ruby
results.each do |row|                                                           
  results.each do |row2|                                                        
    puts "#{row['row']} #{row2['row']}"                                         
  end                                                                           
end
```

Produces
```
1 2
1 3
...
1 10000
test.rb:18:in `each': You have already fetched all the rows for this query and cache_rows is false. (to reiterate you must requery). (Mysql2::Error)
```

But we probably want it to error immediately.

ajaska:
Oh dear. I was trying to compare it to the 0.4.x behavior... and got another segfault.

The following code segfaults for me somewhere around ~60% of the time.
```ruby
require 'mysql2'

puts Gem.loaded_specs["mysql2"].version # 0.4.3, built from origin/HEAD

client = Mysql2::Client.new(:host => "localhost", :username => "vagrant", :database => "test")

results = client.query("SELECT @row := @row + 1 as row FROM " \
"(select 0 union all select 1 union all select 3 union all select 4 union all select 5 union all select 6 union all select 6 union all select 7 union all select 8 union all select 9) t, "\
"(select 0 union all select 1 union all select 3 union all select 4 union all select 5 union all select 6 union all select 6 union all select 7 union all select 8 union all select 9) t2, "\
"(select 0 union all select 1 union all select 3 union all select 4 union all select 5 union all select 6 union all select 6 union all select 7 union all select 8 union all select 9) t3, "\
"(select 0 union all select 1 union all select 3 union all select 4 union all select 5 union all select 6 union all select 6 union all select 7 union all select 8 union all select 9) t4, "\
"(select 0 union all select 1 union all select 3 union all select 4 union all select 5 union all select 6 union all select 6 union all select 7 union all select 8 union all select 9) t5, "\
"(SELECT @row:=0) as f", :cache_rows => false)

results.each do |row|
  results.each do |row2|
  end
end

puts results.to_a.length
puts results.to_a.length
puts results.to_a.length
puts results.to_a.length
puts results.to_a.length
puts results.to_a.length
puts results.to_a.length
puts results.to_a.length
puts results.to_a.length
puts results.to_a.length
puts results.to_a.length
puts results.to_a.length
puts results.to_a.length
```

ajaska:
via `README.md` it looks the like appropriate response is to throw an error if a results with `:cache_rows` false is iterated more than once

> If you only plan on using each row once, then it's much more efficient to disable this behavior by setting the :cache_rows option to false. This would be helpful if you wanted to iterate over the results in a streaming manner. Meaning the GC would cleanup rows you don't need anymore as you're iterating over the result set.

ajaska:
Hm. This, however, is at odds with one of the specs (0.4.3):

```ruby
    it "should not cache previously yielded results if cache_rows is disabled" do
      result = @client.query "SELECT 1", :cache_rows => false
      expect(result.first.object_id).not_to eql(result.first.object_id)
    end
```

Suggesting a result can be reused. 

While testing to see what that test did exactly, I wrote:

```ruby
    it "should not cache previously yielded results if cache_rows is disabled" do
      result = @client.query "SELECT 1", :cache_rows => false                      
      puts result.first                                                            
      puts result.first                                                            
      expect(result.first.object_id).not_to eql(result.first.object_id)            
    end   
```

This resulted in a segfault:

```
  #each
    should be able to return results as an array
    should yield rows as hash's with symbol keys if :symbolize_keys was set to true
    should cache previously yielded results by default
    should throw an exception if we try to iterate twice when streaming is enabled
    should yield the same value for #first if streaming is disabled
    should throw an exception if we try to iterate twice when cache_rows is disabled (FAILED - 38)
{"1"=>1}

/home/vagrant/repos/mysql2/spec/mysql2/result_spec.rb:84: [BUG] Segmentation fault
ruby 1.9.3p327 (2012-11-10) [x86_64-linux] Brightbox

-- Control frame information -----------------------------------------------
c:0037 p:---- s:0135 b:0135 l:000134 d:000134 CFUNC  :each
c:0036 p:---- s:0133 b:0133 l:0000c0 d:0000c0 CFUNC  :first
c:0035 p:0072 s:0130 b:0129 l:000ae8 d:000128 BLOCK  /home/vagrant/repos/mysql2/spec/mysql2/result_spec.rb:84
[...]
```

Therefore, I'm pretty sure that there's something wrong with that spec.

sodabrew:
OK now I follow. The code paths are complicated because of streaming and cache_rows. The idea behind cache_rows is that mysql2 isn't going to slurp the result set into Ruby space, but will instead keep the result set buffers in libmysqlclient and reuse those every time `.each` is called. We may simply be freeing the result set too soon.

|   |      **streaming true**      |  **streaming false (default)** |
|----------|:-------------:|------:|
| **cache_rows true (default)** |  ignored  | valid  |
| **cache_rows false** |    required   |   valid |

ajaska:
Ok, I think I understand...
If `:cache_rows => false` it is still allowed to `.each` again, but we return the results from libmysqlclient buffers (i.e. don't hit the network again, but have to deal with the computational/memory requirements of generating the ruby rows)

ajaska:
[This is one way to accomplish that](https://github.com/brianmario/mysql2/compare/master...ajaska:allow-repeated-each) (0.4.x tree)

sodabrew:
Yes, that looks right. I think that block may be incorrect. I will take another look at it soon to double check.

ajaska:
Actually, how's this?

```diff
diff --git a/ext/mysql2/result.c b/ext/mysql2/result.c
index a24d28c..b51c46a 100644
--- a/ext/mysql2/result.c
+++ b/ext/mysql2/result.c
@@ -834,7 +834,9 @@ static VALUE rb_mysql_result_each_(VALUE self,
 
         if (row == Qnil) {
           /* we don't need the mysql C dataset around anymore, peace it */
-          rb_mysql_result_free_result(wrapper);
+          if (args->cacheRows) {
+            rb_mysql_result_free_result(wrapper);
+          }
           return Qnil;
         }
 
@@ -842,7 +844,7 @@ static VALUE rb_mysql_result_each_(VALUE self,
           rb_yield(row);
         }
       }
-      if (wrapper->lastRowProcessed == wrapper->numberOfRows) {
+      if (wrapper->lastRowProcessed == wrapper->numberOfRows && args->cacheRows) {
         /* we don't need the mysql C dataset around anymore, peace it */
         rb_mysql_result_free_result(wrapper);
       }
@@ -917,6 +919,10 @@ static VALUE rb_mysql_result_each(int argc, VALUE * argv, VALUE self) {
   if (wrapper->rows == Qnil && !wrapper->is_streaming) {
     wrapper->numberOfRows = wrapper->stmt_wrapper ? mysql_stmt_num_rows(wrapper->stmt_wrapper->stmt) : mysql_num_rows(wrapper->result);
     wrapper->rows = rb_ary_new2(wrapper->numberOfRows);
+  } else if (wrapper->rows && !cacheRows) {
+    mysql_data_seek(wrapper->result, 0);
+    wrapper->lastRowProcessed = 0;
+    wrapper->rows = rb_ary_new2(wrapper->numberOfRows);
   }
 
   // Backward compat
diff --git a/spec/mysql2/result_spec.rb b/spec/mysql2/result_spec.rb
index 9633588..ce797ba 100644
--- a/spec/mysql2/result_spec.rb
+++ b/spec/mysql2/result_spec.rb
@@ -82,6 +82,11 @@ RSpec.describe Mysql2::Result do
       expect(result.first.object_id).not_to eql(result.first.object_id)
     end
 
+    it "should be able to iterate a second time even if cache_rows is disabled" do
+      result = @client.query "SELECT 1 UNION SELECT 2", :cache_rows => false
+      expect(result.to_a).to eql(result.to_a)
+    end
+
     it "should yield different value for #first if streaming" do
       result = @client.query "SELECT 1 UNION SELECT 2", :stream => true, :cache_rows => false
       expect(result.first).not_to eql(result.first)
```

sodabrew:
That looks pretty good, too.

sodabrew:
Oh, this PR is targeted to 0.3.x -- I think the same issue will need to be addressed in master, so I'd prefer to have the PR target to master and then I'll backport it to the 0.3.x branch.

ajaska:
Sorry, github automatically closed the request, here's the reopened it. This is the backported version and I'll make the master PR in 30 secs

PR #744, opened 2016-03-09
Disallow iterating results multiple times with false :cache_rows

ajaska:
See notes from #742 . Pretty sure this should be backported to 0.3.x as well.

This should fix some cases of segfaulting.

Note: I opted to add the lines in result.c where I did to minimize the numbers of lines with a diff :)


ajaska:
Should be fixed by #745, closing.

PR #745, opened 2016-03-11
Allow repeated each with :cache_rows => false

ajaska:
Fixes segfaults... see discussion in

#691 (initial report for 0.3.x)
#742 (diagnosis, discovery that this affects 0.4.x, and discussion)
#744 (fix that worked the wrong way)


sodabrew:
So much awesome!

PR #731, opened 2016-02-15
Apply the use `Thread.handle_interrupt` to protect `query` fix to the 0.3.x branch

rrosenblum:
This is to back port of @tamird's #592 to the `0.3.x` branch.

I recently came across the issue, `This connection is still waiting for a result, try again once you have the result`. I am running Ruby 2.2.4, Rails 3.2.22, Mysql2 0.3.20, and MySQL 5.6. I came across issues #99 and #566. These led me to PR #592. I saw that the fix was applied to `master` (`0.4.x`), but not to `0.3.x`. This change was able to solve the problem that I was experiencing. I was wondering why the fix was not applied to the `0.3.x` branch?

PR #751, opened 2016-05-07
Backport 592

sodabrew:
Second pass at #731 with rebased commits from #592 and #679.


rrosenblum:
Thank you for taking care of this and cutting a new version of the gem. This is going to be a big help to me and many others.

PR #752, opened 2016-05-07
Update RuboCop to 0.50.x

tamird:


sodabrew:
I've awoken a slumbering giant.

tamird:
😘

tamird:
I've cherry picked the rspec matcher change from #747 since the alternative (replacing parallel assignment with serial assignment) was causing failures on 1.8.7 and segfaults on REE o.O

tamird:
Looks like there are some (unrelated?) failures; segfault on REE and memory corruption weirdness in some 2.0.0 tests.

tamird:
@sodabrew can i bribe you to take a look at this?

examples/threaded.rb
```diff
@@ -6,7 +6,7 @@
 
 # Should never exceed worst case 3.5 secs across all 20 threads
 Timeout.timeout(3.5) do
-  20.times.map do
+  Array.new(20) do
```
sodabrew: There's a style against Integer#times now?

tamird: yep, this is apparently more performant

jaredbeck:
Oh, I didn't realize you were working on this, Tamir.

I opened https://github.com/brianmario/mysql2/pull/909, which upgrades rubocop to 0.41.2. But, it looks like you've done a lot of work here, and I don't want to cause a conflict, so maybe we should close mine.

tamird:
Yeah, this one goes a bit further than yours. Let's see what @sodabrew says.

Rakefile
```diff
@@ -8,17 +9,11 @@ load 'tasks/compile.rake'
 load 'tasks/generate.rake'
 load 'tasks/benchmarks.rake'
 
-# TODO: remove engine check when rubinius stops crashing on RuboCop
```
sodabrew: Oh nice, it stopped crashing on recent Rubinius?

tamird: I haven't actually tested that yet (I'm about to), but Rubinius support looks have been removed some time ago (1df20eb4aa8666ceea664c01e76eac402332f039).

tamird: Rubinius is in a bit of a rough state it seems:
https://github.com/bundler/bundler/issues/6163
https://github.com/rubinius/rubinius/issues/3769

I'll check again when either of those issues is fixed.

sodabrew:
This looks good. Also, I'm thinking about when to drop Ruby 1.9, too.

Rakefile
sodabrew: Yes, true - Rubinius was in rough shape, and even as an allowed failure in Travis, was just gumming up the works to try to keep up.

tamird:
Great, I'd be happy to prepare a PR that does that (with no urgency to merge) after this goes in.

jaredbeck:
```ruby
gem 'rubocop', '~> 0.50.0' unless RUBY_VERSION =~ /1.9/
```

This is an interesting way to get us to a modern rubocop version! Works for me. I can't think of any downside to skip linting on 1.9.3. There aren't any 1.9.3-specific linter offenses ("cops"), are there?

tamird:
@jaredbeck the linter isn't being run on 1.9.3, but it is still configured to target 1.9 - see `.rubocop.yml`.

0.51 removed the ability to **target** 1.9, which is why this is the terminal version until 1.9 support is dropped from mysql2.

tamird:
@sodabrew thanks for the review. Is this waiting on anything?

sodabrew:
Right - that's what I mean - for the next mysql2 rev, I'm considering whether to drop 1.9 as well, and make Ruby 2.0 the lowest supported version. Given that each version branch has multi year lifespan, that 0.4.x supports all the way back to 1.8.7 gives a lot of breathing room to make 0.5.x more aggressively "modern".

tamird:
Understood. Still, it would be beneficial to merge this as-is and pursue (and evaluate the benefits of) dropping 1.9 in a separate PR -- if nothing else, it will make review at least somewhat palatable.

jaredbeck:
> @sodabrew sodabrew merged commit 17fe755 into brianmario:master 31 minutes ago

:tada: Nice work, Tamir!

PR #755, opened 2016-05-20
Make server_status available

cdegroot:
This patch sets server_status in MySql2::Client whenever it's relevant. These flags contain important information about how the server viewed the query; it's useful to log this (or even throw errors ;-)) during development so slow queries can be prevented from reaching production. 

First patch, first time I'm doing Ruby extensions, so flame away. If the code is good in principle, I'll look into adding some test coverage.

### DIFF
--- a/ext/mysql2/client.c
+++ b/ext/mysql2/client.c
@@ -525,6 +525,8 @@ static VALUE rb_mysql_client_async_result(VALUE self) {
   Check_Type(current, T_HASH);
   resultObj = rb_mysql_result_to_obj(self, wrapper->encoding, current, result, Qnil);
 
+  rb_iv_set(self, "@server_status", INT2NUM(wrapper->client->server_status));
+
   return resultObj;
 }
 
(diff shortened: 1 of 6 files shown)


sodabrew:
Thanks, I need to do a bit of reading since the MYSQL structure is semi-opaque per http://dev.mysql.com/doc/refman/5.7/en/c-api-data-structures.html -- I think we reach into it in just one other place.

Would you also take a look at the spec/ directory to add a test?

cdegroot:
If you look positively at this patch, I'm absolutely going to be creative and add a test, no worries :-). We have a long weekend up here in Canada, will whip something up on Tuesday.

sodabrew:
Thinking some more about this, it might make sense to break these out as separate true/false properties of the Result object rather than the Client object.

cdegroot:
Either works for me - your call. The reason I took this approach is mainly that it stays "in the same object" as the MySQL C client, and it just sets a single value minimizing overhead - I have no idea how many instructions are involved in setting Ruby properties from C, so I wanted to stay on the safe side.

cdegroot:
Update: moved to three individual flags in a `@server_flags` hash and added a quick test.

ext/mysql2/client.h
```diff
@@ -65,9 +65,11 @@ void rb_mysql_client_set_active_thread(VALUE self);
 
 void init_mysql2_client(void);
 void decr_mysql2_client(mysql_client_wrapper *wrapper);
+void set_server_query_flags(MYSQL *client, VALUE result);
```
sodabrew: This function is only called from within client.c, so you can make it static and leave it out of the header file.

sodabrew: Oh, it's also called from statement.c, sorry.

sodabrew:
I like the second version a lot! Just a nit about the function name.

I have half an idea to add accessors like these:
```
Mysql2::Result#no_index_used?
Mysql2::Result#no_good_index_used?
Mysql2::Result#query_was_slow?
```

The test failures are unrelated, I will look into those as I have time. Your new test passes.

cdegroot:
(Updated function name and location in header file)

ext/mysql2/client.c
```diff
@@ -1465,3 +1467,30 @@ void init_mysql2_client() {
...
 #endif
 }
+
+#define flag_to_bool(f) ((client->server_status & f) ? Qtrue : Qfalse)
+
+void rb_mysql_set_server_query_flags(MYSQL *client, VALUE result) {
+  VALUE server_flags = rb_hash_new();
+
+#ifdef SERVER_QUERY_NO_GOOD_INDEX_USED
+  rb_hash_aset(server_flags, ID2SYM(rb_intern("no_good_index_used")), flag_to_bool(SERVER_QUERY_NO_GOOD_INDEX_USED));
```
sodabrew: Could you add these symbols as `static VALUE`s at the top of the file and in the module init? You should see a laundry list sym_foo values initialized around line 1350.

sodabrew:
Merged as 811a57a

PR #759, opened 2016-06-11
Fix SEGV when `wrapper->result` is used after it is freed.

kubo:
I got SEGV when I ran rspec.
I found that the `wrapper->result` was used after it was freed.
I'm not sure whether this fix is correct but it works fine for me.

### DIFF
--- a/ext/mysql2/result.c
+++ b/ext/mysql2/result.c
@@ -920,6 +920,9 @@ static VALUE rb_mysql_result_each(int argc, VALUE * argv, VALUE self) {
     wrapper->numberOfRows = wrapper->stmt_wrapper ? mysql_stmt_num_rows(wrapper->stmt_wrapper->stmt) : mysql_num_rows(wrapper->result);
     wrapper->rows = rb_ary_new2(wrapper->numberOfRows);
   } else if (wrapper->rows && !cacheRows) {
+    if (wrapper->resultFreed) {
+      rb_raise(cMysql2Error, "Result set has already been freed");
+    }
     mysql_data_seek(wrapper->result, 0);
     wrapper->lastRowProcessed = 0;
     wrapper->rows = rb_ary_new2(wrapper->numberOfRows);


ext/mysql2/result.c
```diff
@@ -920,6 +920,9 @@ static VALUE rb_mysql_result_each(int argc, VALUE * argv, VALUE self) {
     wrapper->numberOfRows = wrapper->stmt_wrapper ? mysql_stmt_num_rows(wrapper->stmt_wrapper->stmt) : mysql_num_rows(wrapper->result);
     wrapper->rows = rb_ary_new2(wrapper->numberOfRows);
   } else if (wrapper->rows && !cacheRows) {
+    if (wrapper->resultFreed) {
+      rb_raise(cMysql2Error, "Statement result already freed");
+    }
```
sodabrew: Nice find. I would raise this higher up in the function - just after the check for `"Statement handle already closed"` above.

That said, I wonder what would cause this. Are you using the word "Statement" here specifically because this only occurs for prepared statements, or would the word "Query" be more appropriate?

sodabrew:
I'd like to figure out a reproduction case for this.

kubo:
I reproduced this issue just by running `rake spec`.
My environment is:
```shell
$ cat /etc/os-release 
NAME="Ubuntu"
VERSION="12.04.5 LTS, Precise Pangolin"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu precise (12.04.5 LTS)"
VERSION_ID="12.04"
$ mysqld --version
mysqld  Ver 5.7.12 for Linux on x86_64 (Source distribution)
$ ruby --version
ruby 2.3.0p0 (2015-12-25 revision 53290) [x86_64-linux]
```

I changed the code as follows,
```
     if (wrapper->resultFreed) {
      fprintf(stderr, "#######################\n");
      rb_raise(cMysql2Error, "Statement result already freed");
    }
```
run `rake spec` and got the following result:
```
Mysql2::Statement
  ....
  #each
    ....
    should yield the same value for #first if streaming is disabled
    should yield rows as hash's with symbol keys if :symbolize_keys was set to true
#######################
    should throw an exception if we try to iterate twice when streaming is enabled
```
I tried several times and all `########` lines were printed just before `should throw an exception if we try to iterate twice when streaming is enabled`.
It looks like that the SEGV occurred there.
IMO, this depends on GC timing and `GC.stress = true` may increase probability which causes SEGV out of my environment.

sodabrew:
Could you rebase this to current master, as another CI test fix has landed, and change `Statement result already freed` to `Result set has already been freed` because this should apply to both ad-hoc Query and Prepared Statement results.

kubo:
I rebased the branch and changed the error message.

pacoguzman:
Hi @sodabrew we're facing that exception while iterating result sets with a lot of rows randomly, but with more changes the bigger the result set is.

As far I can understand the garbage collector is freeing the result set at some point.

Which is weird and correct me because probably I'm wrong, is that the we're seeing the error while we're iterating the result, and these changes seems to raise when building the `enumerable`.

We're using this through `Sequel` so probably the error could be there too, but just in case you can give us some light on this.

Thanks in advance.

sodabrew:
@pacoguzman Could you open a new issue with a more complete description of the problem? This PR was merged a year and a half ago, so I don't want to start a new thread here.

pacoguzman:
I'll follow on https://github.com/brianmario/mysql2/issues/938 as it seems the underlying problem is a `Lost connection to MySQL server during query`

PR #760, opened 2016-06-11
Support connect attributes and set the script name to `program_name` attribute if it is not specified.

kubo:
This is same feature with Perl DBD-MySQL `mysql_conn_attrs` option.
http://search.cpan.org/~capttofu/DBD-mysql-4.033/lib/DBD/mysql.pm#mysql_conn_attrs

### DIFF
--- a/ext/mysql2/client.c
+++ b/ext/mysql2/client.c
@@ -352,7 +352,22 @@ static VALUE rb_mysql_get_ssl_cipher(VALUE self)
   return rb_str;
 }
 
-static VALUE rb_connect(VALUE self, VALUE user, VALUE pass, VALUE host, VALUE port, VALUE database, VALUE socket, VALUE flags) {
+#if HAVE_MYSQL_OPTIONS4
+static int opt_connect_attr_add_i(VALUE key, VALUE value, VALUE arg)
+{
+  mysql_client_wrapper *wrapper = (mysql_client_wrapper *)arg;
+#ifdef HAVE_RUBY_ENCODING_H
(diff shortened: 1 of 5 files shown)


ext/mysql2/client.c
```diff
@@ -352,7 +352,36 @@ static VALUE rb_mysql_get_ssl_cipher(VALUE self)
   return rb_str;
 }
 
-static VALUE rb_connect(VALUE self, VALUE user, VALUE pass, VALUE host, VALUE port, VALUE database, VALUE socket, VALUE flags) {
+#if MYSQL_VERSION_ID >= 50606
```
sodabrew: I would prefer a have-function test on `mysql_options4` rather than a version check.

lib/mysql2/client.rb
```diff
@@ -83,8 +88,11 @@ def initialize(opts = {})
       port = port.to_i unless port.nil?
       database = database.to_s unless database.nil?
       socket = socket.to_s unless socket.nil?
+      conn_attrs = conn_attrs.each_with_object({}) do |(key, value), hash|
+        hash[key.to_sym] = value.to_s
```
sodabrew: If this were `hash[key.to_s] = value.to_s` and avoided symbols, it looks like a lot of code in `opt_connect_addr_add_i` could be removed, and just StringValueToCStr used for the key?

sodabrew:
This looks awesome, thank you for the contribution! Review comments added.

kubo:
I changed the code as you commented.

sodabrew:
Merged!

Issue #58, opened 2010-09-11
AR aggregate calculations always return BigDecimal


yolk:
When calling sum or average on a AR-Model it always returns a BigDecimal, even if the used attribute is of type INT:

```
TimeEntry.sum(:minutes)
=> #<BigDecimal:105dade18,'0.10091662E8',9(18)>

TimeEntry.first.minutes
=> 123
```

With the mysql gem sum returns an integer:

```
TimeEntry.sum(:minutes)
=> 10091662
```

This also occurs when using average, but not maximum or minimum. Is this behaviour intended?

I testet with Rails 3.0.0 on ruby-1.9.2.p0

PR #764, opened 2016-06-28
Avoid RangeError on integers larger than LONG_LONG

nurse:
CRuby's rb_big2ull raises RangeError if the argument is out of [LONG_MIN, ULONG_MAX].
To avoid that calling to_s and set as decimal with some tricks for edge cases.

### DIFF
--- a/ext/mysql2/statement.c
+++ b/ext/mysql2/statement.c
@@ -4,6 +4,9 @@ VALUE cMysql2Statement;
 extern VALUE mMysql2, cMysql2Error, cBigDecimal, cDateTime, cDate;
 static VALUE sym_stream, intern_new_with_args, intern_each;
 static VALUE intern_usec, intern_sec, intern_min, intern_hour, intern_day, intern_month, intern_year, intern_to_s;
+#ifndef HAVE_RB_BIG_CMP
+static ID id_cmp;
+#endif
 
 #define GET_STATEMENT(self) \
   mysql_stmt_wrapper *stmt_wrapper; \
@@ -203,6 +206,50 @@ static void set_buffer_for_string(MYSQL_BIND* bind_buffer, unsigned long *length
     xfree(length_buffers);                                  \
   }
 
+/* return 0 if the given bignum can cast as LONG_LONG, otherwise 1 */
+static int my_big2ll(VALUE bignum, LONG_LONG *ptr)
+{
+  unsigned LONG_LONG num;
+  size_t len;
+#ifdef HAVE_RB_ABSINT_SIZE
+  int nlz_bits = 0;
+  len = rb_absint_size(bignum, &nlz_bits);
+#else
+  len = RBIGNUM_LEN(bignum) * SIZEOF_BDIGITS;
+#endif
+  if (len > sizeof(LONG_LONG)) goto overflow;
+  if (RBIGNUM_POSITIVE_P(bignum)) {
+    num = rb_big2ull(bignum);
+    if (num > LLONG_MAX)
+      goto overflow;
+    *ptr = num;
+  }
+  else {
+    if (len == 8 &&
+#ifdef HAVE_RB_ABSINT_SIZE
+        nlz_bits == 0 &&
+#endif
+#if defined(HAVE_RB_ABSINT_SIZE) && defined(HAVE_RB_ABSINT_SINGLEBIT_P)
+        /* Optimized to avoid object allocation for Ruby 2.1+
+         * only -0x8000000000000000 is safe if `len == 8 && nlz_bits == 0`
+         */
+        !rb_absint_singlebit_p(bignum)
+#elif defined(HAVE_RB_BIG_CMP)
+        rb_big_cmp(bignum, LL2NUM(LLONG_MIN)) == INT2FIX(-1)
+#else
+        /* Ruby 1.8.7 and REE doesn't have rb_big_cmp */
+        rb_funcall(bignum, id_cmp, 1, LL2NUM(LLONG_MIN)) == INT2FIX(-1)
+#endif
+       ) {
+      goto overflow;
+    }
+    *ptr = rb_big2ll(bignum);
+  }
+  return 0;
+overflow:
+  return 1;
+}
+
 /* call-seq: stmt.execute
  *
  * Executes the current prepared statement, returns +result+.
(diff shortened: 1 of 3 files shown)


spec/mysql2/statement_spec.rb
```diff
@@ -59,6 +59,24 @@
     expect(rows).to eq([{ "1" => 1 }])
   end
 
+  it "should handle bignum but in int64_t" do
+    stmt = @client.prepare('SELECT ? AS max, ? AS min')
+    int64_max = (1 << 63) - 1
+    int64_min = -(1 << 63)
+    result = stmt.execute(int64_max, int64_min)
+    expect(result.to_a).to eq(['max' => int64_max, 'min' => int64_min])
+    @client.query 'DROP TABLE IF EXISTS mysql2_stmt_q'
```
sodabrew: No need to drop the table since these tests aren't creating/querying from it.

ext/mysql2/statement.c
```diff
@@ -266,8 +291,22 @@ static VALUE execute(int argc, VALUE *argv, VALUE self) {
           break;
         case T_BIGNUM:
           bind_buffers[i].buffer_type = MYSQL_TYPE_LONGLONG;
-          bind_buffers[i].buffer = xmalloc(sizeof(long long int));
-          *(LONG_LONG*)(bind_buffers[i].buffer) = rb_big2ll(argv[i]);
+          {
+              LONG_LONG num;
+              if (my_big2ll(argv[i], &num)) {
```
sodabrew: It wasn't immediately clear to me why `my_big2ll` would return "truthy" and then proceed on the workaround path. Maybe flip the cases around? I think this would be more readable:

``` C
if (my_big2ll(argv[i], &num) == 0) {
                bind_buffers[i].buffer_type = MYSQL_TYPE_LONGLONG;
                bind_buffers[i].buffer = xmalloc(sizeof(long long int));
                *(LONG_LONG*)(bind_buffers[i].buffer) = num;
} else {
    /* The bignum was larger than we can fit in LONG_LONG, send it as a string */
    ... code as written ...
}
```

spec/mysql2/error_spec.rb
```diff
@@ -34,7 +34,7 @@
       end
     end
 
-    let(:invalid_utf8) { "\xE5\xC6\x7D\x1F" }
+    let(:invalid_utf8) { ["e5c67d1f"].pack('H*').force_encoding(Encoding::UTF_8) }
```
sodabrew: I merged this from the other PR, you can rebase to master to include this commit.

sodabrew:
Thank you! Will this also work in the future merged Integer / FixNum / BigNum world of Ruby 2.4.0?

nurse:
Yeah, `TYPE(v) == T_BIGNUM` still works with Ruby 2.4.

sodabrew:
Ruby 1.8.7 does have `rb_big_cmp`, but it is static: http://rxr.whitequark.org/mri/source/bignum.c?v=1.8.7-p374#1107

You could call Bignum#<=> to get the same effect?

``` C
/* elsewhere in the file */
static ID id_cmp;
id_cmp = rb_intern("<=>");
/**/

#ifdef HAVE_RB_BIG_CMP
  VALUE less_than_llong_min = rb_big_cmp(bignum, LL2NUM(LLONG_MIN));
#else
  VALUE less_than_llong_min = rb_funcall(bignum, id_cmp, 1, LL2NUM(LLONG_MIN));
#endif
  if (less_than_llong_min == INT2FIX(-1)) {
    // code
  } else {
    // code
  }
```

nurse:
`rb_funcall` can be an alternative but it is slow.

I noticed `len == 8 && nlz_bits == 0 && rb_absint_singlebit_p(bignum)` is the best condition because it doesn't allocate extra object.
For older rubies I'm using `rb_big_and` or `rb_big_cmp` now, but `rb_funcall` may be better for the people maintain the code. I'll fix them if you prefer that.

ext/mysql2/statement.c
```diff
@@ -204,6 +204,44 @@ static void set_buffer_for_string(MYSQL_BIND* bind_buffer, unsigned long *length
...
+    size_t len;
+    int nlz_bits = 0;
+#ifdef HAVE_RB_ABSINT_SIZE
+    len = rb_absint_size(bignum, &nlz_bits);
+#else
+    len = RBIGNUM_LEN(bignum) * SIZEOF_BDIGITS;
+#endif
+    if (len > 8) goto overflow;
+    if (RBIGNUM_POSITIVE_P(bignum)) {
+        num = rb_big2ull(bignum);
```
kamipo: Looks like if `BIGSIZE(bignum) > SIZEOF_LONG_LONG`, should be `goto overflow`.
http://rxr.whitequark.org/mri/source/bignum.c#3587

kamipo: Oh, enough to `if (len > 8) goto overflow;`?

nurse: Yes. `len > 8` means "Bignum's raw size is larger than `8` byte.
It is the same meaning with `BIGSIZE(bignum) > SIZEOF_LONG_LONG`.

Anyway the spec should have `int64_max3 = (1 << 64) - 1`; I'll add it tomorrow.

ext/mysql2/statement.c
```diff
@@ -204,6 +204,44 @@ static void set_buffer_for_string(MYSQL_BIND* bind_buffer, unsigned long *length
...
+{
+  unsigned LONG_LONG num;
+  size_t len;
+  int nlz_bits = 0;
+#ifdef HAVE_RB_ABSINT_SIZE
+  len = rb_absint_size(bignum, &nlz_bits);
+#else
+  len = RBIGNUM_LEN(bignum) * SIZEOF_BDIGITS;
+#endif
+  if (len > 8) goto overflow;
```
sodabrew: Length 8 is intended to be the same as `sizeof(long long int)` right? Would it make sense for the integer size to be passed in as an argument?

ext/mysql2/statement.c
```diff
@@ -204,6 +204,44 @@ static void set_buffer_for_string(MYSQL_BIND* bind_buffer, unsigned long *length
...
+  else {
+    if (len == 8 && nlz_bits == 0 &&
+        /* only -0x8000000000000000 is safe if `len == 8 && nlz_bits == 0` */
+#ifdef HAVE_RB_ABSINT_SINGLEBIT_P
+        !rb_absint_singlebit_p(bignum)
+#elif defined(HAVE_RB_BIG_AND)
+        rb_big_and(bignum, LL2NUM(LLONG_MAX)) != INT2FIX(0)
+#else
+        rb_big_cmp(bignum, LL2NUM(LLONG_MIN)) == INT2FIX(-1)
+#endif
```
sodabrew: Could you add a comment which Ruby is the target for each ifdef branch?

nurse: Added at 5a3ec97c1c29a4de979c526c46c6366b88bd689b

ext/mysql2/statement.c
nurse: fixed!

nurse:
@sodabrew I reflected your comments. What do you think about current one?

sodabrew:
Sorry for the long delay on this PR, and it's run into a merge conflict since the recent `ssl_mode` PR added other items to extconf.rb. Do you have time to rebase this PR? If not I can do a rebase and squash merge mysql.

nurse:
Hmm, rebased but it fails. I'll investigate later...

sodabrew:
Thanks! There's an interaction with #783 that results in the values getting cast as BigDecimal instead of Bigint. This reminds me of #563 and #58.

sodabrew:
I think #783 actually correctly fixed the big number types, and you can remove the `.to_s` that look like a workaround in the test?

```
diff --git a/spec/mysql2/statement_spec.rb b/spec/mysql2/statement_spec.rb
index 517bd31..3d00535 100644
--- a/spec/mysql2/statement_spec.rb
+++ b/spec/mysql2/statement_spec.rb
@@ -78,7 +78,7 @@ RSpec.describe Mysql2::Statement do
     int64_min2 = -(1 << 64) + 1
     int64_min3 = -0xC000000000000000
     result = stmt.execute(int64_max1, int64_max2, int64_max3, int64_min1, int64_min2, int64_min3)
-    expect(result.to_a).to eq(['max1' => int64_max1.to_s, 'max2' => int64_max2.to_s, 'max3' => int64_max3.to_s, 'min1' => int64_min1.to_s, 'min2' => int64_min2.to_s, 'min3' => int6
+    expect(result.to_a).to eq(['max1' => int64_max1, 'max2' => int64_max2, 'max3' => int64_max3, 'min1' => int64_min1, 'min2' => int64_min2, 'min3' => int64_min3])
   end
```

nurse:
@sodabrew Fixed as you says removed `to_s`. It still fails on some environments but seems not related with this PR...

PR #632, opened 2015-06-10
Remove `{1.8.7,ree}` support

tamird:
Not sure if you're prepared to do this, but thought I'd put it out there.

[Review on Reviewable](https://reviewable.io/reviews/brianmario/mysql2/632)

PR #766, opened 2016-06-28
Drop Ruby 1.8.7 and REE support

nurse:
Their CI statuses are badly broken now: https://travis-ci.org/brianmario/mysql2/builds/124159847
(others are also broken but some of them can be fixed by https://github.com/brianmario/mysql2/pull/765)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,10 +24,6 @@ rvm:
   - ruby-head
 matrix:
   include:
-    - rvm: 1.8.7
-      dist: precise
-    - rvm: ree
-      dist: precise
     - rvm: rbx-2
       dist: precise
     - rvm: 2.0.0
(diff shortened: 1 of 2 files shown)


justincase:
Related: #632

sodabrew:
There's a segfault bug in #759 that should resolve the other test failures including 1.8.7 and ree. It's not a "normal case" in production, but the unit tests intentionally trigger it. I will remove the oldest Rubies eventually and pull out all of the encoding workarounds, but not today :)

kamipo:
I need to fix the issue #764 for https://github.com/rails/rails/pull/25230.
But I heard from @nurse that Ruby 1.8.7 does not have `rb_big_cmp`.
Do we need to support Ruby 1.8.7 and ree for mysql2 0.4.x and fix #764 for the ruby's?

PR #767, opened 2016-06-28
Deflake CI

tamird:


.travis.yml
```diff
@@ -2,9 +2,7 @@ sudo: required
 dist: trusty
 language: ruby
 bundler_args: --without benchmarks development
-# Temporary workaround for broken Rubygems on Travis
```
sodabrew: Every so often, a new Rubygems release hoses tests. I want to leave the pin. No problem updating the pinned version though.

.travis.yml
```diff
@@ -63,11 +61,11 @@ matrix:
...
-        apt:
-          packages:
-          - mysql-server-5.7
-          - mysql-client-core-5.7
-          - mysql-client-5.7
+        # apt:
+        #   packages:
+        #   - mysql-server-5.7
+        #   - mysql-client-core-5.7
+        #   - mysql-client-5.7
```
sodabrew: I probably merged this by mistake, I remember working on it but it wasn't working as expected.

.travis_mysql57.sh
```diff
@@ -1,8 +1,7 @@
 #!/usr/bin/env bash
 
-set -eu
+set -eux
 
-service mysql stop
 apt-get purge '^mysql*' 'libmysql*'
```
sodabrew: This needs a `-y` added: https://travis-ci.org/brianmario/mysql2/jobs/140877169

spec/mysql2/statement_spec.rb
```diff
@@ -689,9 +691,9 @@
   context 'close' do
     it 'should free server resources' do
       stmt = @client.prepare 'SELECT 1'
-      expect { stmt.close }.to change {
-        @client.query("SHOW STATUS LIKE 'Prepared_stmt_count'").first['Value'].to_i
-      }.by(-1)
+      GC.disable
```
sodabrew: What's the failure this fixes?

tamird: Sometimes this fails with "expected to change by -1, but changed by -2" but this patch didn't fix it =/

.travis_mysql57.sh
tamird: Done.

.travis.yml
tamird: Removed this commit.

.travis.yml
tamird: Probably. I've mentioned the upstream issue in the commit message as a bread crumb for future work.

sodabrew:
Oh, now I remember - the AppArmor situation changed, and I couldn't figure out whether to update or just remote this bit:
```
 # Replace the final line of the mysql apparmor, allowing /etc/mysql/*.pem
 sed -ie '$ s|}|\
   /etc/mysql/*.pem r,\
 }|' /etc/apparmor.d/usr.sbin.mysqld
 service apparmor restart
```

sodabrew:
Ok, removing AppArmor was good. The next problem I remember now is that MySQL 5.7 defaults to a randomized root password, and I couldn't figure out how to automate either resetting that password to blank or using that password to create an alternate account for tests and running tests as a non-root user.

tamird:
Sigh. I've been messing with this damned apt package for a few hours and I can't reproduce the random-password behaviour locally. Worse than that, I can't find travis' images anywhere so that I can properly reproduce. Any suggestions?

sodabrew:
I got stuck at the same point and haven't revisited in two months :/  thanks a million for giving it a solid try!

My concern is that MySQL 5.7 is really rolling out there in production environments and I don't want to just ignore it for testing purposes, but it's being such a PITA that I can't see how to get it in CI yet.

tamird:
I must be taking crazy pills. In `.travis_setup.sh` we execute
```
sudo mysql -u root -e "CREATE USER '$USER'@'localhost'"
sudo mysql -u root -e "GRANT ALL ON test.* TO '$USER'@'localhost'"`
```
after install mysql 5.7, and these commands work! what the hell is going on?

tamird:
Hard to believe, but here we go, it's working!

The mystery was that `mysql -h localhost` will still use unix sockets, and this socket_auth plugin thing was making it impossible to connect any other way. I've linked to a percona blog post from the install script.

sodabrew:
Gah, I hate that interpretation of `localhost` as "check the local socket first". Great find!

tamird:
What do you think; can we merge this guy?

PR #778, opened 2016-07-29
Stop supporting older Rubies

thedrow:
I think that once we hit 1.0 we should stop supporting older Rubies since they are no longer maintained.
If it's possible to stop supporting them now it'd be great.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,14 +21,9 @@ rvm:
   - 2.2
   - 2.1
   - 2.0.0
-  - 1.9.3
   - ruby-head
 matrix:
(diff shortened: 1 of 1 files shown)


justincase:
Checkout #632.

thedrow:
I have to say I agree. It's not productive to support these.

sodabrew:
Thanks - however dropping 1.8.7 from tests should happen at the same time as removing `#ifdef HAVE_ENCODING` blocks and other support for older syntax. There are open PR's for this, but I'm not in any rush.

PR #791, opened 2016-10-11
Avoid type error when using connect_timeout, read_timeout and write_timeout in a DATABASE_URL env variable

helgeblod:
Why:

* When using a DATABASE_URL env variable with one or more of  these
  params, a type-error is triggered

Example: 

`DATABASE_URL="mysql2://<database>:<password>@<host>:<port>/<db>?encoding=utf8&pool=1&connect_timeout=2&read_timeout=5&write_timeout=5"`

Results in:

```
TypeError: wrong argument type String (expected Fixnum)
/usr/local/bundle/gems/mysql2-0.4.4/lib/mysql2/client.rb:39:in `connect_timeout='
/usr/local/bundle/gems/mysql2-0.4.4/lib/mysql2/client.rb:39:in `block in initialize'
/usr/local/bundle/gems/mysql2-0.4.4/lib/mysql2/client.rb:33:in `each'
/usr/local/bundle/gems/mysql2-0.4.4/lib/mysql2/client.rb:33:in `initialize'
...
```

This fix solves this issue. I have reproduced this in alpine-linux inside a docker container, and as far as I can see it should not break anything.

### DIFF
--- a/lib/mysql2/client.rb
+++ b/lib/mysql2/client.rb
@@ -36,7 +36,7 @@ module Mysql2
         when :reconnect, :local_infile, :secure_auth, :automatic_close
           send(:"#{key}=", !!opts[key]) # rubocop:disable Style/DoubleNegation
         when :connect_timeout, :read_timeout, :write_timeout
-          send(:"#{key}=", opts[key]) unless opts[key].nil?
+          send(:"#{key}=", opts[key].to_i) unless opts[key].nil?
         else
           send(:"#{key}=", opts[key])
         end


sodabrew:
Thanks! I'll land this once Travis tests finish.

helgeblod:
Looks like the tests for ruby 1.9.3 and rbx-2 fails, but looks unrelated to my fix.

sodabrew:
I'm reminded of this behavior difference in Ruby:

```
$ irb
irb(main):001:0> Integer('5junk')
ArgumentError: invalid value for Integer(): "0junk0"
	from (irb):4:in `Integer'
	from (irb):4
	from /usr/bin/irb:12:in `<main>'
irb(main):002:0> '5junk'.to_i
=> 5
```

sodabrew:
Turns out you're in good company:
https://github.com/rails/rails/issues/26556
 https://github.com/rails/rails/pull/26559

helgeblod:
Might be that using `Integer(opts[key])` might be a better approach to avoid weird behaviour, so I'm glad to resubmit that change if you want. Say if you omit a `&`in the example variable: `DATABASE_URL="mysql2://<database>:<password>@<host>:<port>/<db>?encoding=utf8&pool=1&connect_timeout=2read_timeout=5&write_timeout=5"`. Then this is better 
```
ArgumentError: invalid value for Integer(): "2read_timeout=5"
```
Than setting `connect_timeout=2` (as would be the result with my fix). Should I make another pr fixing this?

PR #676, opened 2015-09-08
Add connection option :ssl for all-defaults SSL connection

sodabrew:

PR #793, opened 2016-10-13
Add ability to set ssl_mode.

apalmblad:
In mysql 5.7.11, an ssl_mode option becomes available to set, and setting values to it before the connection happens can allow connections to servers that I could not otherwise connect to using 5.7.11.

http://dev.mysql.com/doc/refman/5.7/en/mysql-options.html

It fails with a warning when it is not supported; is that the behaviour you'd want?

### DIFF
--- a/README.md
+++ b/README.md
@@ -203,7 +203,7 @@ Mysql2::Client.new(
   :reconnect = true/false,
   :local_infile = true/false,
   :secure_auth = true/false,
-  :ssl_mode = SSL_MODE_DISABLED / SSL_MODE_PREFERRED / REQUIRED / VERIFY_CA / VERIFY_IDENTITY,
+  :ssl_mode = :disabled / :preferred / :required / :verify_ca / :verify_identity,
   :default_file = '/path/to/my.cfg',
   :default_group = 'my.cfg section',
   :init_command => sql
--- a/README.md
+++ b/README.md
@@ -203,6 +203,7 @@ Mysql2::Client.new(
   :reconnect = true/false,
   :local_infile = true/false,
   :secure_auth = true/false,
+  :ssl_mode = "DISABLED" | "PREFERRED" | "REQUIRED" | "VERIFY_CA" | "VERIFY_IDENTITY",
   :default_file = '/path/to/my.cfg',
   :default_group = 'my.cfg section',
   :init_command => sql
--- a/README.md
+++ b/README.md
@@ -203,7 +203,7 @@ Mysql2::Client.new(
   :reconnect = true/false,
   :local_infile = true/false,
   :secure_auth = true/false,
-  :ssl_mode = "DISABLED" | "PREFERRED" | "REQUIRED" | "VERIFY_CA" | "VERIFY_IDENTITY",
+  :ssl_mode = SSL_MODE_DISABLED / SSL_MODE_PREFERRED / REQUIRED / VERIFY_CA / VERIFY_IDENTITY,
   :default_file = '/path/to/my.cfg',
   :default_group = 'my.cfg section',
   :init_command => sql
--- a/ext/mysql2/client.c
+++ b/ext/mysql2/client.c
@@ -47,7 +47,6 @@ VALUE rb_hash_dup(VALUE other) {
   #define MYSQL_LINK_VERSION MYSQL_SERVER_VERSION
 #endif
 
-
 /*
  * used to pass all arguments to mysql_real_connect while inside
  * rb_thread_call_without_gvl
@@ -83,6 +82,7 @@ struct nogvl_select_db_args {
   MYSQL *mysql;
   char *db;
 };
+
 static VALUE rb_set_ssl_mode_option(VALUE self, VALUE setting) {
   unsigned long version = mysql_get_client_version();
 
--- a/ext/mysql2/client.c
+++ b/ext/mysql2/client.c
@@ -90,13 +90,13 @@ static VALUE rb_set_ssl_mode_option(VALUE self, VALUE setting) {
     rb_warn( "Your mysql client library does not support setting ssl_mode; full support comes with 5.7.11." );
     return Qnil;
   }
-  GET_CLIENT(self); 
   Check_Type( setting, T_FIXNUM);
   if( NIL_P( setting ) ) {
     rb_raise(cMysql2Error, "ssl_mode= takes DISABLED, PREFERRED, REQUIRED, VERIFY_CA, VERIFY_IDENTITY, you passed nil" );
   }
-  int val = NUM2INT( setting );
 #ifdef HAVE_CONST_MYSQL_OPT_SSL_ENFORCE
+  GET_CLIENT(self); 
+  int val = NUM2INT( setting );
   if( version >= 50703 && version < 50711 ) {
     if( val == SSL_MODE_DISABLED || val == SSL_MODE_REQUIRED ) {
       bool b = ( val == SSL_MODE_REQUIRED );
@@ -110,6 +110,7 @@ static VALUE rb_set_ssl_mode_option(VALUE self, VALUE setting) {
   }
 #endif
 #ifdef FULL_SSL_MODE_SUPPORT
+  GET_CLIENT(self); 
 
   if( val != SSL_MODE_DISABLED && val != SSL_MODE_PREFERRED && val != SSL_MODE_REQUIRED && val != SSL_MODE_VERIFY_CA && val != SSL_MODE_VERIFY_IDENTITY ) {
     rb_raise(cMysql2Error, "ssl_mode= takes DISABLED, PREFERRED, REQUIRED, VERIFY_CA, VERIFY_IDENTITY, you passed: %d", val );
--- a/ext/mysql2/client.c
+++ b/ext/mysql2/client.c
@@ -22,16 +22,16 @@ static ID intern_brackets, intern_merge, intern_merge_bang, intern_new_with_args
 #define SSL_MODE_DISABLED 0
 #endif
 #ifndef SSL_MODE_PREFERRED
-#define SSL_MODE_PREFERRED 0
+#define SSL_MODE_PREFERRED 1
 #endif
 #ifndef SSL_MODE_REQUIRED
-#define SSL_MODE_REQUIRED 0
+#define SSL_MODE_REQUIRED 2
 #endif
 #ifndef SSL_MODE_VERIFY_CA
-#define SSL_MODE_VERIFY_CA 0
+#define SSL_MODE_VERIFY_CA 3
 #endif
 #ifndef SSL_MODE_VERIFY_IDENTITY
-#define SSL_MODE_VERIFY_IDENTITY 0
+#define SSL_MODE_VERIFY_IDENTITY 4
 #endif
 
 #ifndef HAVE_RB_HASH_DUP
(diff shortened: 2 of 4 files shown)


ext/mysql2/client.c
```diff
@@ -82,7 +82,35 @@ struct nogvl_select_db_args {
...
+    return Qnil;
+  }
 
+  GET_CLIENT(self); 
+  if( NIL_P( str ) ) {
+    rb_raise(cMysql2Error, "ssl_mode= takes DISABLED, PREFERRED, REQUIRED, VERIFY_CA, VERIFY_IDENTITY, you passed nil" );
+  }
+  char *val = StringValueCStr( str );
+  unsigned int setting;
+  if( strncasecmp( val, "DISABLED", 9 ) == 0 ) {
```
sodabrew: Much easier to do this string comparison in Ruby.

README.md
```diff
@@ -203,6 +203,7 @@ Mysql2::Client.new(
   :reconnect = true/false,
   :local_infile = true/false,
   :secure_auth = true/false,
+  :ssl_mode = "DISABLED" | "PREFERRED" | "REQUIRED" | "VERIFY_CA" | "VERIFY_IDENTITY",
```
sodabrew: The Flags above are constant integers and can be OR-ed together. Based on http://dev.mysql.com/doc/refman/5.7/en/secure-connection-options.html#option_general_ssl it looks like these are mutually-exclusive values -- e.g. VERIFY_CA implied REQUIRED and does not need both options to be specified?

sodabrew:
Well that was quick for `--ssl`, it's deprecated already client-side: http://dev.mysql.com/doc/refman/5.7/en/secure-connection-options.html#option_general_ssl so that means #676 is effectively deprecated before it started.

README.md
apalmblad: Right, I suppose what I really meant would have been to use slashes, to indicate that one of the values should be set.

ext/mysql2/client.c
apalmblad: Fair, something similar tot he existing code in https://github.com/brianmario/mysql2/blob/master/lib/mysql2/client.rb#L90?

Do you want something that would accept strings or ints?  Constant-wise, should we name the same as the constants in C, or should we shorten them?

README.md
sodabrew: I was thinking that these could be symbols, `:disabled` (or `false`), `:preferred`, `:required`, `:verify_ca`, `:verify_identity`. That said, there's already `:ssl_verify` which means `:verify_ca` - there's a bunch of overlap that should be resolved as much as possible.

apalmblad:
Happy to do some work.

How do handle backwards compatibility with libraries?  Better to define missing constants that are intended to be put on the client always, or just when the underlying library has them.  That is, should users depend on Mysql::Client::SSL_MODE_DISABLED to always be around or not?

README.md
apalmblad: I'd agree with your reading that they are mutually exclusive.

ext/mysql2/extconf.rb
```diff
@@ -62,6 +62,14 @@ def asplode(lib)
...
   warn "-----\nUsing mysql_config at #{mc}\n-----"
   ver = `#{mc} --version`.chomp.to_f
+  major, minor, inc = `#{mc} --version`.chomp.split('.').map( &:to_i )
+  if major > 5 || major == 5 && minor == 7 && inc >= 11
+    $CFLAGS << ' -D SSL_MODE_SUPPORT'
+  elsif major == 5 && minor == 7 && inc >= 3 && inc < 11
+    $CFLAGS << ' -D SSL_TOGGLE_SUPPORT'
+  else
+    $CFLAGS << ' -D NO_SSL_MODE_SUPPORT'
+  end
```
apalmblad: This is pretty gross and I'd love it if someone had a nicer idea.

ext/mysql2/client.c
```diff
@@ -1464,4 +1501,17 @@ void init_mysql2_client() {
   rb_const_set(cMysql2Client, rb_intern("BASIC_FLAGS"),
       LONG2NUM(CLIENT_BASIC_FLAGS));
 #endif
+#ifdef SSL_MODE_SUPPORT
```
sodabrew: '#ifdef SSL_MODE_DISABLED` etc. because these values from mysql mysql.h.

ext/mysql2/client.c
```diff
@@ -1464,4 +1501,17 @@ void init_mysql2_client() {
...
+  rb_const_set(cMysql2Client, rb_intern("SSL_MODE_DISABLED"),INT2NUM( SSL_MODE_DISABLED ) );
+  rb_const_set(cMysql2Client, rb_intern("SSL_MODE_PREFERRED"),INT2NUM( SSL_MODE_PREFERRED ) );
+  rb_const_set(cMysql2Client, rb_intern("SSL_MODE_REQUIRED"),INT2NUM( SSL_MODE_REQUIRED ) );
+  rb_const_set(cMysql2Client, rb_intern("SSL_MODE_VERIFY_CA"),INT2NUM( SSL_MODE_VERIFY_CA ) );
+  rb_const_set(cMysql2Client, rb_intern("SSL_MODE_VERIFY_IDENTITY"),INT2NUM( SSL_MODE_VERIFY_IDENTITY ) );
+#endif
+#ifdef SSL_TOGGLE_SUPPORT
+  #define SSL_MODE_DISABLED 1
```
sodabrew: These constants should still be available in earlier MySQL 5.7 headers.

lib/mysql2/client.rb
```diff
@@ -47,6 +47,11 @@ def initialize(opts = {})
 
       ssl_options = opts.values_at(:sslkey, :sslcert, :sslca, :sslcapath, :sslcipher)
       ssl_set(*ssl_options) if ssl_options.any?
+      if opts[:ssl_mode]
+        m = parse_ssl_mode( opts[:ssl_mode] )
+        self.ssl_mode = m if m
+      end
```
sodabrew: I think this will suffice:
```
ssl_mode = parse_ssl_mode(opts[:ssl_mode]) if opts[:ssl_mode]
```

sodabrew:
There's a precedent in the case of SECURE_CONNECTION of creating the missing constants with a 0 value. https://github.com/brianmario/mysql2/blob/master/ext/mysql2/client.c#L1435

ext/mysql2/client.c
apalmblad: SSL_MODE_DISABLED, etc are not #define'd in mysql.h, they're enums... hence the ugly hack with the version stuff in extconf.rb

apalmblad: See https://github.com/mysql/mysql-server/blob/5.7/include/mysql.h#L238

lib/mysql2/client.rb
```diff
@@ -87,6 +92,23 @@ def initialize(opts = {})
       connect user, pass, host, port, database, socket, flags
     end
 
+    def parse_ssl_mode( m )
+      return nil if m.nil?
+      if m.is_a?( String ) || m.is_a?( Symbol )
+        m.to_s.upcase!
```
apalmblad: I've made your changes, that will remove the ability to pass in an integer, ala `ssl_mode: Mysql2::Client::SSL_MODE_DISABLED`.

lib/mysql2/client.rb
apalmblad: Works for me, we might end up passing nil values into ssl_mode=, and getting an error, but thats fine.

ext/mysql2/extconf.rb
apalmblad: mkmf func | enum member result  (ie. SSL_MODE_PREFRRED ) | enum result (ie mysql_ssl_mode)
-----------|---------------------------|------
have_const | true | false
have_func | true | false
have_header | true | false
have_library | false | false
have_macro | false | false
have_struct_member | false | n/a 
have_type | false | false
have_var | false | false

After hve_type failed, I tried everything.  Yay, learning about mkmf.  I think have_const is the winner.

ext/mysql2/client.c
apalmblad: :ok:

README.md
```diff
@@ -203,6 +203,7 @@ Mysql2::Client.new(
   :reconnect = true/false,
   :local_infile = true/false,
   :secure_auth = true/false,
+  :ssl_mode = SSL_MODE_DISABLED / SSL_MODE_PREFERRED / REQUIRED / VERIFY_CA / VERIFY_IDENTITY,
```
sodabrew: Can remove `SSL_MODE_` prefix from the first two.

ext/mysql2/client.c
```diff
@@ -47,6 +47,7 @@ VALUE rb_hash_dup(VALUE other) {
   #define MYSQL_LINK_VERSION MYSQL_SERVER_VERSION
 #endif
 
+
```
sodabrew: Nit: remove extra blank line.

ext/mysql2/client.c
```diff
@@ -82,7 +83,47 @@ struct nogvl_select_db_args {
   MYSQL *mysql;
   char *db;
 };
+static VALUE rb_set_ssl_mode_option(VALUE self, VALUE setting) {
```
sodabrew: Nit: add blank line above.

ext/mysql2/extconf.rb
```diff
@@ -81,8 +91,10 @@ def asplode(lib)
 
 if have_header('mysql.h')
   prefix = nil
+  add_ssl_defines('mysql.h')
 elsif have_header('mysql/mysql.h')
   prefix = 'mysql'
+  add_ssl_defines('mysql/mysql.h')
```
sodabrew: Using pattern for `errmsg.h`, this could be written as:

```
add_ssl_defines([prefix, 'mysql.h'].compact.join '/')
```

Upside: only one line of code, and not inside a conditional block (move it below the if-have-header block). Downside: looks complicated.

lib/mysql2/client.rb
```diff
@@ -87,6 +88,18 @@ def initialize(opts = {})
...
 
+    def parse_ssl_mode(mode)
+      m = mode.to_s.upcase
+      if m.start_with?('SSL_MODE_')
+        return Mysql2::Client.const_get(m) if Mysql2::Client.const_defined?(m)
+      else
+        x = 'SSL_MODE_' + m
+        return Mysql2::Client.const_get(x) if Mysql2::Client.const_defined?(x)
+      end
+      warn "ssl_mode must be one of SSL_MODE_DISABLED, SSL_MODE_PREFERRED, SSL_MODE_REQUIRED, SSL_MODE_VERIFY_CA, SSL_MODE_VERIFY_IDENTITY"
```
sodabrew: For consistency with `parse_flags_array` this should be `warn "Unknown MySQL ssl_mode flag: #{f}"`

ext/mysql2/client.c
```diff
@@ -82,7 +83,47 @@ struct nogvl_select_db_args {
...
+    }
+  }
+#endif
+#ifdef FULL_SSL_MODE_SUPPORT
+  GET_CLIENT(self); 
+  int val = NUM2INT( setting );
+
+  if( val != SSL_MODE_DISABLED && val != SSL_MODE_PREFERRED && val != SSL_MODE_REQUIRED && val != SSL_MODE_VERIFY_CA && val != SSL_MODE_VERIFY_IDENTITY ) {
+    rb_raise(cMysql2Error, "ssl_mode= takes DISABLED, PREFERRED, REQUIRED, VERIFY_CA, VERIFY_IDENTITY, you passed: %d", val );
+  }
```
sodabrew: I think the values are fully vetted at this point, and should just pass it through?

ext/mysql2/client.c
```diff
@@ -82,7 +83,47 @@ struct nogvl_select_db_args {
...
+  GET_CLIENT(self); 
+  int val = NUM2INT( setting );
+  if( version >= 50703 && version < 50711 ) {
+    if( val == SSL_MODE_DISABLED || val == SSL_MODE_REQUIRED ) {
+      bool b = ( val == SSL_MODE_REQUIRED );
+      int result = mysql_options( wrapper->client, MYSQL_OPT_SSL_ENFORCE, &b );
+      return INT2NUM(result);
+      
+    } else {
+      rb_warn( "Mysql client libraries between 5.7.3 and 5.7.10 do not support other values for ssl_mode" );
```
sodabrew: Suggested text: `"MySQL client libraries between 5.7.3 and 5.7.10 only support SSL_MODE_DISABLED and SSL_MODE_REQUIRED"`

ext/mysql2/client.c
```diff
@@ -82,7 +83,47 @@ struct nogvl_select_db_args {
...
+static VALUE rb_set_ssl_mode_option(VALUE self, VALUE setting) {
+  unsigned long version = mysql_get_client_version();
+
+  if( version < 50703 ) {
+    rb_warn( "Your mysql client library does not support setting ssl_mode; full support comes with 5.7.11." );
+    return Qnil;
+  }
+  Check_Type( setting, T_FIXNUM);
+  if( NIL_P( setting ) ) {
+    rb_raise(cMysql2Error, "ssl_mode= takes DISABLED, PREFERRED, REQUIRED, VERIFY_CA, VERIFY_IDENTITY, you passed nil" );
```
sodabrew: The values are already vetted in Ruby, so this check could go away.

ext/mysql2/client.c
```diff
@@ -83,6 +83,43 @@ struct nogvl_select_db_args {
   char *db;
 };
 
+static VALUE rb_set_ssl_mode_option(VALUE self, VALUE setting) {
+  unsigned long version = mysql_get_client_version();
+
+  if( version < 50703 ) {
+    rb_warn( "Your mysql client library does not support setting ssl_mode; full support comes with 5.7.11." );
+    return Qnil;
+  }
```
sodabrew: C style nit: spaces outside of parens and spaces after commas, e.g.:

```
if (foo || bar) {
   qux(argone, argtwo);
}

**Comment by sodabrew** [comment]
This is awesome. Thanks for sticking with it through multiple code reviews! Looking forward to merge after last nits resolved!

**Comment by apalmblad** [comment]
Happy to help; hopefully there's nothing more after this.

**Comment by apalmblad** [comment]
Ugh, rubocop doesn't like that the initialize method is now 51 lines.  

I could find something to extract, but that seems beyond the scope of this PR, so I guess I'll disable that check?  Otherwise, I'd suggest pulling out the ssl option setting, and maybe the loop on keys that map to setters?

File: lib/mysql2/client.rb [hunk 17]
```diff
@@ -18,6 +18,9 @@ def self.default_query_options
       }
     end
 
+    # rubocop:disable Metrics/AbcSize
+    # rubocop:disable Metrics/CyclomaticComplexity
+    # rubocop:disable Metrics/MethodLength
```
sodabrew: Oh, no, please increase the limits in .rubocop_todo.yml - or just remove this, I need to update the limits for another pending PR anyways.

apalmblad: Done!

PR #796, opened 2016-10-17
Add a Mysql2::Client#closed? method

dylanahsmith:
I would like to be able to check if the mysql connection was closed without actually pinging the server, however, I don't see that easily state accessible on the Mysql2::Client.

I did find a workaround which works

``` ruby
def connected?
  !!@connection.socket
rescue Mysql2::Error
  false
end
```

but it would be more convenient to be able to just do `@connection.connected?` which this PR adds

### DIFF
--- a/ext/mysql2/client.c
+++ b/ext/mysql2/client.c
@@ -472,6 +472,16 @@ static VALUE rb_mysql_client_close(VALUE self) {
   return Qnil;
 }
 
+/* call-seq:
+ *    client.closed?
+ *
+ * @return [Boolean]
+ */
+static VALUE rb_mysql_client_closed(VALUE self) {
+  GET_CLIENT(self);
+  return CONNECTED(wrapper) ? Qfalse : Qtrue;
+}
+
 /*
  * mysql_send_query is unlikely to block since most queries are small
  * enough to fit in a socket buffer, but sometimes large UPDATE and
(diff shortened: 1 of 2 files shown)


sodabrew:
My concern is that `connected?` might return true, but `ping` would return false or cause a reconnect in situations where the server has timed out the client but the client hasn't realized it yet.

That said, it is a representation whether we _think_ we're connected, even if we're not.

dylanahsmith:
I'm fine with it not knowing if the next request will pass or fail, I really just want a cheap way to see if we know the next query will fail, like if `close` was already called on the connection.  We don't want the overhead of unnecessary pings from all our servers connected to the database in the common case where we are still connected and are fine with waiting until a query fails to be able to detect that we aren't connected.

Would `closed?` be a better method name?

sodabrew:
The indeterminate state of "connected, at least we think so" matches reality, so it might be the better name:

```
if (!client.closed?)
   # Try the query, but be sure to catch if the connection blows up spontaneously :)
end
```

dylanahsmith:
Is anything else needed to get this PR merged?

The CI failures is just from flakiness in the test setup, not related to the code itself.

pschambacher:
This would actually solve the issue I just opened: https://github.com/brianmario/mysql2/issues/827

👍

dylanahsmith:
Rebased now that https://github.com/brianmario/mysql2/pull/847 has been merged

Issue #600, opened 2015-03-18
Multiple statements + Streaming may need a mysql2 API adjustment


sodabrew:
The `:streaming` flag controls whether the `query` method should call the mysqlclient C API's `use_result` or `store_result` for the first result.

When multiple results are enabled, this still happens for the first result, but there's no equivalent decision for subsequent results. The multiple statements API includes `next_result`, `store_result`, and `abandon_result!`.

Idea #1: Add `client.use_result` and expose the difference in behavior to the user.

Idea #2: Secretly call `mysql_use_result` from `client.store_result` when streaming was enabled on the original query.

Idea #3: Remove `Mysql2::Client#store_result` from the mysql2 gem API, hide the difference inside the client state, and have `Mysql2::Client#next_result` return a `Mysql2::Result` directly. I'm not sure where `abandon_result!` would need to be moved, however. Maybe `Mysql2::Result#abandon!` ?

brianmario:
I kinda like #3 the best, even though it's the most complex (I think).

This next topic is probably best discussed in another issue but:

-------------------------------

Having spent a lot more time dealing with the wire protocol lately, I'm starting to lean toward just _always_ using `mysql_use_result` and leaving the whole "buffer everything locally" or "stream" decision up to the caller. If they want it all buffered, they can just iterate over all the results as fasts as you can. Then deal with processing. Otherwise, just iterate over each result and process them as you get them. This would mean the socket buffer would fill and stack up all the way up the network chain until the server's call to `write(2)` would eventually block. And I think that's all ok?

For what it's worth, `ActiveRecord` reads all the results into memory to build `ActiveRecord::Result` objects so there's no difference between us using `use_result` or `store_result`. But even still, for people using mysql2 directly, they can do what I suggested above if they want to emulate `store_result`.

-------------------------------

Would that change how we think about this API change? (which I think should happen anyway)

sodabrew:
I'm imagining a benchmark to see the cost of store_result vs. use_result, where the default behavior on the mysql2 side would be pulling the entire result set into Ruby memory space.

Today's defaults:

* mysql2 calls mysql_query and then mysql_store_result
* libmysqlclient reads the result into its data structures
* mysql2 calls mysql_fetch_row to retrieve each row into the data structure in the Mysql2::Result

Proposed new defaults:

* mysql2 calls mysql_query and then mysql_use_result
* mysql2 calls mysql_fetch_row to retrieve each row into the Mysql2::Result

[...]

PR #807, opened 2016-11-19
RFC: non-blocking behavior for `next_result`

osheroff:
We were a multi-statement query in our, and discovered that when queries other than the first query blocked, (`SELECT 1; SELECT SLEEP(3)`)  the call to `next_result` became completely uninterruptible.  This was fixed well enough by by wrapping the `next_result` call in `rb_thread_call_without_gvl`, but we subsequently found that the `next_result` call still wouldn't respond to `Thread#raise`, which we needed in order to perform some timeout logic.  

Went down the rabbit hole further to find that the `vio` suite of functions will over-read a buffer, so mysql2's methodology of calling `rb_wait_for_single_fd` really won't work reliably for `next_result` -- reading the first result in the `do_query` path will trigger a read of the result of the results (if they're ready), meaning that we'll poll forever; ugly.  I worked around this by directly calling `vio_has_data`, but in order to do that I had to include all of `violite.h`, and I can't say I'm particularly happy about it. 

Anyway, we probably are simply backing off our using of multi-statement selects, if you wanted to merge this (or had better ideas about how to properly poll for results when the c client library had buffered stuff), great.

(note, PR targets 0.3 just because thats what we're on still).

### DIFF
--- a/ext/mysql2/client.c
+++ b/ext/mysql2/client.c
@@ -539,12 +539,31 @@ static VALUE disconnect_and_raise(VALUE self, VALUE error) {
   return Qnil;
 }
 
+static void wait_for_fd(int fd, struct timeval *tvp) {
+  int retval;
+
+  for(;;) {
+    retval = rb_wait_for_single_fd(fd, RB_WAITFD_IN, tvp);
+
+    if (retval == 0) {
+      rb_raise(cMysql2Error, "Timeout waiting for a response from the last query. (waited %d seconds)", FIX2INT(tvp->tv_sec));
+    }
+
+    if (retval < 0) {
+      rb_sys_fail(0);
+    }
+
+    if (retval > 0) {
+      break;
+    }
+  }
+}
+
 static VALUE do_query(void *args) {
   struct async_query_args *async_args;
   struct timeval tv;
   struct timeval* tvp;
   long int sec;
-  int retval;
   VALUE read_timeout;
 
   async_args = (struct async_query_args *)args;
@@ -565,21 +584,7 @@ static VALUE do_query(void *args) {
     tvp->tv_usec = 0;
   }
 
-  for(;;) {
-    retval = rb_wait_for_single_fd(async_args->fd, RB_WAITFD_IN, tvp);
-
-    if (retval == 0) {
-      rb_raise(cMysql2Error, "Timeout waiting for a response from the last query. (waited %d seconds)", FIX2INT(read_timeout));
-    }
-
-    if (retval < 0) {
-      rb_sys_fail(0);
-    }
-
-    if (retval > 0) {
-      break;
-    }
-  }
+  wait_for_fd(async_args->fd, tvp);
 
   return Qnil;
 }
@@ -1022,6 +1027,12 @@ static VALUE rb_mysql_client_more_results(VALUE self)
       return Qtrue;
 }
 
+static void *nogvl_next_result(void *ptr) {
+  mysql_client_wrapper *wrapper = ptr;
+
+  return (void *)INT2NUM(mysql_next_result(wrapper->client));
+}
+
 /* call-seq:
  *    client.next_result
  *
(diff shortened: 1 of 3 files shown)


sodabrew:
Wow! This'll take me a bit to read through, thanks for the thorough research and PR!

ext/mysql2/client.c
```diff
@@ -539,21 +540,35 @@ static VALUE disconnect_and_raise(VALUE self, VALUE error) {
...
+      rb_sys_fail(0);
+    }
+
+    if (retval > 0) {
+      break;
+    }
+  }
+}
+
+static struct timeval *get_read_timeout(VALUE self, struct timeval *tvp)
```
sodabrew: When this function returns it returns the pointer from the `tvp` argument?

ext/mysql2/client.c
```diff
@@ -563,23 +578,20 @@ static VALUE do_query(void *args) {
       rb_raise(cMysql2Error, "read_timeout must be a positive integer, you passed %ld", sec);
     }
     tvp->tv_usec = 0;
```
sodabrew: While you're in here, would you move the tv_usec = 0 to be directly below the tv_sec = sec line?

ext/mysql2/client.c
```diff
@@ -1031,8 +1049,24 @@ static VALUE rb_mysql_client_more_results(VALUE self)
...
-    ret = mysql_next_result(wrapper->client);
+
+    if (mysql_more_results(wrapper->client) == 0)
+      return Qfalse;
+
+    /* if both queries were ready by the time we finished the first query,
+     * the underlying mysql client library will have read both results into the buffer,
+     * defeating our attempt to poll() until ready.  detect that case by peeking at the "vio" buffer.
+     */
+    if ( !wrapper->client->net.vio->has_data(wrapper->client->net.vio) ) {
```
sodabrew: Is `has_data` the only method that required the `violite.h` headers?

ext/mysql2/client.c
```diff
@@ -539,21 +540,35 @@ static VALUE disconnect_and_raise(VALUE self, VALUE error) {
...
-static VALUE do_query(void *args) {
-  struct async_query_args *async_args;
-  struct timeval tv;
-  struct timeval* tvp;
-  long int sec;
+static void wait_for_fd(int fd, struct timeval *tvp) {
   int retval;
+
+  for(;;) {
+    retval = rb_wait_for_single_fd(fd, RB_WAITFD_IN, tvp);
```
sodabrew: I was surprised, so I double-checked that `rb_wait_for_single_fd` does allow a NULL timeval argument. Could you add a comment to let the future coders know that NULL can get through from `get_read_timeout` and is a legal value for this function?

https://github.com/ruby/ruby/search?utf8=%E2%9C%93&q=rb_wait_for_single_fd

ext/mysql2/violite.h
```diff
@@ -0,0 +1,280 @@
+/* Copyright (c) 2000, 2015, Oracle and/or its affiliates. All rights
```
sodabrew: Looking more closely, it's a little worse than just defining the one `has_data` method, we're actually becoming aware of the `st_vio` structure in order to call one of its member function pointers. I'm concerned that this isn't part of the "public API" and would be subject to change across MySQL versions.

osheroff: let's talk through this one before I address the rest of the stuff.  Yes, you're definitely right that this isn't part of the public api; I'm not particularly happy with the solution, but I looked for quite a bit and couldn't find any function, public or not, that would advise you about whether you had to block.  

Having said that, the history of this structure is that it hasn't had a breaking change since (maybe) 2104 or (probably) 2012: https://github.com/mysql/mysql-server/commits/5.7/include/violite.h

sodabrew:
A while back I thought about changing the multiple results API a bit to help support streaming on the 2nd - nth queries in a multiple statements, see #600 and let me know what you think of the idea proposed there.

ext/mysql2/violite.h
osheroff: the other possibility here is to declare dependency on the two functions that the `has_next` pointer could point to, `vio_buff_has_data` or `vio_ssl_has_data`, but this would mean figuring out whether the underlying connection has negotiated an SSL connection or not, and it's not clear to me if we can figure that out.

sodabrew: Ok, that's a great lead. Let's see if there's a way to determine the connection type. It might be nice to expose that information, too like `Mysql2::Client#ssl_connection?`

sodabrew:
I want to pick up work on this PR again soon, but I prefer to bring it into a next major (0.5.0) rather than making major changes to 0.3.x or even 0.4.x now. Would you be amenable to rebasing your branch to master?

> (note, PR targets 0.3 just because thats what we're on still)

sodabrew:
Closing in favor of rebase #962

PR #811, opened 2016-12-08
Handle being terminated while reading query result

darnaut:
The problem is that if execution is terminated while reading a result set, the next attempt to execute a query on the connection fails with "This connection is still waiting for a result, try again once you have the result". This error persists even if reconnect is enabled.

The issue stems from reading the result set using `rb_thread_call_without_gvl`. If the calling thread is terminated inside `rb_thread_call_without_gvl`, the function does not return and the caller doesn't get a chance to reset the active thread associated with the connection — which normally happens after the result set has been read.

The solution is to disconnect and mark the connection as inactive if the thread is terminated while reading a result set.

### DIFF
--- a/ext/mysql2/client.c
+++ b/ext/mysql2/client.c
@@ -647,26 +647,26 @@ static VALUE do_query(void *args) {
 
   return Qnil;
 }
-#else
-static VALUE finish_and_mark_inactive(void *args) {
-  VALUE self = args;
-  MYSQL_RES *result;
+#endif
 
+static VALUE disconnect_and_mark_inactive(VALUE self) {
   GET_CLIENT(self);
 
+  /* Check if execution terminated while result was still being read. */
   if (!NIL_P(wrapper->active_thread)) {
-    /* if we got here, the result hasn't been read off the wire yet
-       so lets do that and then throw it away because we have no way
-       of getting it back up to the caller from here */
-    result = (MYSQL_RES *)rb_thread_call_without_gvl(nogvl_store_result, wrapper, RUBY_UBF_IO, 0);
-    mysql_free_result(result);
-
+    /* Invalidate the MySQL socket to prevent further communication. */
+    if (invalidate_fd(wrapper->client->net.fd) == Qfalse) {
+      rb_warn("mysql2 failed to invalidate FD safely, closing unsafely\n");
+      close(wrapper->client->net.fd);
+    }
+    /* Skip mysql client check performed before command execution. */
+    wrapper->client->status = MYSQL_STATUS_READY;
     wrapper->active_thread = Qnil;
+    wrapper->connected = 0;
   }
 
   return Qnil;
 }
-#endif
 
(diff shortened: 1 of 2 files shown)


darnaut:
This is similar in nature to #592 but also works for earlier Ruby versions where `Thread.handle_interrupt` is not available.

/cc @sodabrew

ext/mysql2/client.c
```diff
@@ -647,26 +647,26 @@ static VALUE do_query(void *args) {
...
-    result = (MYSQL_RES *)rb_thread_call_without_gvl(nogvl_store_result, wrapper, RUBY_UBF_IO, 0);
-    mysql_free_result(result);
-
+    /* Invalidate the MySQL socket to prevent further communication. */
+    if (invalidate_fd(wrapper->client->net.fd) == Qfalse) {
+      rb_warn("mysql2 failed to invalidate FD safely, closing unsafely\n");
+      close(wrapper->client->net.fd);
+    }
+    /* Skip mysql client check performed before command execution. */
+    wrapper->client->status = MYSQL_STATUS_READY;
```
sodabrew: This looks like another reach into the MySQL structures, could you explain more about what this causes the libmysqlclient code to do?

sodabrew:
Sorry for the long delay before I reviewed this! Just left some comments and questions.

ext/mysql2/client.c
darnaut: It's already tested elsewhere and the assumption that the connection won't be broken is wrong (unless reconnect is enabled).

spec/mysql2/client_spec.rb
```diff
@@ -538,15 +538,6 @@ def run_gc
...
 
-      it 'should be impervious to connection-corrupting timeouts in #query' do
-        pending('`Thread.handle_interrupt` is not defined') unless Thread.respond_to?(:handle_interrupt)
-        # attempt to break the connection
-        expect { Timeout.timeout(0.1) { @client.query('SELECT SLEEP(0.2)') } }.to raise_error(Timeout::Error)
-
-        # expect the connection to not be broken
-        expect { @client.query('SELECT 1') }.to_not raise_error
-      end
-
```
darnaut: It's already tested elsewhere and the assumption that the connection won't be broken is wrong (unless reconnect is enabled).

ext/mysql2/client.c
darnaut: Yes, another reach unfortunately. But basically libmysql has a top-level check for `status != MYSQL_STATUS_READ` before executing new commands, which prevents the (later) internal reconnect logic to kick in. This is basically to avoid the check and allow reconnect to kick-in if the connection was interrupted while reading the result set. For what it's worth, even libmysql itself resets the flag in a few places to make things work.

epinault:
Any update on this PR? IT would be great to get this into the next version of the gem if possible

sodabrew:
Please try the new mysql2 0.4.6 release and confirm is the issue is resolved as expected.

Issue #813, opened 2016-12-28
Segfault on ubuntu with mysql 5.7.16 and ruby 2.3.3


duduribeiro:
Hi guys.

I'm currently having a problem with mysql2 (and I don't know if this problem is in mysql2 gem).
I'm using sidekiq 4.2.5 with mysql2 0.4.5 (I upgraded from 0.4.4 to test if this resolve the problem, but without success) and ruby 2.3.3 (upgraded from ruby 2.3.1 to test if this resolve the problem, but without success).

And my jobs works fine, until I receive this Segfault, and it kills my Sidekiq.
The stacktrace is here: http://pastebin.com/3pZUeeFG

I tried to found some issue with this behave but I haven't found anything.

Is this a problem with mysql2?

Thanks a lot 🍻

sodabrew:
The relevant part of the stack trace, good news it had line numbers! I'll dig further in the new year:

```

-- C level backtrace information -------------------------------------------
/home/deploy/.rvm/rubies/ruby-2.3.3/bin/../lib/libruby.so.2.3(rb_vm_bugreport+0x4ea) [0x7f68eddc98da] vm_dump.c:692
/home/deploy/.rvm/rubies/ruby-2.3.3/bin/../lib/libruby.so.2.3(rb_bug_context+0xd4) [0x7f68edc5a2b4] error.c:435
/home/deploy/.rvm/rubies/ruby-2.3.3/bin/../lib/libruby.so.2.3(sigsegv+0x3e) [0x7f68edd37e0e] signal.c:890
/lib/x86_64-linux-gnu/libc.so.6 [0x7f68ed8484b0]
/home/deploy/threshold-dashboard/releases/97/vendor/bundle/ruby/2.3.0/gems/mysql2-0.4.5/lib/mysql2/mysql2.so(disconnect_and_raise+0x3d) [0x7f68e720423d] client.c:594
/home/deploy/.rvm/rubies/ruby-2.3.3/bin/../lib/libruby.so.2.3(rb_rescue2+0x1f8) [0x7f68edc631e8] eval.c:846
/home/deploy/threshold-dashboard/releases/97/vendor/bundle/ruby/2.3.0/gems/mysql2-0.4.5/lib/mysql2/mysql2.so(rb_query+0x14b) [0x7f68e72063eb] client.c:761
```

sodabrew:
Resolved by #816

duduribeiro:
[...]

PR #816, opened 2017-01-02
Segfault on SIGINT (race condition)

kpumuk:
We have faced a pretty weird edge case recently. Multiple threads, slow MySQL query, and Unicorn killing the worker due to timeout lead to segfault in Mysql2 gem (`disconnect_and_raise` method).

Short version: when there are multiple threads and a signal, it is possible that `disconnect_and_raise` might be called after `close`, leading to a segfault.

Reproducible on Linux x86_64 with the following code:

```ruby
clients = []
10.times do |idx|
  Thread.new do
    # Simulate SQL query
    conn = Mysql2::Client.new
    clients << conn
    conn.query('SELECT SLEEP(120)')
  end
end

Thread.new do
  # Emulates Unicorn sending SIGINT signal when timeout is reached.
  # This code will be executed with a delay of 1 second
  # so the SIGINT will arrive during sleep in the main thread.
  sleep 1
  Process.kill('INT', $$)
end
sleep 5
# Segfault here
clients.each(&:close)
```

### DIFF
--- a/ext/mysql2/client.c
+++ b/ext/mysql2/client.c
@@ -596,9 +596,11 @@ static VALUE disconnect_and_raise(VALUE self, VALUE error) {
   /* Invalidate the MySQL socket to prevent further communication.
    * The GC will come along later and call mysql_close to free it.
    */
-  if (invalidate_fd(wrapper->client->net.fd) == Qfalse) {
-    fprintf(stderr, "[WARN] mysql2 failed to invalidate FD safely, closing unsafely\n");
-    close(wrapper->client->net.fd);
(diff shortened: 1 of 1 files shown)


sodabrew:
This appears to be a the same issue and a solution to #813, thank you!

abedra:
@sodabrew do you know when 0.4.6 will be cut?

sodabrew:
I'll probably do 0.4.6 in January, there's another bug and open PR in the milestone that might wrap soon.

sodabrew:
Well January ended up busy. But 0.4.6 is now released!

PR #817, opened 2017-01-03
Delete existing mysql data at installation MySQL 5.7.

takeshinoda:
Delete Travis default MySQL data directories. MySQL 5.7 test passes.


sodabrew:
Thank you!

PR #819, opened 2017-01-04
Only MySQL 5.5 tests run with precise(12.04).

takeshinoda:
MySQL 5.5 test passes.


sodabrew:
Thanks for your help getting the tests back to passing!

PR #820, opened 2017-01-05
Change default CI Ruby version to 2.4

yui-knk:


.travis.yml
```diff
@@ -49,25 +49,25 @@ matrix:
...
           - mysql2gem.example.com
-    - rvm: 2.0.0
+    - rvm: 2.4
       env: DB=mariadb10.1
       addons:
         mariadb: 10.1
         hosts:
           - mysql2gem.example.com
-    - rvm: 2.0.0
+    - rvm: 2.4
```
sodabrew: Ruby 2.4 isn't available on the `precise` image.

sodabrew:
A couple of the builds failed randomly, I restarted them. The precise image does't have Ruby 2.4, I think we need to hold that one back to 2.0.0. The OS X build images might be out of date for Ruby 2.4, but I expect they will catch up soon.

.travis.yml
yui-knk: I found out we should write "2.4.0"....

see: https://github.com/brianmario/mysql2/pull/821

PR #831, opened 2017-03-06
Better error messages when the connection options have errors

helgeblod:
Why:

To avoid cryptic errors and to give better error messages back to
  the user. Say the user uses this connection string:

```
DATABASE_URL="mysql2://<database>:<password>@<host>:<port>/<db>?encoding=utf8&pool=1&connect_timeout=2read_timeout=5&write_timeout=5"
```

With the current code, the `connect_timeout=2` and `read_timeout` will
be ignored (as the connection string is missing a `&`). A better
approach would be to evaluate the input as an integer and output an
error message if it isn't one. For more info see the discussion in the
PR below.

This change addresses the need by:

* https://github.com/brianmario/mysql2/pull/791


helgeblod:
Not really sure why Travis is complaining btw. Looks unrelated to my change, and all other PRs looks like they fail as well.

Issue #838, opened 2017-04-11
bundle fails after installing MySQL 8.0.1 DMR


yahonda:
[The MySQL 8.0.1 Milestone Release is available](http://mysqlserverteam.com/the-mysql-8-0-1-milestone-release-is-available/) . I was trying to use it with mysql2 gem. It fails. It used to work witth MySQL 8.0.0 DMR in my environment.

#### Environment:
```
$ more /etc/redhat-release
Fedora release 25 (Twenty Five)
```

* Downloaded https://dev.mysql.com/get/Downloads/MySQL-8.0/mysql-8.0.1-0.1.dmr.fc25.x86_64.rpm-bundle.tar and install MySQL 8.0.1 DMR
```
$ rpm -qa |grep mysql-community
mysql-community-devel-8.0.1-0.1.dmr.fc25.x86_64
mysql-community-client-8.0.1-0.1.dmr.fc25.x86_64
mysql-community-server-8.0.1-0.1.dmr.fc25.x86_64
mysql-community-common-8.0.1-0.1.dmr.fc25.x86_64
mysql-community-test-8.0.1-0.1.dmr.fc25.x86_64
mysql-community-libs-8.0.1-0.1.dmr.fc25.x86_64
$
```

* Ruby and bundler version
```
$ ruby -v
ruby 2.4.1p111 (2017-03-22 revision 58053) [x86_64-linux]
$ bundle -v
Bundler version 1.14.6
```

#### Steps to reproduce:
```
$ git clone https://github.com/rails/rails.git
$ cd rails/activerecord
$ bundle
```
#### Results:
```
$ bundle
Fetching gem metadata from https://rubygems.org/.........
Fetching version metadata from https://rubygems.org/..
Fetching dependency metadata from https://rubygems.org/.
Using rake 12.0.0
Using concurrent-ruby 1.0.4
Using i18n 0.8.1
Using minitest 5.3.3
Using thread_safe 0.3.5
Using builder 3.2.3
Using erubi 1.6.0
Using mini_portile2 2.1.0
Using rack 2.0.1
Using nio4r 2.0.0
Using websocket-extensions 0.1.2
Using mime-types-data 3.2016.0521
Using arel 8.0.0 from https://github.com/rails/arel.git (at master@437aa3a)
Using public_suffix 2.0.5
Using amq-protocol 2.1.0
Using ast 2.3.0
Using beaneater 1.0.0
Using dante 0.2.0
Using bcrypt 3.1.11
[...]

PR #840, opened 2017-04-14
Use `bool` instead of `my_bool` which has been removed since MySQL 8.0.1

yahonda:
This pull request addresses #838

Refer "Changes in MySQL 8.0.1 (2017-04-10, Development Milestone)"
https://dev.mysql.com/doc/relnotes/mysql/8.0/en/news-8-0-1.html

> Incompatible Change: The mysql.h header file now requires a C++ or C99 compiler to compile.
> The my_bool type is no longer used in MySQL source code.
> Any third-party code that used this type to represent C boolean variables
> should use the bool or int C type instead. (Bug #25597667)

### DIFF
--- a/ext/mysql2/result.h
+++ b/ext/mysql2/result.h
@@ -1,5 +1,6 @@
 #ifndef MYSQL2_RESULT_H
 #define MYSQL2_RESULT_H
+#include <stdbool.h>
 
 void init_mysql2_result(void);
 VALUE rb_mysql_result_to_obj(VALUE client, VALUE encoding, VALUE options, MYSQL_RES *r, VALUE statement);
@@ -21,8 +22,8 @@ typedef struct {
   mysql_client_wrapper *client_wrapper;
   /* statement result bind buffers */
   MYSQL_BIND *result_buffers;
-  my_bool *is_null;
-  my_bool *error;
+  bool *is_null;
+  bool *error;
   unsigned long *length;
 } mysql2_result_wrapper;
 
--- a/ext/mysql2/client.c
+++ b/ext/mysql2/client.c
@@ -111,7 +111,7 @@ static VALUE rb_set_ssl_mode_option(VALUE self, VALUE setting) {
   int val = NUM2INT( setting );
   if (version >= 50703 && version < 50711) {
     if (val == SSL_MODE_DISABLED || val == SSL_MODE_REQUIRED) {
-      my_bool b = ( val == SSL_MODE_REQUIRED );
+      bool b = ( val == SSL_MODE_REQUIRED );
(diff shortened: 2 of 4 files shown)


ext/mysql2/result.h
```diff
@@ -1,5 +1,6 @@
 #ifndef MYSQL2_RESULT_H
 #define MYSQL2_RESULT_H
+#include <stdbool.h>
```
sodabrew: Are there platforms these days that don't have stdbool.h available?

I'll check my Solaris and Windows environments. Off the cuff I think this will be a safe change.

yahonda: According to [Bug #85131	Remove my_bool](https://bugs.mysql.com/bug.php?id=85131)

> - mysql.h now requires a C++ or C99 header to compile. (All supported platforms have C99 compilers, but some are C89 by default and need mode switches like --std=gnu99 or --std=c99.)

Then as long as it is one of the "supported platforms", C99 should be supported.

I'm not familiar with Windows or Solaris C Compiler but googled and found these links:

* [What's New for Visual C++ in Visual Studio 2013
](https://msdn.microsoft.com/en-us/library/hh409293(v=vs.120).aspx)

>> Supports these ISO C99 language features:
>> _Bool


* [Sun Studio 12: C User's Guide
](http://docs.oracle.com/cd/E19205-01/819-5265/bjate/index.html)

>> B.2.73 -xc99[=o]
>> The -xc99 option controls compiler recognition of the implemented features from the C99 
>> standard (ISO/IEC 9899:1999, Programming Language - C).

sodabrew: I remember reading at one point that Ruby wants extensions to compile with C89? Is that still true?

yahonda: Sorry I did not find the information if Ruby requires C extensions compiled with C89.

yahonda:
@tenderlove 

Hi, This is the pull request we were talking at RailsConf this morning. Would you mind taking a look at this pull request and have some comments if mysql2 gem can use C99 boolean feature?

Thanks.

ext/mysql2/result.h
sodabrew: I found it! While it looks dated at this point, the assumptions are stated here: https://bugs.ruby-lang.org/projects/ruby-trunk/wiki/Assumptions

sodabrew: I also found that `result.c` is already using `uintptr_t` which is a C99 type from `stdint.h` but we've not included `stdint.h` ourselves, which is probably an error.

yahonda:
Thanks!

sodabrew:
It may be the case that we need to use `my_bool` on older MySQL versions to avoid problems with endianness expansion missing the relevant bits. @junaruga here's where we switched from `my_bool` to `bool` in the first place.

PR #845, opened 2017-05-15
Adding support for MYSQL_ENABLE_CLEARTEXT_PLUGIN flag

denisenkom:
This will allow usage of AWS IAM user login for MySQL database

### DIFF
--- a/ext/mysql2/client.c
+++ b/ext/mysql2/client.c
@@ -405,11 +405,10 @@ static VALUE rb_mysql_get_ssl_cipher(VALUE self)
   return rb_str;
 }
 
-static VALUE rb_connect(VALUE self, VALUE user, VALUE pass, VALUE host, VALUE port, VALUE database, VALUE socket, VALUE flags, VALUE enable_cleartext_plugin) {
+static VALUE rb_connect(VALUE self, VALUE user, VALUE pass, VALUE host, VALUE port, VALUE database, VALUE socket, VALUE flags) {
   struct nogvl_connect_args args;
   time_t start_time, end_time, elapsed_time, connect_timeout;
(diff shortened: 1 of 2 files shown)


denisenkom:
Will take a look

sodabrew:
+1 for @jeremy's suggestion, that approach will be consistent with the other connection hash optoins.

@denisenkom take a look at the block of code at https://github.com/denisenkom/mysql2/blob/master/lib/mysql2/client.rb#L34-L44 – The pattern there is each of the hash keys from the connections hash maps onto a setter method, and each setter method is a wrapper around `_mysql_client_options` with the value conversion needed for the Ruby VALUE to the C argument, for example:

```
$ grep -r _mysql_client_options ext/
ext/mysql2/client.c
...
1204:  return _mysql_client_options(self, MYSQL_OPT_RECONNECT, value);
1208:  return _mysql_client_options(self, MYSQL_OPT_LOCAL_INFILE, value);
1218:  return _mysql_client_options(self, MYSQL_OPT_CONNECT_TIMEOUT, value);
...
```

denisenkom:
Agree, updated PR with new code, please take a look.

sodabrew:
Looks good, thanks for the quick turnaround from feedback! Will merge if the tests all run green; is there a reasonable way to unit-test this with the MySQL installed on Travis CI?

I'll need to figure out the timing for when I get a release out, whether it's 0.4.7 or 0.5.0 next to include this.

denisenkom:
I can't think of a way to test this in Travis CI instance.  I did test it with our RDS MySQL instance with IAM authentication plugin enabled.

marutosi:
I have created #866.

PR #846, opened 2017-05-23
Fix use after free of client field when closing with reconnect option

dylanahsmith:
## Problem

The following script would cause a crash

```ruby
require 'mysql2'
client = Mysql2::Client.new(username: 'root', reconnect: true)
client.close
client.query "SELECT 1"
```

because Mysql2::Client#close frees the `wrapper->client`, then several methods were missing any checks to make sure `wrapper->client != NULL`.

When the `reconnect: true` option isn't used, then Mysql2::Client#close will set `wrapper->connected = 0`, so the `REQUIRE_CONNECTED` macro would raise, but the `reconnect: true` option skips this check.

## Solution

~~Add a `REQUIRE_CLIENT` macro which raises if `wrapper->client == NULL`, and is used in `REQUIRE_CONNECTED` as well as in methods that didn't use the REQUIRE_CONNECTED macro.~~

Delay freeing the `wrapper->client` until the wrapper is being freed during garbage collection.  Instead, just have the Mysql2::Client#close method call mysql_close, mark the connection not connected and disable the reconnect option.  That way the `REQUIRE_CONNECTED` macro will raise after the close method is called and methods that don't use the REQUIRE_CONNECTED macro won't access freed memory.

### DIFF
--- a/ext/mysql2/client.c
+++ b/ext/mysql2/client.c
@@ -32,14 +32,21 @@ VALUE rb_hash_dup(VALUE other) {
 
 #define CONNECTED(wrapper) (wrapper->client && wrapper->client->net.vio != NULL && wrapper->client->net.fd != -1)
 
+#define REQUIRE_CLIENT(wrapper) \
+  if (!wrapper->client) { \
+    rb_raise(cMysql2Error, "MySQL client is closed"); \
+  }
+
 #define REQUIRE_CONNECTED(wrapper) \
   REQUIRE_INITIALIZED(wrapper) \
+  REQUIRE_CLIENT(wrapper) \
   if (!CONNECTED(wrapper) && !wrapper->reconnect_enabled) { \
     rb_raise(cMysql2Error, "MySQL client is not connected"); \
   }
 
 #define REQUIRE_NOT_CONNECTED(wrapper) \
   REQUIRE_INITIALIZED(wrapper) \
+  REQUIRE_CLIENT(wrapper) \
   if (CONNECTED(wrapper)) { \
     rb_raise(cMysql2Error, "MySQL connection is already open"); \
   }
@@ -288,7 +295,7 @@ void decr_mysql2_client(mysql_client_wrapper *wrapper)
 
   if (wrapper->refcount == 0) {
 #ifndef _WIN32
-    if (CONNECTED(wrapper) && !wrapper->automatic_close) {
+    if (wrapper->client && CONNECTED(wrapper) && !wrapper->automatic_close) {
       /* The client is being garbage collected while connected. Prevent
        * mysql_close() from sending a mysql-QUIT or from calling shutdown() on
        * the socket by invalidating it. invalidate_fd() will drop this
@@ -362,6 +369,7 @@ static VALUE rb_mysql_client_warning_count(VALUE self) {
   unsigned int warning_count;
   GET_CLIENT(self);
 
+  REQUIRE_CLIENT(wrapper);
   warning_count = mysql_warning_count(wrapper->client);
 
   return UINT2NUM(warning_count);
@@ -372,6 +380,7 @@ static VALUE rb_mysql_info(VALUE self) {
   VALUE rb_str;
   GET_CLIENT(self);
 
+  REQUIRE_CLIENT(wrapper);
   info = mysql_info(wrapper->client);
 
   if (info == NULL) {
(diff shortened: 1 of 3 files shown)


dylanahsmith:
The only CI failure is a flaky test failure that was also seen in https://github.com/brianmario/mysql2/pull/848, so isn't related to this PR.

dylanahsmith:
Right, the alternative would be to use `mysql_close(wrapper->client);` without actually freeing the client, which would be simpler.  Should I do that instead?  Or keep the PR as is?

jeremy:
Not sure about that. We changed to freeing in close to work around a memory leak: https://github.com/brianmario/mysql2/commit/569aa3d996a794073b0704771e45f78c7dba2182

We could fix that leak by not nulling out `wrapper->client`, too, though.

sodabrew:
Thank you! I will review this tomorrow!

dylanahsmith:
I pushed a change that delays freeing the `wrapper->client` until just before the `wrapper` is freed.  The `wrapper->client` isn't set to NULL, so there isn't a problem with memory leaks.  We don't need to set `wrapper->client` to NULL, since `mysql_close` sets the pointers it frees to NULL, so it is idempotent.

I have kept the initial solution in the first commit in case you would prefer to just not use the MYSQL struct after it has been closed.

ext/mysql2/client.c
```diff
@@ -30,14 +30,21 @@ VALUE rb_hash_dup(VALUE other) {
     rb_raise(cMysql2Error, "MySQL client is not initialized"); \
   }
 
+#define REQUIRE_CLIENT(wrapper) \
```
sodabrew: Are there cases where a `GET_CLIENT` would _not_ be followed by a `REQUIRE_CLIENT` ? If not, then this check for a valid client pointer could go into `GET_CLIENT`.

ext/mysql2/client.c
```diff
@@ -271,13 +264,10 @@ static VALUE invalidate_fd(int clientfd)
...
   mysql_client_wrapper *wrapper = ptr;
 
-  if (wrapper->client) {
-    mysql_close(wrapper->client);
-    xfree(wrapper->client);
-    wrapper->client = NULL;
-    wrapper->connected = 0;
-    wrapper->active_thread = Qnil;
-  }
+  mysql_close(wrapper->client);
```
sodabrew: I need to look back to where this defensive programming was added, but we definitely had cases of invalid pointer dereferencing in the past that this check was intended to prevent.

spec/mysql2/client_spec.rb
```diff
@@ -288,6 +288,14 @@ def run_gc
     }.to raise_error(Mysql2::Error)
   end
 
+  it "should not try to query closed mysql connection" do
```
sodabrew: This test is correct. If the current behavior is not respecting that the connection has been closed, we may consider adding an extra flag or additional state/value for the `wrapper->connected` field.

sodabrew:
I re-ran the failed test, it was spurious.

sodabrew:
I merged #847 which removes the `connected` field, so let's take that route and run with it.

ext/mysql2/client.c
dylanahsmith: The naming of `GET_CLIENT` can be a bit confusing, since it actually gets the client wrapper.  `GET_CLIENT_WRAPPER` might be a more appropriate name.

Here are some functions that need the client wrapper without the `REQUIRE_CLIENT` check:
* **initialize_ext**: to set the client field
* **rb_mysql_client_close**: to be idempotent, not raising on the second close
* **rb_mysql_client_ping**: to return false when closed
* **rb_mysql_client_encoding**: value is in the wrapper
* **get_automatic_close**: value is in the wrapper
* **set_automatic_close**: value is in the wrapper

sodabrew:
Oof, also conflicts with merge of #796

sodabrew:
I am generally nervous about removing the defensive checks against dereferencing NULL pointers, even though they don't help about dereferencing invalid / freed pointers.

dylanahsmith:
I found https://github.com/brianmario/mysql2/commit/e7924df06ace8937b3d4f42728f92d1bf4032990 where it looks like you might have decided that it isn't safe to call `mysql_close` on an already closed connection, despite the comment `/* It's safe to call mysql_close() on an already closed connection. */` not being removed in that commit.  So I pushed a commit to add a similar `closed` flag to avoid whatever problems we had with this in the past.

ext/mysql2/client.h
```diff
@@ -46,7 +46,7 @@ typedef struct {
   int automatic_close;
   int initialized;
   int refcount;
-  int freed;
```
dylanahsmith: This was unused, that is why I removed it here.

PR #848, opened 2017-05-24
Close several unclosed clients in tests

dylanahsmith:
## Problem

I noticed some flaky test failures that seem like they were related to unclosed connections, like this test failure in https://travis-ci.org/brianmario/mysql2/jobs/235381348

```
  1) Mysql2::Client should terminate connections when calling close
     Failure/Error:
       expect {
         Mysql2::Client.new(DatabaseCredentials['root']).close
       }.to_not change {
         @client.query("SHOW STATUS LIKE 'Aborted_%'").to_a +
           @client.query("SHOW STATUS LIKE 'Threads_connected'").to_a
       }
     
       expected result not to have changed, but did change from [{"Variable_name"=>"Aborted_clients", "Value"=>"0"}, {"Variable_name"=>"Aborted_connects", "Value"=>"0"}, {"Variable_name"=>"Threads_connected", "Value"=>"6"}] to [{"Variable_name"=>"Aborted_clients", "Value"=>"0"}, {"Variable_name"=>"Aborted_connects", "Value"=>"0"}, {"Variable_name"=>"Threads_connected", "Value"=>"5"}]
     # ./spec/mysql2/client_spec.rb:174:in `block (2 levels) in <top (required)>'
```

Notice that the test failed because an extra connection was closed.  Also, there were 6 Threads_connected at the start of the test, when there should only be one for `@client`.

## Solution

I went through the tests looking for unclosed clients and added a `close` call for the ones I found.

### DIFF
--- a/spec/mysql2/client_spec.rb
+++ b/spec/mysql2/client_spec.rb
@@ -6,6 +6,7 @@ RSpec.describe Mysql2::Client do
     let(:cnf_file) { File.expand_path('../../my.cnf', __FILE__) }
 
     it "should not raise an exception for valid defaults group" do
+      @client.close
       expect {
         opts = DatabaseCredentials['root'].merge(:default_file => cnf_file, :default_group => "test")
         @client = Mysql2::Client.new(opts)
@@ -13,6 +14,7 @@ RSpec.describe Mysql2::Client do
     end
 
     it "should not raise an exception without default group" do
+      @client.close
       expect {
         @client = Mysql2::Client.new(DatabaseCredentials['root'].merge(:default_file => cnf_file))
       }.not_to raise_error
@@ -91,6 +93,7 @@ RSpec.describe Mysql2::Client do
     client = Mysql2::Client.new(options)
     result = client.query("SELECT @something;")
     expect(result.first['@something']).to eq('setting_value')
+    client.close
   end
 
   it "should send init_command after reconnect" do
@@ -121,6 +124,7 @@ RSpec.describe Mysql2::Client do
     # At last, check that the init command executed
     result = client.query("SELECT @something;")
     expect(result.first['@something']).to eq('setting_value')
+    client.close
   end
 
   it "should have a global default_query_options hash" do
@@ -212,30 +216,36 @@ RSpec.describe Mysql2::Client do
     it "is enabled by default" do
       client = Mysql2::Client.new(DatabaseCredentials['root'])
       expect(client.automatic_close?).to be(true)
+      client.close
     end
 
     if RUBY_PLATFORM =~ /mingw|mswin/
       it "cannot be disabled" do
+        client = nil
         expect do
           client = Mysql2::Client.new(DatabaseCredentials['root'].merge(:automatic_close => false))
           expect(client.automatic_close?).to be(true)
         end.to output(/always closed by garbage collector/).to_stderr
+        client.close
 
         expect do
           client = Mysql2::Client.new(DatabaseCredentials['root'].merge(:automatic_close => true))
           expect(client.automatic_close?).to be(true)
         end.to_not output(/always closed by garbage collector/).to_stderr
+        client.close
 
         expect do
           client = Mysql2::Client.new(DatabaseCredentials['root'].merge(:automatic_close => true))
           client.automatic_close = false
           expect(client.automatic_close?).to be(true)
         end.to output(/always closed by garbage collector/).to_stderr
+        client.close
       end
     else
       it "can be configured" do
         client = Mysql2::Client.new(DatabaseCredentials['root'].merge(:automatic_close => false))
         expect(client.automatic_close?).to be(false)
+        client.close
       end
 
       it "can be assigned" do
@@ -251,6 +261,7 @@ RSpec.describe Mysql2::Client do
 
         client.automatic_close = 9
         expect(client.automatic_close?).to be(true)
+        client.close
       end
 
       it "should not close connections when running in a child process" do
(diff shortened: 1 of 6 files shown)


dylanahsmith:
That same test I mentioned in the PR description is failing, but this time with Threads_connected increasing by 1 after opening a connection and closing it.  Seems like this test is flaky on master.

```
  1) Mysql2::Client should terminate connections when calling close
     Failure/Error:
       expect {
         Mysql2::Client.new(DatabaseCredentials['root']).close
       }.to_not change {
         @client.query("SHOW STATUS LIKE 'Aborted_%'").to_a +
           @client.query("SHOW STATUS LIKE 'Threads_connected'").to_a
       }
     
       expected result not to have changed, but did change from [{"Variable_name"=>"Aborted_clients", "Value"=>"0"}, {"Variable_name"=>"Aborted_connects", "Value"=>"0"}, {"Variable_name"=>"Threads_connected", "Value"=>"1"}] to [{"Variable_name"=>"Aborted_clients", "Value"=>"0"}, {"Variable_name"=>"Aborted_connects", "Value"=>"0"}, {"Variable_name"=>"Threads_connected", "Value"=>"2"}]
     # ./spec/mysql2/client_spec.rb:178:in `block (2 levels) in <top (required)>'
```

It looks like the mysql server might not have synchronously closed the connection when receiving a the QUIT message.  Otherwise, there must be something else connecting to the database in the middle of the test.

Either way, the flaky test failure isn't caused by this PR.

dylanahsmith:
I can reproduce the flaky test failure locally using the following changes to make it happen more reliably

```diff
diff --git a/spec/mysql2/client_spec.rb b/spec/mysql2/client_spec.rb
index dab4036..3390b60 100644
--- a/spec/mysql2/client_spec.rb
+++ b/spec/mysql2/client_spec.rb
@@ -171,12 +171,13 @@ RSpec.describe Mysql2::Client do
   end

   it "should terminate connections when calling close" do
-    expect {
-      Mysql2::Client.new(DatabaseCredentials['root']).close
-    }.to_not change {
-      @client.query("SHOW STATUS LIKE 'Aborted_%'").to_a +
+    10_000.times do |i|
+      expect {
+        Mysql2::Client.new(DatabaseCredentials['root']).close
+      }.to_not change {
         @client.query("SHOW STATUS LIKE 'Threads_connected'").to_a
-    }
+      }
+    end
   end

   it "should not leave dangling connections after garbage collection" do
```

dylanahsmith:
`mysql_close` is implemented using `simple_command(mysql,COM_QUIT,(uchar*) 0,0,1)` where that last argument is `skip_check` which `cli_advanced_command` checks with

```c
  if (!skip_check)
    result= ((mysql->packet_length=cli_safe_read(mysql)) == packet_error ?
	     1 : 0);
```

so it doesn't wait for the response to the quit command, which is what makes this race condition feasible.  I'll open another PR to address this, since this PR is dealing with a different reason why that test can fail.

spec/mysql2/client_spec.rb
```diff
@@ -6,13 +6,15 @@
...
     it "should not raise an exception for valid defaults group" do
+      @client.close
       expect {
         opts = DatabaseCredentials['root'].merge(:default_file => cnf_file, :default_group => "test")
         @client = Mysql2::Client.new(opts)
       }.not_to raise_error
     end
 
     it "should not raise an exception without default group" do
+      @client.close
```
jeremy: Perhaps clearer to have these tests simply not set `@client`. Let them be responsible for creating and closing this separate client instance.

spec/mysql2/statement_spec.rb
```diff
@@ -3,6 +3,7 @@
 
 RSpec.describe Mysql2::Statement do
   before :each do
+    @client.close
```
jeremy: Feels odd to close the client before instantiating it. Should this be in an `after { … }`?

spec/mysql2/error_spec.rb
```diff
@@ -15,6 +15,8 @@
     error
   end
 
+  after(:each) { client.close }
+
```
jeremy: (Note that `after { … }` is equivalent to `after(:each) { … }`)

spec/mysql2/statement_spec.rb
dylanahsmith: [spec/spec_helper.rb](https://github.com/brianmario/mysql2/blob/master/spec/spec_helper.rb#L27) already created a client.  This one creates another one with a change in its configuration.

spec/mysql2/error_spec.rb
dylanahsmith: I didn't know that.  All the other test files use `after(:each)`, so should I keep it as `after(:each)` for consistency?

spec/mysql2/statement_spec.rb
jeremy: Right, just feels odd to start a test by cleaning up after setup.

dylanahsmith:
Introduced an abstraction as you suggested for creating clients, which avoid the problems with spec files wanting to override `@client` with one with a different configuration.  I decided to call it `new_client` since it is a wrapper around `Mysql2::Client.new`

sodabrew:
Well this is awesome!

sodabrew:
The test at line 265 is much worse now - could you take another look to be sure that you're exercising that a parent processes's connection is not garbage collected and closed by a child process that inherits the connection?

```
     
         # ./spec/mysql2/client_spec.rb:265:in `block (4 levels) in <top (required)>'
         # ./spec/mysql2/client_spec.rb:265:in `block (3 levels) in <top (required)>'
     # ./spec/mysql2/client_spec.rb:265:in `block (3 levels) in <top (required)>'
```

sodabrew:
Please rebase following merge of #846 and #853. Thanks!

PR #849, opened 2017-05-24
Fix flaky test that used Process.wait without specifying a pid

dylanahsmith:
## Problem

In CI build (https://travis-ci.org/brianmario/mysql2/jobs/235381342) for another PR I got this flaky test failure

```
     
         # ./spec/mysql2/client_spec.rb:264:in `block (4 levels) in <top (required)>'
         # ./spec/mysql2/client_spec.rb:264:in `block (3 levels) in <top (required)>'
     # ./spec/mysql2/client_spec.rb:264:in `block (3 levels) in <top (required)>'
```

The problem was that [Process.wait](http://ruby-doc.org/core-2.3.3/Process.html#method-c-wait) was being called without specifying a pid, which defaults to pid=-1, "Waits for any child process (the default if no pid is given)."  So it was actually waiting for the empty fork that was added as a hack for some version of rubinius

## Solution

Passed pid for the fork we care about into Process.wait.

I also removed the empty fork now, since CI no longer runs with rubinius.

### DIFF
--- a/spec/mysql2/client_spec.rb
+++ b/spec/mysql2/client_spec.rb
@@ -247,17 +247,13 @@ RSpec.describe Mysql2::Client do
         client = Mysql2::Client.new(DatabaseCredentials['root'])
         client.automatic_close = false
 
-        # this empty `fork` call fixes this tests on RBX; without it, the next
-        # `fork` call hangs forever. WTF?
-        fork {}
-
-        fork do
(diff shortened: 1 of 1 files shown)


sodabrew:
The empty fork was needed for Rubinius 2.5 which is out of date at this point anyhow.

PR #850, opened 2017-05-24
Wait for close to be processed by server in test for Threads_connected

dylanahsmith:
## Problem

As discussed in https://github.com/brianmario/mysql2/pull/848#issuecomment-303704563, the test I am changing fails in a flaky way due to `mysql_close` sending a quit command without waiting for a response, so the following query for the Threads_connected can return a count that includes the connection that was been closed on the client side.

## Solution

Give the server a second to detect and response to the client closing the connection by using the `SHOW PROCESSLIST` command to check if the server still thinks the client is connected.

I wrapped the contents of the test in a loop as I commented in https://github.com/brianmario/mysql2/pull/848#issuecomment-303746961 to confirm that this deals with this type of flaky test failure.

### DIFF
--- a/spec/mysql2/client_spec.rb
+++ b/spec/mysql2/client_spec.rb
@@ -172,7 +172,16 @@ RSpec.describe Mysql2::Client do
 
   it "should terminate connections when calling close" do
     expect {
-      Mysql2::Client.new(DatabaseCredentials['root']).close
+      client = Mysql2::Client.new(DatabaseCredentials['root'])
+      connection_id = client.thread_id
+      client.close
+
+      # mysql_close sends a quit command without waiting for a response
+      # so give the server some time to handle the detect the closed connection
+      10.times do
+        break unless @client.query("SHOW PROCESSLIST").detect { |row| row['Id'] == connection_id }
+        sleep(0.1)
+      end
     }.to_not change {
       @client.query("SHOW STATUS LIKE 'Aborted_%'").to_a +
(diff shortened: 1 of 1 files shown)


spec/mysql2/client_spec.rb
```diff
@@ -172,7 +172,16 @@ def run_gc
...
+
+      # mysql_close sends a quit command without waiting for a response
+      # so give the server some time to handle the detect the closed connection
+      10.times do
+        break unless @client.query("SHOW PROCESSLIST").detect { |row| row['Id'] == connection_id }
+        sleep(0.1)
+      end
     }.to_not change {
       @client.query("SHOW STATUS LIKE 'Aborted_%'").to_a +
         @client.query("SHOW STATUS LIKE 'Threads_connected'").to_a
```
jeremy: `Threads_connected` *should* change, huh?

Perhaps we should just check that the client connection wasn't aborted. The `SHOW PROCESSLIST` check is already demonstrating that the client is no longer connected, which is the aim of the spec.

dylanahsmith: > `Threads_connected` *should* change, huh?

It should *not* change as the `}.to_not change {` line specifies, since the client was opened and closed in the expect block.

Sure, I could just check that it hasn't aborted here, which should hopefully also make this test likely to fail due to some connection from another test being closed by the garbage collector during the `expect` block.

dylanahsmith: Pushed this change

Issue #851, opened 2017-05-25
installation error in MariaDB 10.2.6 


kim612:
Hi.
macOS Homebrew started to distribute MariaDB 10.2.6 recently.
But when I installed mysql2 for this version, I found some errors.
MariaDB 10.1.X is still ok, so I changed it and it has no problem.

$ gem install mysql2 -- --with-mysql-config=/usr/local/bin/mysql_config
Building native extensions with: '--with-mysql-config=/usr/local/bin/mysql_config'
This could take a while...
ERROR:  Error installing mysql2:
	ERROR: Failed to build gem native extension.

    current directory: /usr/local/lib/ruby/gems/2.4.0/gems/mysql2-0.4.6/ext/mysql2
/usr/local/opt/ruby/bin/ruby -r ./siteconf20170525-2285-kb0wzp.rb extconf.rb --with-mysql-config=/usr/local/bin/mysql_config
checking for rb_absint_size()... yes
checking for rb_absint_singlebit_p()... yes
checking for ruby/thread.h... yes
checking for rb_thread_call_without_gvl() in ruby/thread.h... yes
checking for rb_thread_blocking_region()... no
checking for rb_wait_for_single_fd()... yes
checking for rb_hash_dup()... yes
checking for rb_intern3()... yes
checking for rb_big_cmp()... yes
-----
Using mysql_config at /usr/local/bin/mysql_config
-----
checking for mysql.h... yes
checking for SSL_MODE_DISABLED in mysql.h... no
checking for MYSQL_OPT_SSL_ENFORCE in mysql.h... yes
checking for errmsg.h... yes
checking for mysqld_error.h... yes
-----
Setting rpath to /usr/local/Cellar/mariadb/10.2.6/lib
-----
creating Makefile

current directory: /usr/local/lib/ruby/gems/2.4.0/gems/mysql2-0.4.6/ext/mysql2
make "DESTDIR=" clean

current directory: /usr/local/lib/ruby/gems/2.4.0/gems/mysql2-0.4.6/ext/mysql2
make "DESTDIR="
compiling client.c
client.c:138:1: warning: control may reach end of non-void function [-Wreturn-type]
}
^
client.c:923:32: error: use of undeclared identifier 'MYSQL_SERVER_VERSION'
  header_version = rb_str_new2(MYSQL_LINK_VERSION);
                               ^
client.c:53:30: note: expanded from macro 'MYSQL_LINK_VERSION'
[...]

PR #857, opened 2017-06-21
Minimal required changes to make it work for mariadb 10.2

damm:
Updated extconf.rb to check if we have the header
Update mysql2_ext.h to load it if the def is set

Signed-off-by: Scott M. Likens <scott@likens.us>

Refs #851

### DIFF
--- a/ext/mysql2/extconf.rb
+++ b/ext/mysql2/extconf.rb
@@ -25,7 +25,6 @@ end
 # MariaDB 10.2+ Needs mysql/mysql_version.h loaded to get the constants we need
 have_header('mysql/mysql_version.h')
 
-
 # 2.1+
 have_func('rb_absint_size')
 have_func('rb_absint_singlebit_p')
--- a/ext/mysql2/extconf.rb
+++ b/ext/mysql2/extconf.rb
@@ -22,9 +22,6 @@ def add_ssl_defines(header)
   $CFLAGS << ' -DNO_SSL_MODE_SUPPORT' if has_no_support
 end
 
-# MariaDB 10.2+ Needs mysql/mysql_version.h loaded to get the constants we need
-have_header('mysql/mysql_version.h')
-
 # 2.1+
 have_func('rb_absint_size')
 have_func('rb_absint_singlebit_p')
--- a/ext/mysql2/extconf.rb
+++ b/ext/mysql2/extconf.rb
@@ -22,6 +22,10 @@ def add_ssl_defines(header)
   $CFLAGS << ' -DNO_SSL_MODE_SUPPORT' if has_no_support
 end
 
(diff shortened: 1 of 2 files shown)


ext/mysql2/extconf.rb
```diff
@@ -22,6 +22,10 @@ def add_ssl_defines(header)
   $CFLAGS << ' -DNO_SSL_MODE_SUPPORT' if has_no_support
 end
 
+# MariaDB 10.2+ Needs mysql/mysql_version.h loaded to get the constants we need
+have_header('mysql/mysql_version.h')
+
+
```
sodabrew: Looks good - nit: please move to line 114, and use the `prefix` variable from the path test at line 100.

ext/mysql2/mysql2_ext.h
```diff
@@ -20,6 +20,9 @@ void Init_mysql2(void);
 #include <mysql/errmsg.h>
 #include <mysql/mysqld_error.h>
 #endif
+#ifdef HAVE_MYSQL_MYSQL_VERSION_H
+#include <mysql/mysql_version.h>
+#endif
```
sodabrew: At the risk of a too many combinations, should this also be conditional on the mysql/ prefix as above?

damm: I'm not against change but please rephrase because I think I am confused.

the HAVE_MYSQL_MYSQL_VERSION_H is coming from extconf.rb from 

```ruby
have_header('mysql/mysql_version.h')
```

Thanks

damm: Is this configuration file available always in other conditions? it looks like yes

ext/mysql2/extconf.rb
damm: I removed this as if we are just using the original ifdef we don't need to check against this?

sodabrew:
I also checked that mysql_version.h is always available in the MySQL Connector/C and MariaDB Connector/C distributions.

AdamWill:
Here's some background on the issue: https://jira.mariadb.org/browse/MDEV-13370

I think upstream's intent is that you shouldn't have to do explicit includes like this (and note that having that include will prevent the project from building against mariadb-connector-c). If I'm reading what they're telling me right, I *think* they must either add `MYSQL_SERVER_VERSION` (and other things from `mysql_version.h`) to `mariadb_version.h`, or document that we're not supposed to use it any more.

I haven't yet been able to find any clear documentation of what interfaces / definitions / etc. are *intended* to be lost in this rather poorly-explained switch to including `mariadb-connector-c` as `libmariadb` in the main mariadb codebase, and what things have only been lost unintentionally as part of the change :/

PR #859, opened 2017-06-22
Fix a grammer

naoki-k:
Fixed a simple grammer. Please review this 🙏


sodabrew:
Thank you, but I think this is better left as-is. There's also a matching "method should exist" on the Statement spec. Have a good day!

PR #864, opened 2017-06-27
Fix vio to pvio in client.c based on undefined error

snood1205:
In installing mysql2 from github, I am getting a build error in c where .vio on client.c line 33 is undefined and replacing it with .pvio fixes this error.

### DIFF
--- a/ext/mysql2/client.c
+++ b/ext/mysql2/client.c
@@ -30,7 +30,7 @@ VALUE rb_hash_dup(VALUE other) {
     rb_raise(cMysql2Error, "MySQL client is not initialized"); \
   }
 
-#define CONNECTED(wrapper) (wrapper->client->net.vio != NULL && wrapper->client->net.fd != -1)
+#define CONNECTED(wrapper) (wrapper->client->net.pvio != NULL && wrapper->client->net.fd != -1)
 
 #define REQUIRE_CONNECTED(wrapper) \
   REQUIRE_INITIALIZED(wrapper) \


sodabrew:
Does this work for MariaDB 10.2? It fails for all other versions of MySQL and MariaDB. It would need to be made conditional with a test in extconf.rb.

bigcurl:
Yes, this buils with MariaDB 10.2.6 but results in` NameError: uninitialized constant Mysql2::Client::LONG_PASSWORD`

sodabrew:
In all versions of MySQL and MariaDB < 10.2, this code change fails:

```
../../../../ext/mysql2/client.c:33:49: error: ‘NET’ has no member named ‘pvio’
 #define CONNECTED(wrapper) (wrapper->client->net.pvio != NULL && wrapper->client->net.fd != -1)             
```

If MariaDB 10.2 has changed the name of this struct member, it will be necessary to detect the change with a test in extconf.rb, and provide both alternative definitions of the macro.

sodabrew:
In this build:
https://travis-ci.org/sodabrew/mysql2/jobs/248639728

I fixed the Travis CI installation of MariaDB 10.2, and then tested for the mysql.net.vio / mysql.net.pvio structures, and found that there is no such thing as mysql.net.pvio. So... could you clarify what client library you're using that requires this change?

```
checking for MYSQL.net.vio in mysql.h... yes
checking for MYSQL.net.pvio in mysql.h... no
```

sodabrew:
Aha, found it: MariaDB-Connector-C 3.0 made the change:
https://github.com/MariaDB/mariadb-connector-c/commit/7500f371501d72b5eb98b7263a807e487440cf60

I don't have a test suite for the Connector-C 3.0 libraries, but they're generally known to work. I will keep this code test in place.

PR #875, opened 2017-07-31
Fix compat with RubyInstaller-2.4 on Windows

### LINKED ISSUE #861: cannot load such file -- mysql2/2.4/mysql2 (LoadError)`
I am new on Ruby on Rails. My Ruby version is ruby2.4.1p111 (2017-03-22 revision 58053) [x64-mingw32].

I have created a new project using rails new sample_cms -d mysql command.

The command rails server is throwing following error.

C:/Ruby24-x64/lib/ruby/gems/2.4.0/gems/mysql2-0.4.6-x64-mingw32/lib/mysql2/mysql2.rb:2:inrequire': cannot load such file -- mysql2/2.4/mysql2 (LoadError)`
I have installed mysql and mysql Connector C. I have installed the mysql2 gem like that.

gem install mysql2 —platform=ruby — ‘—with-mysql-dir="C:\Program Files\MySQL\MySQL Connector C 6.1"’
I have uninstall and install the mysql2 gem but still facing the issue. Here is the directory which may have issue.

27505154-7b8f3f4e-58b2-11e7-8c8a-66980553443c

Here is the gem file entry for mysql2

# Bundle edge Rails instead: gem 'rails', github: 'rails/rails'
gem 'rails', '~> 5.1.1'
# Use mysql as the database for Active Record
gem 'mysql2', '>= 0.3.18', '< 0.5'
# Use Puma as the app server
gem 'puma', '~> 3.7'
Any Help?

larskanis:
Since RubyInstaller-2.4+ is bundled with MSYS2 and the libmariadbclient can be installed per gemspec library dependency, it is easy to build the mysql2 gem in Windows.

The MSYS2/MINGW dependency feature is documented here: https://github.com/oneclick/rubyinstaller2/wiki/For-gem-developers#msys2-library-dependency

This also adds ruby-2.4 binaries (per rake-compiler-dock-0.6.0), so that the mysql2 is still useabel as a binary gem.

Fixes #861

The change in the spec is required for mariadbclient. It throws an error if no query was executed.

Due to the stdcall convention on i686, the mysql_query() function check fails, so that it is omitted, now.

### DIFF
--- a/lib/mysql2.rb
+++ b/lib/mysql2.rb
@@ -13,16 +13,20 @@ if RUBY_PLATFORM =~ /mswin|mingw/
     ENV['RUBY_MYSQL2_LIBMYSQL_DLL']
   elsif File.exist?(File.expand_path('../vendor/libmysql.dll', File.dirname(__FILE__)))
     # Use vendor/libmysql.dll if it exists, convert slashes for Win32 LoadLibrary
-    File.expand_path('../vendor/libmysql.dll', File.dirname(__FILE__)).tr('/', '\\')
+    File.expand_path('../vendor/libmysql.dll', File.dirname(__FILE__))
+  elsif defined?(RubyInstaller)
+    # RubyInstaller-2.4+ native build doesn't need DLL preloading
   else
     # This will use default / system library paths
     'libmysql.dll'
   end
 
-  require 'Win32API'
-  LoadLibrary = Win32API.new('Kernel32', 'LoadLibrary', ['P'], 'I')
-  if 0 == LoadLibrary.call(dll_path)
-    abort "Failed to load libmysql.dll from #{dll_path}"
+  if dll_path
+    require 'Win32API'
+    LoadLibrary = Win32API.new('Kernel32', 'LoadLibrary', ['P'], 'I')
+    if 0 == LoadLibrary.call(dll_path)
+      abort "Failed to load libmysql.dll from #{dll_path}"
+    end
   end
 end
 
--- a/spec/mysql2/statement_spec.rb
+++ b/spec/mysql2/statement_spec.rb
@@ -712,7 +712,6 @@ RSpec.describe Mysql2::Statement do
 
     it 'should return number of rows affected by an insert' do
       stmt = @client.prepare 'INSERT INTO lastIdTest (blah) VALUES (?)'
-      expect(stmt.affected_rows).to eq 0
       stmt.execute 1
       expect(stmt.affected_rows).to eq 1
     end
--- a/tasks/compile.rake
+++ b/tasks/compile.rake
@@ -9,7 +9,7 @@ Rake::ExtensionTask.new("mysql2", Mysql2::GEMSPEC) do |ext|
   # clean compiled extension
   CLEAN.include "#{ext.lib_dir}/*.#{RbConfig::CONFIG['DLEXT']}"
 
-  if RUBY_PLATFORM =~ /mswin|mingw/
+  if RUBY_PLATFORM =~ /mswin|mingw/ && !defined?(RubyInstaller)
     # Expand the path because the build dir is 3-4 levels deep in tmp/platform/version/
     connector_dir = File.expand_path("../../vendor/#{vendor_mysql_dir}", __FILE__)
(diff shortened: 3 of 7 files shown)


lib/mysql2.rb
```diff
@@ -13,16 +13,20 @@
     ENV['RUBY_MYSQL2_LIBMYSQL_DLL']
   elsif File.exist?(File.expand_path('../vendor/libmysql.dll', File.dirname(__FILE__)))
     # Use vendor/libmysql.dll if it exists, convert slashes for Win32 LoadLibrary
-    File.expand_path('../vendor/libmysql.dll', File.dirname(__FILE__)).tr('/', '\\')
+    File.expand_path('../vendor/libmysql.dll', File.dirname(__FILE__))
```
sodabrew: I remember the backslashes being required, since the LoadLibrary call is native Windows and not Ruby.

lib/mysql2.rb
```diff
@@ -13,16 +13,20 @@
...
+  elsif defined?(RubyInstaller)
```
sodabrew: What is `RUBY_PLATFORM` when `RubyInstaller` is defined?

spec/mysql2/statement_spec.rb
```diff
@@ -712,7 +712,6 @@ def stmt_count
 
     it 'should return number of rows affected by an insert' do
       stmt = @client.prepare 'INSERT INTO lastIdTest (blah) VALUES (?)'
-      expect(stmt.affected_rows).to eq 0
```
sodabrew: Why was this removed?

tasks/compile.rake
```diff
@@ -26,6 +26,10 @@ Rake::ExtensionTask.new("mysql2", Mysql2::GEMSPEC) do |ext|
       Rake::Task['lib/mysql2/mysql2.rb'].invoke
       # vendor/libmysql.dll is invoked from extconf.rb
       Rake::Task['vendor/README'].invoke
+
+      # only the source gem has a package dependency - the binary gem ships it's own DLL version
```
sodabrew: nit: `its` for the possessive :)

tasks/compile.rake
```diff
@@ -77,7 +81,7 @@ task :devkit do
 end
 
 if RUBY_PLATFORM =~ /mingw|mswin/
-  Rake::Task['compile'].prerequisites.unshift 'vendor:mysql'
+  Rake::Task['compile'].prerequisites.unshift 'vendor:mysql' unless defined?(RubyInstaller)
```
sodabrew: I'm not following why the unless RubyInstaller case is needed. This file shouldn't execute in the source code case, only when I'm building the binaries at my desktop.

lib/mysql2.rb
larskanis: LoadLibrary accepts both, forward and backward slashes as all filesystem related WINAPI functions do. There's no need to translate path separators. cmd and powershell don't recognize forward slashs, but can handle them as well.

lib/mysql2.rb
larskanis: `x64-mingw32` and `i386-mingw32` - the same as previous RubyInstaller versions.

spec/mysql2/statement_spec.rb
larskanis: As commented: The change in the spec is required for mariadbclient. It throws an error if no query was executed.

tasks/compile.rake
larskanis: Sorry, but thanks for the hint!

tasks/compile.rake
larskanis: `rake spec` depends on `compile`, which is OK, but it shouldn't depend on the vendored mysql library on RubyInstaller-2.4.

The reason is: the simplest way to build the mysql2 gem is by using the `libmariadbclient` pacman package. It is automatically installed though the `metadata['msys2_mingw_dependencies']` tag when running `gem install mysql2`

On the other hand the vendored library works as well, provided that the `unzip` command is installed. This can be done per `pacman -S unzip`. But I think that this shouldn't the default way on RubyInstaller-2.4.

larskanis:
Is there anything I can do to get this merged and released?

Issue #878, opened 2017-08-18
Compilation failures against MariaDB Connector/C 3.0.2


ljfranklin:
I'm getting the following error when compiling the gem against the [MariaDB C Connector 3.0.2](https://downloads.mariadb.org/connector-c/3.0.2/):

```
creating Makefile

To see why this extension failed to compile, please check the mkmf.log which can
be found here:

/var/vcap/data/packages/cloud_controller_ng/c35814b300ca9c86b4f84011949d2eadbf08707e/cloud_controller_ng/vendor/bundle/ruby/2.3.0/extensions/x86_64-linux/2.3.0-static/mysql2-0.4.8/mkmf.log

current directory:
/var/vcap/data/packages/cloud_controller_ng/c35814b300ca9c86b4f84011949d2eadbf08707e/cloud_controller_ng/vendor/bundle/ruby/2.3.0/gems/mysql2-0.4.8/ext/mysql2
make "DESTDIR=" clean

current directory:
/var/vcap/data/packages/cloud_controller_ng/c35814b300ca9c86b4f84011949d2eadbf08707e/cloud_controller_ng/vendor/bundle/ruby/2.3.0/gems/mysql2-0.4.8/ext/mysql2
make "DESTDIR="
compiling statement.c
In file included from statement.c:1:0:
./mysql2_ext.h:14:23: fatal error: mysql_com.h: No such file or directory
 #include <mysql_com.h>
                       ^
compilation terminated.
make: *** [statement.o] Error 1

make failed, exit code 2

Gem files will remain installed in
/var/vcap/data/packages/cloud_controller_ng/c35814b300ca9c86b4f84011949d2eadbf08707e/cloud_controller_ng/vendor/bundle/ruby/2.3.0/gems/mysql2-0.4.8
for inspection.
Results logged to
/var/vcap/data/packages/cloud_controller_ng/c35814b300ca9c86b4f84011949d2eadbf08707e/cloud_controller_ng/vendor/bundle/ruby/2.3.0/extensions/x86_64-linux/2.3.0-static/mysql2-0.4.8/gem_make.out
```

[...]

PR #880, opened 2017-08-20
patch for issue #878

actsasflinn:
Newer version of mariadb seem to have moved or renamed a few header files.

### DIFF
--- a/ext/mysql2/mysql2_ext.h
+++ b/ext/mysql2/mysql2_ext.h
@@ -9,7 +9,13 @@ void Init_mysql2(void);
 #define RSTRING_NOT_MODIFIED
 #include <ruby.h>
 
-#ifdef HAVE_MYSQL_H
+#ifdef HAVE_SERVER_MYSQL_H
+#include <mysql.h>
+#include <server/mysql_com.h>
+#include <errmsg.h>
+#include <mysqld_error.h>
+#include <server/mysql_version.h>
+#elif HAVE_MYSQL_H
 #include <mysql.h>
 #include <mysql_com.h>
 #include <errmsg.h>
--- a/ext/mysql2/extconf.rb
+++ b/ext/mysql2/extconf.rb
@@ -97,7 +97,9 @@ else
   rpath_dir = usr_local_lib
 end
 
-if have_header('mysql.h')
+if have_header('server/mysql.h')
+  prefix = nil
+elsif have_header('mysql.h')
   prefix = nil
 elsif have_header('mysql/mysql.h')
   prefix = 'mysql'


ext/mysql2/mysql2_ext.h
```diff
@@ -9,7 +9,13 @@ void Init_mysql2(void);
 #define RSTRING_NOT_MODIFIED
 #include <ruby.h>
 
-#ifdef HAVE_MYSQL_H
+#ifdef HAVE_SERVER_MYSQL_H
+#include <mysql.h>
+#include <server/mysql_com.h>
+#include <errmsg.h>
+#include <mysqld_error.h>
+#include <server/mysql_version.h>
```
sodabrew: I wonder why these headers are spread out across two paths like this. What did they change?

jaredbeck: Looks like https://github.com/MariaDB/server/commit/3ec96c18244b448b2adf30deb856e34907c12326#diff-46193e2e675adcde81f1d095d8b3e1db is the culprit?

> MDEV-13370 Ambiguous behaviour regarding installation of header files
> install all server includes under /usr/include/mysql/server/ and C/C includes under /usr/include/mysql/

jaredbeck:
I have tested this PR with the following:

```
mysql --version
mysql  Ver 15.1 Distrib 10.2.8-MariaDB, for osx10.12 (x86_64) using readline 5.1

gem uninstall mysql2
# Uninstall all versions

# Edit my Gemfile
gem "mysql2",
  git: "https://github.com/actsasflinn/mysql2.git",
  ref: "76ee6acb09077503a579f4ba5aae9b4137144957"

bundle
Using mysql2 0.4.9 from https://github.com/actsasflinn/mysql2.git (at 76ee6ac@76ee6ac)
```

Apparently when you can use `git:` to download a gem it still compiles native extensions.

```
file /Users/jared/.rbenv/versions/2.3.4/lib/ruby/gems/2.3.0/bundler/gems/mysql2-76ee6acb0907/ext/mysql2/mysql2.bundle 
/Users/jared/.rbenv/versions/2.3.4/lib/ruby/gems/2.3.0/bundler/gems/mysql2-76ee6acb0907/ext/mysql2/mysql2.bundle: Mach-O 64-bit bundle x86_64
```

So, I think this is a reasonable test of this PR. I guess I'd recommend merging.

sodabrew:
Thanks for the link to the upstream bug! What I took from it is that if we make this suggested change, it will allow mysql2 to draw from the MariaDB _server_ package headers. I actually want to be sure that the mysql2 gem can compile and run with only the MariaDB Connector/C _client_ package installed.

jaredbeck:
> Thanks for the link to the upstream bug! 

You're welcome! It definitely took some research (and some `git log`) to track it down.

> What I took from it is that if we make this suggested change, it will allow mysql2 to draw from the MariaDB server package headers. I actually want to be sure that the mysql2 gem can compile and run with only the MariaDB Connector/C client package installed.

Oh, good point! But, please be aware that in homebrew the `mariadb` and `mariadb-connector-c` packages are marked as "conflicting", ie. mutually exclusive. We must install `mariadb` for local development, so we cannot install the `mariadb-connector-c` package.

sodabrew:
Oh, that's interesting that the packages conflict currently. If it's the case that this PR allows mysql2 to operate with the MariaDB-server-client package, and it already works with the MariaDB-connector-c package, then it's sort of a no-brainer to just merge this as-is and let the have_header checks sort things out.

Having this new Connector/C on the scene really makes the testing matrix complicated... :|

jaredbeck:
> If it's the case that this PR allows mysql2 to operate with the MariaDB-server-client package, and it already works with the MariaDB-connector-c package, then it's sort of a no-brainer to just merge this as-is and let the have_header checks sort things out.

Fine by me.

It's no reason to hold up the PR, but just FYI I'm going to try to upgrade our staging server to mariadb 10.2.8 so I can test this PR against the ubuntu packages. I'm not sure how valuable that test is, because we install both the `mariadb-server` and the `libmariadb-dev` packages on our servers, but .. can't hurt.

jaredbeck:
> .. I'm going to try to upgrade our staging server to mariadb 10.2.8 so I can test this PR against the ubuntu packages. I'm not sure how valuable that test is, because we install both the mariadb-server and the libmariadb-dev packages on our servers, but .. can't hurt.

I've performed this test and am now running mariadb 10.2.8 on my staging server and using the following in my Gemfile:

```
gem "mysql2",
  git: "https://github.com/actsasflinn/mysql2.git",
  ref: "76ee6acb09077503a579f4ba5aae9b4137144957"
```

@sodabrew anything else we need to do before merging this? Just a reminder this is a priority because homebrew users are completely unable to use the `mysql` gem right now.

jaredbeck:
@sodabrew two week ping. Anything else you need from us?

grknight:
I think this is the wrong approach.

The real fix is to not include mysql_com.h (renamed to mariadb_com.h) nor mysql_version.h (renamed to mariadb_version.h) when they don't exist as the `#include <mysql.h>` will pull in their mariadb counterparts.

actsasflinn:
@grknight I agree, the gem should be less coupled to file locations. There's some additional changes to make in order to handle differences in what is defined in mysql_version.h and mariadb_version.h - they do not appear to be equal. Using server/mysql_version.h retains compatibility.

At a bare minimum referencing LIBMYSQL_VERSION doesn't seem to be supported by mariadb_version.h
See: https://github.com/brianmario/mysql2/blob/master/ext/mysql2/client.c#L56-L60

In the short term does it make sense to accept the patch and then circle back on refactoring to a better long term solution?

actsasflinn:
I've got a second fix implementation addressing @grknight concerns... not sure if it holistically addresses the coupling issues.

https://github.com/actsasflinn/mysql2/commit/f60600dae11d3cf629c1b895a4051e5572c13978

sodabrew:
Thank you for the PR! I worked this out a bit differently in #900, please take a look to make sure it works for you!

PR #888, opened 2017-09-09
fix for issue #878

actsasflinn:
Alternative fix that addresses @grknight concern on coupling to the file locations.

### DIFF
--- a/ext/mysql2/client.c
+++ b/ext/mysql2/client.c
@@ -53,8 +53,10 @@ VALUE rb_hash_dup(VALUE other) {
  * variable to use, but MYSQL_SERVER_VERSION gives the correct numbers when
  * linking against the server itself
  */
-#ifdef LIBMYSQL_VERSION
-  #define MYSQL_LINK_VERSION LIBMYSQL_VERSION
+#ifdef MARIADB_CLIENT_VERSION_STR
+  #define MYSQL_LINK_VERSION MARIADB_CLIENT_VERSION_STR
+#elif LIBMYSQL_VERSION
+  #define MYSQL_LINK_VERSION MYSQL_VERSION_ID
 #else
   #define MYSQL_LINK_VERSION MYSQL_SERVER_VERSION
 #endif
--- a/ext/mysql2/mysql2_ext.h
+++ b/ext/mysql2/mysql2_ext.h
@@ -11,10 +11,8 @@ void Init_mysql2(void);
 
 #ifdef HAVE_MYSQL_H
 #include <mysql.h>
-#include <mysql_com.h>
 #include <errmsg.h>
 #include <mysqld_error.h>
-#include <mysql_version.h>
 #else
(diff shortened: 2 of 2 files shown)


ext/mysql2/client.c
```diff
@@ -53,8 +53,10 @@ VALUE rb_hash_dup(VALUE other) {
  * variable to use, but MYSQL_SERVER_VERSION gives the correct numbers when
  * linking against the server itself
  */
-#ifdef LIBMYSQL_VERSION
-  #define MYSQL_LINK_VERSION LIBMYSQL_VERSION
+#ifdef MARIADB_CLIENT_VERSION_STR
+  #define MYSQL_LINK_VERSION MARIADB_CLIENT_VERSION_STR
```
sodabrew: This is a suggested fix I found as well, works for me.

ext/mysql2/client.c
```diff
@@ -53,8 +53,10 @@ VALUE rb_hash_dup(VALUE other) {
...
+#elif LIBMYSQL_VERSION
+  #define MYSQL_LINK_VERSION MYSQL_VERSION_ID
```
sodabrew: MYSQL_VERSION_ID is numeric and will fail the major.minor dot tests later in this file.

ext/mysql2/mysql2_ext.h
```diff
@@ -11,10 +11,8 @@ void Init_mysql2(void);
 
 #ifdef HAVE_MYSQL_H
 #include <mysql.h>
-#include <mysql_com.h>
 #include <errmsg.h>
 #include <mysqld_error.h>
-#include <mysql_version.h>
```
sodabrew: Was added in #857 to get MariaDB 10.2 working :)

Also remove the second path below.

ext/mysql2/mysql2_ext.h
```diff
@@ -11,10 +11,8 @@ void Init_mysql2(void);
 
 #ifdef HAVE_MYSQL_H
 #include <mysql.h>
-#include <mysql_com.h>
```
sodabrew: `mysql_com.h` has been included since the very beginning, and has probably always been superfluous it looks - but I am loathe to remove it at a micro release.

sodabrew:
I spelunked some old MySQL mailing lists, and the problem here is MYSQL_VERSION_ID is purely numeric, and will fail the tests that were written for testing against major.minor (but not .micro) in the version string.

The documentation for these constants suggests they should always be available: https://dev.mysql.com/doc/refman/5.7/en/c-api-server-client-versions.html

A better approach that parsing the major.minor might be dividing the MYSQL_VERSION_ID by 100, but again I would queue up this change for the next major release of the gem.

sodabrew:
Thank you for the PR! I worked this out a bit differently in #900, please take a look to make sure it works for you!

Issue #879, opened 2017-08-18
Gem does not verify hostname when `ssl_mode: :verify_identity`


ljfranklin:
From the MySQL documentation, the `VERIFY_IDENTITY` SSL mode should verify the hostname in the server's certificate matches the hostname to which to client tried to connect. However, when setting `ssl_mode: :verify_identity` the gem connects even if there is a hostname mismatch. If I set `ssl_mode: :verify_identity` AND `sslverify: true`, the gem returns an SSL validation error as expected.

Using mysql2 version 0.4.8 and MariaDB connector C version 2.3.0 as the driver.

sodabrew:
`ssl_mode: :verify_identify` sets a different flag than the connection option SSL_VERIFY_SERVER_CERT.

Also:
>VERIFY_IDENTITY: Like VERIFY_CA, but additionally check the server's Common Name value in the certificate that the server sends to the client. The client verifies that name against the host name the client uses for connecting to the server, and the connection fails if there is a mismatch. For encrypted connections, this option helps prevent man-in-the-middle attacks. This is like the legacy --ssl-verify-server-cert option.

Also:
> The --ssl-verify-server-cert option is deprecated as of MySQL 5.7.11 and is removed in MySQL 8.0. Use --ssl-mode=VERIFY_IDENTITY instead.

So I don't understand why you're seeing a different behavior between `ssl_mode: :verify_identity` and `sslverify: true`, the documentation indicates that `ssl_mode: :verify_identity` should have taken care of this.

OH - I see the problem! Same as #889. In the connection step we need to call  to call `ssl_set()` even if all the arguments are NULL so that SSL is enabled at all.

sodabrew:
I have this suspicion that it is related to #889 but I'm not quite following my own hunch. This doesn't make sense to me in your OP, it must be a bug in MySQL client!?

>  If I set ssl_mode: :verify_identity AND sslverify: true, the gem returns an SSL validation error as expected.

sodabrew:
Please check of 0.4.10 resolves this issue for you.

sodabrew:
Did version 0.4.10 resolve the issue for you?

ljfranklin:
@sodabrew unfortunately I've since rolled off the team that was hitting this issue. Luckily (or unluckily), my coworker @pivotal-jamil-shamy is hitting a [similar issue](https://github.com/brianmario/mysql2/issues/936). Maybe they can check if version 0.4.10 works.

[...]

PR #889, opened 2017-09-20
Make sure ssl is enabled if only :sslverify is set

rhafer:
Previously, when "sslverify: false/true" was the only ssl related
options passed to the constructor, the module skipped the call to
"mysql_ssl_set". It seems however that for some variants for the mysql
client libraries calling "mysql_ssl_set" is the only way to enable SSL
for the client connections. (E.g. the libraries shipped as part of
mariadb 10.1 still lack support for MYSQL_OPT_SSL_ENFORCE and
MYSQL_OPT_SSL_MODE)

This change allows enabling ssl with default values for all other
options by just passing "sslverify: true" or "sslverify: false" to the
constructor. (Depending on whether server certificate verification is
wanted or not)

### DIFF
--- a/lib/mysql2/client.rb
+++ b/lib/mysql2/client.rb
@@ -47,7 +47,7 @@ module Mysql2
       self.charset_name = opts[:encoding] || 'utf8'
 
       ssl_options = opts.values_at(:sslkey, :sslcert, :sslca, :sslcapath, :sslcipher)
-      ssl_set(*ssl_options) if ssl_options.any?
+      ssl_set(*ssl_options) if ssl_options.any? || opts.key?(:sslverify)
       self.ssl_mode = parse_ssl_mode(opts[:ssl_mode]) if opts[:ssl_mode]
 
       case opts[:flags]
@@ -62,7 +62,7 @@ module Mysql2
       end
 
       # SSL verify is a connection flag rather than a mysql_ssl_set option
-      flags |= SSL_VERIFY_SERVER_CERT if opts[:sslverify] && ssl_options.any?
+      flags |= SSL_VERIFY_SERVER_CERT if opts[:sslverify]
 
       if [:user, :pass, :hostname, :dbname, :db, :sock].any? { |k| @query_options.key?(k) }
         warn "============= WARNING FROM mysql2 ============="


sodabrew:
Based on my reading of https://dev.mysql.com/doc/refman/5.7/en/mysql-ssl-set.html if all of the arguments to `mysql_ssl_set` are NULL, then it is a no-op. I haven't confirmed if the actual client code matches the documentation on this point.

Based on https://dev.mysql.com/doc/refman/5.7/en/c-api-encrypted-connections.html the way to enforce SSL is to call `mysql_options(... MYSQL_OPT_SSL_MODE ... SSL_MODE_REQUIRED)`.

rhafer:
> Based on my reading of https://dev.mysql.com/doc/refman/5.7/en/mysql-ssl-set.html if all of the arguments to mysql_ssl_set are NULL, then it is a no-op. I haven't confirmed if the actual client code matches the documentation on this point.

This doesn't seem to be true for mariadb 10.1:
https://github.com/MariaDB/server/blob/10.1/sql-common/client.c#L1683

Even if it's called with all arguments NULL it still sets  `mysql->options.use_ssl= TRUE;`

> Based on https://dev.mysql.com/doc/refman/5.7/en/c-api-encrypted-connections.html the way to enforce SSL is to call mysql_options(... MYSQL_OPT_SSL_MODE ... SSL_MODE_REQUIRED).

Unfortunately mariadb 10.1 (which AFAIK is still officially maintained) doesn't have those options.

sodabrew:
Makes sense now! Thanks for linking to the relevant line of code in the driver.

rhafer:
@sodabrew Hi. Is there any chance to get this merged?

lib/mysql2/client.rb
```diff
@@ -47,7 +47,7 @@ def initialize(opts = {})
       self.charset_name = opts[:encoding] || 'utf8'
 
       ssl_options = opts.values_at(:sslkey, :sslcert, :sslca, :sslcapath, :sslcipher)
-      ssl_set(*ssl_options) if ssl_options.any?
+      ssl_set(*ssl_options) if ssl_options.any? || opts.key?(:sslverify)
```
sodabrew: This will turn on ssl even if the only flag is `sslverify: false` which doesn't quite make sense to me.

sodabrew: I think given #879 that `ssl_set` should be called if any value of `:ssl_mode` is set, too.

Issue #891, opened 2017-09-21
MySQL 8.0.3 Release Candidate removes MYSQL_SECURE_AUTH


yahonda:
MySQL 8.0.3 Release Candidate has been released which removes MYSQL_SECURE_AUTH.

https://dev.mysql.com/doc/relnotes/mysql/8.0/en/news-8-0-3.html#mysqld-8-0-3-capi

> The deprecated secure_auth system variable and --secure-auth client option have been removed. The MYSQL_SECURE_AUTH option for the mysql_options() C API function was removed.

According to MySQL 5.7 document `MYSQL_SECURE_AUTH` has been enabled by default in 5.7
https://dev.mysql.com/doc/refman/5.7/en/mysql-options.html

> MYSQL_SECURE_AUTH (argument type: my_bool *)
> Whether to connect to a server that does not support the password hashing used in MySQL 4.1.1 and later. This option is enabled by default.

### Steps to reproduce

* Installed MySQL 8.0.3 Release Candidate
```ruby
$ rpm -qa |grep -i mysql-community
mysql-community-common-8.0.3-0.1.rc.fc26.x86_64
mysql-community-libs-compat-8.0.3-0.1.rc.fc26.x86_64
mysql-community-client-8.0.3-0.1.rc.fc26.x86_64
mysql-community-server-8.0.3-0.1.rc.fc26.x86_64
mysql-community-devel-8.0.3-0.1.rc.fc26.x86_64
mysql-community-libs-8.0.3-0.1.rc.fc26.x86_64
$
```

* Remove `mysql` gem from Gemfile 

* bundle exec rake spec

```ruby
$ bundle exec rake spec
cd tmp/x86_64-linux/mysql2/2.4.2
/usr/bin/gmake
compiling ../../../../ext/mysql2/client.c
../../../../ext/mysql2/client.c: In function ‘_mysql_client_options’:
../../../../ext/mysql2/client.c:886:10: error: ‘MYSQL_SECURE_AUTH’ undeclared (first use in this function); did you mean ‘MYSQL_DEFAULT_AUTH’?
     case MYSQL_SECURE_AUTH:
          ^~~~~~~~~~~~~~~~~
          MYSQL_DEFAULT_AUTH
../../../../ext/mysql2/client.c:886:10: note: each undeclared identifier is reported only once for each function it appears in
../../../../ext/mysql2/client.c: In function ‘set_secure_auth’:
[...]

PR #892, opened 2017-09-22
MYSQL_SECURE_AUTH has been removed in MySQL 8.0.3 RC

yahonda:
https://dev.mysql.com/doc/relnotes/mysql/8.0/en/news-8-0-3.html#mysqld-8-0-3-capi
> The deprecated secure_auth system variable and --secure-auth client option have been removed.
> The MYSQL_SECURE_AUTH option for the mysql_options() C API function was removed.

This pull request addresses #891

### DIFF
--- a/ext/mysql2/client.c
+++ b/ext/mysql2/client.c
@@ -883,10 +883,12 @@ static VALUE _mysql_client_options(VALUE self, int opt, VALUE value) {
       retval = &boolval;
       break;
 
+#if defined(MYSQL_SECURE_AUTH)
     case MYSQL_SECURE_AUTH:
       boolval = (value == Qfalse ? 0 : 1);
       retval = &boolval;
       break;
+#endif
 
     case MYSQL_READ_DEFAULT_FILE:
       charval = (const char *)StringValueCStr(value);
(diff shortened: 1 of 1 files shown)


yahonda:
There are 3  failures reported with this pull request.

[One of them](https://travis-ci.org/brianmario/mysql2/jobs/278781692) is explained in #893
Other two with [MariaDB 10.0](https://travis-ci.org/brianmario/mysql2/jobs/278781679) and [10.1](https://travis-ci.org/brianmario/mysql2/jobs/278781681) also reproduce with the current master branch, I have not found the reason why MariaDB 10.0 and 10.1 are failing.

ext/mysql2/client.c
```diff
@@ -1425,7 +1429,9 @@ void init_mysql2_client() {
   rb_define_private_method(cMysql2Client, "write_timeout=", set_write_timeout, 1);
   rb_define_private_method(cMysql2Client, "local_infile=", set_local_infile, 1);
   rb_define_private_method(cMysql2Client, "charset_name=", set_charset_name, 1);
+#if defined(MYSQL_SECURE_AUTH)
   rb_define_private_method(cMysql2Client, "secure_auth=", set_secure_auth, 1);
+#endif
```
sodabrew: For consistency with several other cases, let's allow this to fail silently by leaving the method in place, but turning it into a no-op.

ext/mysql2/client.c
```diff
@@ -1311,9 +1313,11 @@ static VALUE set_ssl_options(VALUE self, VALUE key, VALUE cert, VALUE ca, VALUE
   return self;
 }
 
+#if defined(MYSQL_SECURE_AUTH)
```
sodabrew: Make this method a no-op by putting the `#ifdef` inside the method definition. Please add a comment, something like:

```
/* This option was deprecated in MySQL 5.x and removed in MySQL 8.0 */

yahonda: Thanks for the review. Changed ifdef location inside the method.

ext/mysql2/client.c
yahonda: Removed ifdef here.  `set_secure_auth` method already handles if `MYSQL_SECURE_AUTH` is defined or not, here we do not need to handle it again. please give additional comment If I do not understand your review comment correctly.

PR #908, opened 2017-11-25
Remove 1.8.7,ree compatibility shim

tamird:
@sodabrew looks like I missed one (that was added after the original PR opened).

### DIFF
--- a/ext/mysql2/statement.c
+++ b/ext/mysql2/statement.c
@@ -202,6 +202,8 @@ static int my_big2ll(VALUE bignum, LONG_LONG *ptr)
 {
   unsigned LONG_LONG num;
   size_t len;
+// rb_absint_size was added in 2.1.0. See:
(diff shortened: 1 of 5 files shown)


ext/mysql2/statement.c
```diff
@@ -225,12 +227,7 @@ static int my_big2ll(VALUE bignum, LONG_LONG *ptr)
          * only -0x8000000000000000 is safe if `len == 8 && nlz_bits == 0`
          */
         !rb_absint_singlebit_p(bignum)
-#elif defined(HAVE_RB_BIG_CMP)
```
sodabrew: This case should still be needed for Ruby 1.9 and 2.0.

tamird: Oh, I misread that `efli` as `ifdef`. Derp.

tamird: Alright, this is fixed up and green now.

PR #909, opened 2017-11-26
rubocop 0.41.2 (was 0.34.0)

jaredbeck:
Now that we have dropped support for ruby 1.8.7, we should be able to upgrade rubocop to 0.41.2. Works on my machine, we'll see what travis says.

That's as far as we can go for now as rubocop 0.42 requires ruby >= 2.0.


jaredbeck:
Closed in favor of https://github.com/brianmario/mysql2/pull/752

PR #545, opened 2014-09-14
Move error_number and sql_state into Mysql2::Error initializer

sodabrew:
An alternative to #544, make all of the needed parameters into the initialize method.

PR #911, opened 2017-11-27
More specific exception classes

casperisfine:
Take over: https://github.com/brianmario/mysql2/pull/870
Partially fix: https://github.com/brianmario/mysql2/issues/404

I've just implemented 2 of them to showcase what it could look like. I can go over all error codes if dimmed necessary.

(Sorry for opening a second PR, but long story short I lost access to the first fork I did).

@sodabrew does this look good to you?

cc @rafaelfranca & @sgrif

### DIFF
--- a/lib/mysql2/error.rb
+++ b/lib/mysql2/error.rb
@@ -6,12 +6,42 @@ module Mysql2
       replace: '?'.freeze,
     }.freeze
 
-    BaseConnectionError = Class.new(Error)
+    ConnectionError = Class.new(Error)
+    TimeoutError = Class.new(Error)
 
     CODES = {
-      1045 => AccessDenied = Class.new(BaseConnectionError),
-      2005 => UnknownHost = Class.new(BaseConnectionError),
-    }
+      1205 => TimeoutError, # ER_LOCK_WAIT_TIMEOUT
+
+      1044 => ConnectionError, # ER_DBACCESS_DENIED_ERROR
+      1045 => ConnectionError, # ER_ACCESS_DENIED_ERROR
+      1152 => ConnectionError, # ER_ABORTING_CONNECTION
+      1153 => ConnectionError, # ER_NET_PACKET_TOO_LARGE
+      1154 => ConnectionError, # ER_NET_READ_ERROR_FROM_PIPE
+      1155 => ConnectionError, # ER_NET_FCNTL_ERROR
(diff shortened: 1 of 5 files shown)


sodabrew:
This looks good to me. Will need a rebase since the Ruby 1.9-syntax changes landed in another PR. Could you run down the unit test failures?

casperisfine:
@sodabrew sorry I should have checked CI after my rebase.

I fixed the test suite. Should be good to go now.

lib/mysql2/error.rb
```diff
@@ -8,25 +8,60 @@ class Error < StandardError
...
-      @server_version ||= nil
+    def initialize(msg, server_version = nil, error_number = nil, sql_state = nil)
+      @server_version = server_version
+      @error_number = error_number
+      @sql_state = sql_state ? sql_state.encode(ENCODE_OPTS) : nil
 
       super(clean_message(msg))
     end
 
     def self.new_with_args(msg, server_version, error_number, sql_state)
```
sodabrew: I remember there was a reason for all of this, but I don't remember what that reason was 🤕

sodabrew:
Thanks very much for the rebase and updating the unit tests!

lib/mysql2/error.rb
casperisfine: I remember git blaming it and figuring out that it wasn't required anymore. Let me blame again.

casperisfine: Hum, don't remember why I dimmed it unnecessary https://github.com/brianmario/mysql2/pull/545

I still think it's better this way, but I can remove that change if you prefer.

sodabrew: I moved the arguments into the constructor in f9eba44b0519580d2cd52b51d2b902f45186f427 but moved them out to a separate method in 1d4de8963c71ce88772a0e27883066160c3dbbd4

sodabrew:
LGTM. Do you want to wait for feedback from the other folks you tagged in the OP?

casperisfine:
> Do you want to wait for feedback from the other folks you tagged in the OP?

No don't worry, they're coworkers, I'll ping them more directly. And since I had to rebase again, I'd rather get this merged ASAP :p

rafaelfranca:
Thanks for doing this :+1:

PR #912, opened 2017-11-28
Accept query options on Statement#execute

sodabrew:
Instead of #640

### DIFF
--- a/ext/mysql2/statement.c
+++ b/ext/mysql2/statement.c
@@ -2,7 +2,7 @@
 
 VALUE cMysql2Statement;
 extern VALUE mMysql2, cMysql2Error, cBigDecimal, cDateTime, cDate;
-static VALUE sym_stream, intern_new_with_args, intern_each, intern_to_s;
+static VALUE sym_stream, intern_new_with_args, intern_each, intern_to_s, intern_merge_bang;
 static VALUE intern_sec_fraction, intern_usec, intern_sec, intern_min, intern_hour, intern_day, intern_month, intern_year;
 
 #define GET_STATEMENT(self) \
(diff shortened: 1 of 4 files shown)


sodabrew:
Turns out this is Ruby 2.0 syntax. Supporting the extra arguments in Ruby 1.9.3 requires a different approach to parsing out a trailing Hash in the argument list.

ext/mysql2/statement.c
```diff
@@ -416,10 +421,16 @@ static VALUE rb_mysql_stmt_execute(int argc, VALUE *argv, VALUE self) {
     return Qnil;
   }
 
+  // Important to duplicate the hash, will receive merge! if extra opts
```
tamird: Might be possible to duplicate only conditionally, or use `merge` instead?

lib/mysql2/statement.rb
```diff
@@ -7,12 +7,14 @@ class Statement
     if Thread.respond_to?(:handle_interrupt)
       def execute(*args)
         Thread.handle_interrupt(::Mysql2::Util::TIMEOUT_ERROR_CLASS => :never) do
-          _execute(*args)
+          options = args.last.is_a?(Hash) ? args.pop : {}
```
tamird: Consider leaving a TODO that this should change when 1.9 is dropped.

ext/mysql2/statement.c
sodabrew: It was duplicate previously - I should actually check if it gets duplicated again in `rb_mysql_result_to_obj` or if that method keeps the same options hash argument that was passed to it.

sodabrew: Confirmed that `rb_mysql_result_to_obj` uses what is passed to it, rather than duplicating its own query options hash, so the comment here (and in client.c) should reflect that.

PR #913, opened 2017-11-29
Drop 1.9.3 support

tamird:
@sodabrew I might've missed something, but doesn't look like there's a huge upshot to doing this (at least not until we're ready to drop 2.0.0).

### DIFF
--- a/ext/mysql2/client.h
+++ b/ext/mysql2/client.h
@@ -1,12 +1,6 @@
 #ifndef MYSQL2_CLIENT_H
 #define MYSQL2_CLIENT_H
 
-#ifndef HAVE_RB_THREAD_CALL_WITHOUT_GVL
-/* emulate rb_thread_call_without_gvl with rb_thread_blocking_region */
-#define rb_thread_call_without_gvl(func, data1, ubf, data2) \
-  rb_thread_blocking_region((rb_blocking_function_t *)func, data1, ubf, data2)
-#endif /* ! HAVE_RB_THREAD_CALL_WITHOUT_GVL */
-
 typedef struct {
   VALUE encoding;
   VALUE active_thread; /* rb_thread_current() or Qnil */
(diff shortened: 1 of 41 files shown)


ext/mysql2/client.h
```diff
@@ -1,6 +1,8 @@
 #ifndef MYSQL2_CLIENT_H
 #define MYSQL2_CLIENT_H
 
+// rb_thread_call_without_gvl was added in 2.0.0. See:
+// https://github.com/ruby/ruby/commit/c51a826.
```
sodabrew: Could be removed? Or is it needed for RBX?

sodabrew:
Can remove `if Thread.respond_to?(:handle_interrupt)` blocks in client.rb and statement.rb, too.

tamird:
Duh, good eye. Removed some more things.

sodabrew:
This matches up with all the changes I was testing locally, thank you!

tamird:
Yikes :(

sodabrew:
_I noticed a few more `if... handle interrupts?` in the specs._

tamird:
Just saw one, done.

tamird:
Interesting that RHSCL also dropped support for 2.0 at the same time they dropped 1.9. Shall we keep this train rolling?

sodabrew:
Just a quick inventory suggests that only Ubuntu Trusty and Debian Wheezy are holding out at 1.9.3/2.0.0. Wheezy EOL is May 2018, and Trusty EOL is April 2019 (ugh, as mentioned, but also by that time will be two full LTS releases behind, so I'm less worried).

Older Mac OS X is at Ruby 2.0, Debian Jessie is on Ruby 2.1, and meanwhile everyone else has congregated at Ruby 2.3 now: Debian Squeeze and Buster, Ubuntu Xenial and (probably) Bionic, OS X High Sierra, and RHSCL 7.

PR #917, opened 2017-12-01
Remove support for Ruby 2.0.0

tamird:
Not a ton of upside here; just the rubocop bump.


sodabrew:
Marking `icebox` for now, but sometime years after Ruby 2.0.0 is unsupported I will come back to drop it :)

fatkodima:
@sodabrew 

Wdyt about reviving this? I can move this PR forward if needed. 
I planned to add support for GC compactions (added in 2.7), but this gem should require ruby >= 2.2 to use `TypedData...` for this to implement.

junaruga:
Note that Travis's Ruby 2.0.0 case was removed because we have Ruby 2.0.0 case for centos 7 system Ruby case on GitHub Actions. There is also Ruby 2.1 case on GitHub Actions.
https://github.com/brianmario/mysql2/commit/ec1dac4f58e95e2725a0f9febc994e9c2d06b7f9

junaruga:
If we want to drop the Ruby 2.0.0 case on CentOS 7, I want to work to change the case from the system Ruby 2.0.0 to Ruby 2.7 by using Red Hat Software Collections rh-ruby27 RPM package on CentOS 7 case.

> I planned to add support for GC compactions (added in 2.7), but this gem should require ruby >= 2.2 to use TypedData... for this to implement.

@fatkodima Did you mean you planned to add support for the GC compactions for mysql2 gem, right? I am curious to know why do not you do the implementation of the GC compactions conditionally checking a current running Ruby version. Is it hard?

fatkodima:
> If we want to drop the Ruby 2.0.0 case on CentOS 7, I want to work to change the case from the system Ruby 2.0.0 to Ruby 2.7 by using Software Collections rh-ruby27 RPM package on CentOS 7 case.

Please, do.

> Did you mean you planned to add support for the GC compactions for mysql2 gem, right? I am curious to know why you do not do the implementation of the GC compactions conditionally checking a current running Ruby version. Is it hard?

Yes, I will try conditional checking and see how clunky that would be.

junaruga:
> Please, do.

If @sodabrew agree we drop Ruby 2.0.0 support, I can work on it. Because the system Ruby 2.0.0 on CentOS 7 is currently the only Ruby 2.0.0 case.

> Yes, I will try conditional checking and see how clunky that would be.

OK. Thanks. Dropping a Ruby version's requirement is really critical. In my opinion, we should not do it on the maintenance version release, and we can do it on the major or minor version release. The timing to drop the requirement is important.

junaruga:
But I think nowadays we are trying to fix the test failures to make every case in our CI success to release the next maintenance version at first. Dropping old Rubies versions could be after that.

tamird:
Nothing valuable here; seems Ruby 2.0 support is here for the long term.

Issue #916, opened 2017-12-01
Big Endian environment issues


junaruga:
## Summary

I am trying to build `mysql2` and run the unit test on not only `x86_64` (little endian), but also `ppc64` (big endian).

1. There are 2 failed failures on the ppc64.

- 1.1. 1st failure is for c client option `local_infile`. I sent PR for that. (https://github.com/brianmario/mysql2/pull/914).
- 1.2 2nd failure is for result object's `YEAR` column.

2. I want to suggest to add big endian environment to Travis CI test.

## 1.2 The detail of Issue for result object's `YEAR` column.

Below test is failed on big endian environments such as ppc64 and s390x architectures.

```
  3) Mysql2::Statement row data type mapping should return Fixnum for a YEAR value
     Failure/Error: expect(@test_result['year_test']).to eql(2009)
       expected: 2009
            got: 131661824
       (compared using eql?)
     # ./spec/mysql2/statement_spec.rb:444:in `block (3 levels) in <top (required)>'
```

  When I ran with below debug logs,

ext/mysql2/result.c rb_mysql_result_fetch_row_stmt

```
         case MYSQL_TYPE_YEAR:         // int
+          printf("debug rb_mysql_result_fetch_row_stmt type_year.\n");
           if (result_buffer->is_unsigned) {
             val = UINT2NUM(*((unsigned int*)result_buffer->buffer));
+            printf("debug rb_mysql_result_fetch_row_stmt type_year unsinged. buffer: [%d], length: [%ld]\n",
+              *((unsigned int*)result_buffer->buffer), result_buffer->buffer_length);
           } else {
             val = INT2NUM(*((int*)result_buffer->buffer));
+            printf("debug rb_mysql_result_fetch_row_stmt type_year singed. [%d]\n",
+              *((int*)result_buffer->buffer));
           }
           break;
```

The result is like this on x86_64 (little endian)
2009 => 111 11011001 on binary number

```
debug rb_mysql_result_fetch_row_stmt type_year unsinged. buffer: [2009], length: [4]
```

```
irb(main):002:0> 2009.to_s(2)
=> "11111011001"
```

And like this on ppc64 (big endian)

```
debug rb_mysql_result_fetch_row_stmt type_year unsinged. buffer: [131661824], length: [4]
```

131661824 => 111 11011001 00000000 00000000 on binary number

```
irb(main):001:0> 131661824.to_s(2)
=> "111110110010000000000000000"
```

But I have no idea about how to fix this issue.
[...]

PR #919, opened 2017-12-03
Revisit PR #840

sodabrew:
See #916


sodabrew:
This might not be needed after all, as it turns out #916 was a bug in MariaDB Connector/C 3.0.

sodabrew:
I like that this removes a warning on older versions of MySQL. Merging.

PR #922, opened 2017-12-07
GitHub is HTTPS by default

amatsuda:
This reduces unneeded redirection from http://... to https://... when someone visits this gem's homepage via https://rubygems.org/gems/mysql2.


sodabrew:
Sounds reasonable to me. Thank you!

PR #931, opened 2018-01-05
Remove 'mysql' gem and benchmarks

tamird:
The 'mysql' gem hasn't been (and likely won't be) updated to work with
RUBY_INTEGER_UNIFICATION (see https://bugs.ruby-lang.org/issues/12005).

See also https://github.com/luislavena/mysql-gem/pull/32.

Benchmarks against this gem are no longer relevant. Note that the
remaining benchmark numbers in [README.md] haven't been updated.


tamird:
@sodabrew any interest in this?

sodabrew:
Thanks, no, I don't see a reason to remove these benchmarks. I suppose it's worth a story of what prompted this gem at the time vs. where things are at now but wouldn't remove the scripts anyhow.

junaruga:
@tamird I suppose your motivation for this PR comes from you who saw the `mysql` gem error by `bundle install`, right? Good new is the `mysql` gem and the benchmark groups are optional by the PR #1173. You do not see the error related to `mysql`. I think we can close this PR. What do you think?

tamird:
Works for me.

PR #907, opened 2017-11-24
Suppress Fixnum and Bignum warnings on Ruby 2.4.

junaruga:
This PR would suppress `Fixnum` and `Bignum` warnings on Ruby 2.4 when running tests.

```
$ bundle exec rspec spec -f p
...
/home/jaruga/git/mysql2/spec/mysql2/client_spec.rb:977: warning: constant ::Fixnum is deprecated
...
/home/jaruga/git/mysql2/spec/mysql2/statement_spec.rb:423: warning: constant ::Bignum is deprecated
...
```

The `0.class` hack is used in https://github.com/rails/rails .

PR #939, opened 2018-02-14
Update usage section of README.md

koic:
Follow up of #907.

Since `Fixnum` is a subclass of `Integer`, I thought it would be OK to replace it with `Integer`. Especially, It will be a calm expression for Ruby 2.4 or higher users.


sodabrew:
Thank you!

PR #957, opened 2018-04-04
Avoid GC run between `mysql_stmt_execute` and `mysql_stmt_store_result`

### LINKED ISSUE #956: Random failures using prepared statements
While investigating some random errors in my MySQL connections, I have discovered that using prepared statements as described in the mysql2 documentation will almost invariably result in random errors, most notably `Commands out of sync`.

Example code:

```
gem 'mysql2', '0.5.0'
require 'mysql2'

mysql_spec = {
  host: ENV['HOSTNAME'],
  username: ENV['USERNAME'],
  password: ENV['PASSWORD'],
  database: ENV['DATABASE']
}

connection_1 =  Mysql2::Client.new(mysql_spec)

1000.times do |n|
  statement_1 = connection_1.prepare("SELECT * FROM test WHERE test_id = ?")
  result_1    = statement_1.execute(1).to_a
end

```

The above code has been tested across various MySQL / Maria installations using both mysql 0.4.10 and 0.5.0, and usually fails before completing 1000 iterations:

```
/home/charlie/.rvm/gems/ruby-2.4.0/gems/mysql2-0.5.0/lib/mysql2/statement.rb:7:in `_execute': Commands out of sync; you can't run this command now (Mysql2::Error)
	from /home/charlie/.rvm/gems/ruby-2.4.0/gems/mysql2-0.5.0/lib/mysql2/statement.rb:7:in `block in execute'
	from /home/charlie/.rvm/gems/ruby-2.4.0/gems/mysql2-0.5.0/lib/mysql2/statement.rb:6:in `handle_interrupt'
	from /home/charlie/.rvm/gems/ruby-2.4.0/gems/mysql2-0.5.0/lib/mysql2/statement.rb:6:in `execute'
	from mysqltest2.rb:15:in `block in <main>'
	from mysqltest2.rb:13:in `times'
	from mysqltest2.rb:13:in `<main>'
```

By trial and error, I discovered 2 ways to resolve this. Firstly the problem doesn't occur when GC is disabled.

```
GC.disable
```

Alternatively, this can be resolved by changing the code to explicitly close the statement after each execute has completed.

```
1000.times do |n|
  statement_1 = connection_1.prepare("SELECT * FROM test WHERE test_id = ?")
  result_1    = statement_1.execute(1).to_a
  statement_1.close
end
```

I would appreciate any advice on whether this is a bug, or whether I am simply using the API incorrectly. If it is the latter, could this be a documentation problem? Thanks!

kamipo:
This fixes a regression caused by #912 due to calling `rb_funcall`
between `mysql_stmt_execute` and `mysql_stmt_store_result`, it will
cause `mysql_stmt_close` to be called in wrong order.

Fixes #956.

### DIFF
--- a/ext/mysql2/statement.c
+++ b/ext/mysql2/statement.c
@@ -426,12 +426,12 @@ static VALUE rb_mysql_stmt_execute(int argc, VALUE *argv, VALUE self) {
   (void)RB_GC_GUARD(current);
   Check_Type(current, T_HASH);
 
-  // Merge in hash opts/keyword arguments
-  if (!NIL_P(opts)) {
-    rb_funcall(current, intern_merge_bang, 1, opts);
+  is_streaming = (Qtrue == rb_hash_aref(current, sym_stream));
+
+  if (!is_streaming && !NIL_P(opts)) {
+    is_streaming = (Qtrue == rb_hash_aref(opts, sym_stream));
   }
 
-  is_streaming = (Qtrue == rb_hash_aref(current, sym_stream));
   if (!is_streaming) {
     // recieve the whole result set from the server
     if (mysql_stmt_store_result(stmt)) {
@@ -439,6 +439,11 @@ static VALUE rb_mysql_stmt_execute(int argc, VALUE *argv, VALUE self) {
       rb_raise_mysql2_stmt_error(stmt_wrapper);
     }
     wrapper->active_thread = Qnil;
+  }
+
+  // Merge in hash opts/keyword arguments
+  if (!NIL_P(opts)) {
+    rb_funcall(current, intern_merge_bang, 1, opts);
   }
 
   resultObj = rb_mysql_result_to_obj(stmt_wrapper->client, wrapper->encoding, current, metadata, self);


sodabrew:
Nice find! I am worried that this will be fragile, and some other call into Ruby space can still creep in here. I'm looking at whether there is a missing refcount.

sodabrew:
In result.c, `rb_mysql_result_to_obj`, there's a comment I left a while back:

     /* Keep a handle to the Statement to ensure it doesn't get garbage collected first */
     wrapper->statement = statement;
     if (statement != Qnil) {
       wrapper->stmt_wrapper = DATA_PTR(statement);
       wrapper->stmt_wrapper->refcount++;
     } else {
       wrapper->stmt_wrapper = NULL;
     }

I am wondering how the statement is possibly getting garbage collected at the point of this `rb_funcall`, if later down in the file, `self` is passed into `rb_mysql_result_to_obj`. I tried adding `RB_GC_GUARD(self)` at several location in `rb_mysql_stmt_execute` but that did not help.

Your solution avoids `rb_funcall` - but I read through `rb_hash_aref` and found that it is possible to call `rb_funcall` if the hash has a default value callback and the key is missing. This is admittedly not likely, but I was able to reproduce the break by adding this to the repro script:

```
connection_1.query_options.default_proc = ->(h, k) { puts "Hello #{k}" }
```

The output `Hello stream` prints a few times and then fails with `Commands out of sync` again.

I put together a different movement of code and was able to protect against this failure as well.

Bottom line, a GC run between `mysql_stmt_execute` and `mysql_stmt_store_result` is collecting the statement handle, and I don't understand why.

sodabrew:
I found a commit to the sqlite gem showing another case of a hard GC workaround: https://github.com/sparklemotion/sqlite3-ruby/commit/380287ae5160fa460132f8cf6b65067adbeb9df4

sodabrew:
Aha, I get it now. The issue is that once a statement begins executing, _any other command is invalid_ such as freeing a much older statement handle, which requires a notification from the client library to the server to release the handle, but isn't valid if the server is in results-streaming-mode.

sodabrew:
Which leads me to wonder... what about streaming results? Yep, that breaks, too! In the repro script:

```
result_1    = statement_1.execute(1, stream: true, cache_rows: false).to_a
```

Even with both of our proposed fixes:
```
outofsync.rb:17:in `each': Row retrieval was canceled by mysql_stmt_close() call (Mysql2::Error)
```

kamipo:
Yeah, originally fixing the case of `stream: true` is out of target in my fix since I didn't know whether it is possible or not.
If possible that, probably it will solve both cases.

I'm closing this in favor of #958.

PR #962, opened 2018-04-14
Rebase of PR 807: non-blocking behavior for `next_result`

sodabrew:
Rebase of #807 

/cc @osheroff

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -22,7 +22,7 @@ Metrics/AbcSize:
 # Offense count: 31
 # Configuration parameters: CountComments, ExcludedMethods.
 Metrics/BlockLength:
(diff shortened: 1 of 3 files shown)


sodabrew:
There are actually more possible states than my implementation accounts for. I'm not sure if there's a clean way to do this without the underlying methods exposed.

sodabrew:
Closing as a historical artifact. This approach (reaching into vio_priv.h to detect already-buffered data before deciding whether to poll the socket) needs revisiting, not resurrecting as-is — see my comment on #1033 for the full history and current status. Continuing this under a fresh branch/PR that starts from a reproducing test and a survey of what current libmysqlclient/MariaDB Connector/C offer for this.

PR #966, opened 2018-04-27
Change mainly tested Ruby from 2.4 to 2.5 on Travis CI.

junaruga:
Would you mind to change mainly tested Ruby from 2.4. to 2.5?
Because I want to see the test case of Ruby 2.5 and MariaDB 10.2, and the latest stable Ruby version is now 2.5.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,47 +28,47 @@ rvm:
   - ruby-head
 matrix:
   include:
-    - rvm: 2.4
+    - rvm: 2.5
       env: DB=mariadb10.0
       addons:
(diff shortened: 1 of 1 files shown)


majormoses:
Personally I think we should test both versions ruby 2.4 is still widely used and its not scheduled to be EOL any time soon. Just my `$0.02`

junaruga:
I am fine not to adapt 2.5 for the cases for now. It might be not good time to change 2.4. :)

junaruga:
I am happy to close this PR, as we mainly start to use GitHub Actions now.

PR #971, opened 2018-05-08
Handle read_timeout for queries executed within EventMachine

gmodarelli:
Hello, I'm not sure if this is the best approach to handle a `read_timeout` when a query is being run within an EventMachine reactor loop, but I wanted to give it a try and pick your brain on the topic :) 

The `read_timeout` option is only used when a query is executed synchronously, so it will be completely ignored for `EM` clients.

https://github.com/brianmario/mysql2/blob/b3fe727b56ba5c8cb5678200f2e99a7119bc41cc/ext/mysql2/client.c#L782-L792

In order to expose a similar behaviour I've introduced a [Deferrable Timeout](https://www.rubydoc.info/github/eventmachine/eventmachine/EventMachine/Deferrable#timeout-instance_method) which will raise a `Mysql2::EM::ReadTimeout`.

In case of a read timeout from the normal non-evented client, a `Mysql::Error` is raised, but I am not sure it would be ok to raise that exception in this case.

I also understand that a change like this one would break compatibility. Maybe there is a better way to expose this timeout logic to the client code. On the other hand it is also a bit counter intuitive to have a `read_timeout` option that doesn't do anything when the client is evented :)

Please let me know what you think about it.
Have a nice day!

### DIFF
--- a/spec/em/em_spec.rb
+++ b/spec/em/em_spec.rb
@@ -65,6 +65,21 @@ begin
       end.to raise_error('some error')
     end
 
+    it "should timeout if we wait longer than :read_timeout" do
+      expect do
+        EM.run do
+          client = Mysql2::EM::Client.new DatabaseCredentials['root'].merge(read_timeout: 1)
+          defer = client.query "SELECT sleep(2)"
+          defer.callback do
+            # This _shouldn't_ be run, but it needed to prevent the specs from
+            # freezing if this test fails.
+            client.close
+            EM.stop_event_loop
+          end
+        end
+      end.to raise_error(Mysql2::EM::ReadTimeout)
+    end
+
     context 'when an exception is raised by the client' do
       let(:client) { Mysql2::EM::Client.new DatabaseCredentials['root'] }
       let(:error) { StandardError.new('some error') }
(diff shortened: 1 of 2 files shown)


gmodarelli:
I think the failing test on `Ruby 2.3.4` is just a random failure. I've run the tests locally with `Ruby 2.3.4` and they are green

spec/em/em_spec.rb
```diff
@@ -65,6 +65,21 @@
...
+          client = Mysql2::EM::Client.new DatabaseCredentials['root'].merge(read_timeout: 1)
+          defer = client.query "SELECT sleep(2)"
+          defer.callback do
+            # This _shouldn't_ be run, but it needed to prevent the specs from
+            # freezing if this test fails.
+            client.close
+            EM.stop_event_loop
+          end
+        end
+      end.to raise_error(Mysql2::EM::ReadTimeout)
```
sodabrew: Can you switch to the `defer.errback` pattern see in the tests below? It is problematic to throw exceptions outside the scope of `EM.run` (throwing exceptions like this in fact would crash EM 1.2.x until I finally finally figured out the problem in EM 1.2.6)

gmodarelli: Sure, I'll switch to `defer.errback` :)

PR #976, opened 2018-05-18
Updating the mysql2_mysql_enc_to_rb conversion table to 8.0 List

### LINKED ISSUE #975: Segfault Debian 9 with MySQL 8.0 using libmysqlclient21
I have recently been trying to update the MySQL version at my work to 8, and the first thing I did was to upgrade my mysql2 dependency to 0.5.1 and ActiveRecord to 5.2.0. This resulted in a segmentation fault when trying to query a simple table. The first thing I tried doing was to use the apt repository provided by the MySQL team to use their build of libmysqlclient21 which got the process farther along but still resulted in a segfault. Any ideas on how to fix this? 

[Link to Segfault Trace](https://gist.github.com/QuantamHD/49704e6f9dd35ab138094662a59c98ff)

QuantamHD:
This commit fixes #975 what happened is that I used the collation utf8mb4_general_ci which is listed as charsetnr 254 in the C API our conversion table mysql2_mysql_enc_to_rb did not have entries that went past 245 previously. Thus, functions that used the result of this selection would be given an index into undefined memory and when they tried to read it they would blow past the end of the segment causing the issue in #975. I've added a check to return NULL if we see a charsetnr that is larger than the conversion table.

### DIFF
--- a/ext/mysql2/mysql_enc_to_ruby.h
+++ b/ext/mysql2/mysql_enc_to_ruby.h
@@ -245,5 +245,15 @@ static const char *mysql2_mysql_enc_to_rb[] = {
   "UTF-8",
   "UTF-8",
   "UTF-8",
+  "UTF-8",
+  NULL,
+  NULL,
+  NULL,
+  NULL,
+  NULL,
+  NULL,
+  NULL,
   "UTF-8"
 };
+
+#define CHARSETNR_SIZE (sizeof(mysql2_mysql_enc_to_rb)/sizeof(mysql2_mysql_enc_to_rb[0]))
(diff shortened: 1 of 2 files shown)


QuantamHD:
@timcharper It looks like the Travis build is failing to download some dependencies.

sodabrew:
Thanks for worrying on the MySQL 5.5 build - I'm OK merging this without fixing Travis for that version.

sodabrew:
This PR looks good to me! Is there anything else you want to do on this branch before I merge it?

QuantamHD:
@sodabrew nope it's all good from my side.

Issue #988, opened 2018-06-28
Adding CentOS to Travics CI


junaruga:
I would propose adding CentOS (+ Fedora Linux) environment to Travis CI.

As you know, Travis CI is run on Ubuntu.
However using Docker, it is possible to run tests on CentOS and Fedora.

The reason why I want to add it is because we sometimes receive questions about installation on CentOS. It is also useful for people to see CentOS test case is passed casually, and to save maintainer's time.

Personally I am maintaining the RPM package of `mysql2`. So, it's good for me to see the passed test. :)

I want to show the example by sending PR.

How do you think?

sodabrew:
Resolved by #989

PR #989, opened 2018-06-28
Add CentOS on Travis CI.

junaruga:
Related: https://github.com/brianmario/mysql2/issues/988 .

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,6 @@
 sudo: required
 dist: trusty
+services: docker
 language: ruby
 bundler_args: --without benchmarks development
 # Pin Rubygems to a working version. Sometimes it breaks upstream. Update now and then.
@@ -73,6 +74,11 @@ matrix:
       addons:
         hosts:
           - mysql2gem.example.com
+    - rvm: 2.4
+      env: DOCKER=centos
+      before_install: true
+      install: docker build -t mysql2 -f .travis_Dockerfile_centos .
+      script: docker run --add-host=mysql2gem.example.com:127.0.0.1 -t mysql2
   fast_finish: true
   allow_failures:
     - rvm: ruby-head
--- a/.travis_Dockerfile_centos
+++ b/.travis_Dockerfile_centos
@@ -1,0 +1,23 @@
+FROM centos:7
+
+WORKDIR /build
+COPY . .
+
+RUN yum -y update
+RUN yum -y install epel-release
+# The options are to install faster.
+RUN yum -y install \
+  --setopt=deltarpm=0 \
+  --setopt=install_weak_deps=false \
+  --setopt=tsflags=nodocs \
+  mariadb-server \
+  mariadb-devel \
+  ruby-devel \
+  git \
+  gcc \
+  gcc-c++ \
+  make
+RUN gem update --system > /dev/null
+RUN gem install bundler
+
+CMD sh .travis_centos.sh
--- a/.travis_centos.sh
+++ b/.travis_centos.sh
@@ -1,0 +1,13 @@
+#!/usr/bin/env bash
+
+set -eux
+
+# Start mysqld service.
+sh .travis_setup_centos.sh
+
+bundle install --path vendor/bundle --without benchmarks development
+
+# USER environment value is not set as a default in the container environment.
+export USER=root
+
+bundle exec rake
--- a/spec/mysql2/client_spec.rb
+++ b/spec/mysql2/client_spec.rb
@@ -135,15 +135,22 @@ RSpec.describe Mysql2::Client do
 
     # You may need to adjust the lines below to match your SSL certificate paths
     ssl_client = nil
+    option_overrides = {
+      'host'     => 'mysql2gem.example.com', # must match the certificates
+      :sslkey    => '/etc/mysql/client-key.pem',
+      :sslcert   => '/etc/mysql/client-cert.pem',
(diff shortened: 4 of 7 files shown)


spec/mysql2/client_spec.rb
```diff
@@ -135,15 +136,22 @@ def connect(*args)
 
     # You may need to adjust the lines below to match your SSL certificate paths
     ssl_client = nil
+    option_overrides = {
```
junaruga: This modification is to pass the test on the CentOS environment that does not have `/etc/mysql/*.pem` files.

spec/mysql2/result_spec.rb
```diff
@@ -164,7 +164,7 @@
       expect do
         res.each_with_index do |_, i|
           # Exhaust the first result packet then trigger a timeout
-          sleep 2 if i > 0 && i % 1000 == 0
+          sleep 4 if i > 0 && i % 1000 == 0
```
junaruga: This modification is to pass the test on CentOS.

spec/mysql2/client_spec.rb
```diff
@@ -1,5 +1,6 @@
 require 'spec_helper'
 
+# rubocop:disable Metrics/BlockLength
```
junaruga: By below modification, the length of the block is more than the limitation.
As I was not sure how to fix, I added this ignored line.

.travis.yml
```diff
@@ -73,6 +74,12 @@ matrix:
       addons:
         hosts:
           - mysql2gem.example.com
+    - rvm: 2.4
```
junaruga: This part is a little bit tricky. Actually the tested Ruby on CentOS is not Ruby 2.4 but default installed Ruby on CentOS. But I was not sure how to update `rvm: foo` to show `rvm` is not used directly on Travis.

.travis_centos.sh
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env bash
+
+set -eux
+
+# Install dependency packages and start mysqld service.
+sh .travis_setup_centos.sh
+
+# To avoid install error.
+sed -i '/eventmachine/d' Gemfile
```
simi: Why is this needed? Can't we fix the install error?

junaruga: Thanks for the comment.
It is to avoid install error by `bundle install` on CentOS.
But let me check again to fix it.

junaruga: The reason of the error is just because `g++` command was not installed on the environment.

junaruga: I fixed it and rebased.

.travis.yml
```diff
@@ -73,6 +74,12 @@ matrix:
...
+      # Mark
```
sodabrew: Who is Mark?

.travis_Dockerfile_centos
```diff
@@ -0,0 +1,6 @@
+FROM centos
```
sodabrew: Suggest to use tagged version with `centos:7`.

.travis_centos.sh
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env bash
+
+set -eux
+
+# Install dependency packages and start mysqld service.
+sh .travis_setup_centos.sh
```
sodabrew: I think this should go into the Dockerfile, so that you can benefit from layer caching of these dependencies.

spec/mysql2/client_spec.rb
sodabrew: See file .rubocop_todo.yml.

But you may be right that the better approach is not to globally increase the limit, but to ignore it for this file and have a more sensible limit apply everywhere else.

junaruga:
@sodabrew thank you for the review! I am sorry I did not noticed until now. I will check those later.

.travis.yml
junaruga: "Mark" is not people's name in this case. I want to describe `env: DOCKER=centos` is just used to identify this test case from Travis test page. But I am going to remove the line.

spec/mysql2/client_spec.rb
junaruga: Okay, I will change current limit 850 to 860 to prevent below situation.

```
spec/mysql2/client_spec.rb:3:1: C: Metrics/BlockLength: Block has too many lines. [851/850]
```

.travis_Dockerfile_centos
junaruga: All right. I will change it to `centos:7`.

.travis_centos.sh
junaruga: All right. Let me move the logic to the Dockefile.

junaruga: There is a problem to move the `.travis_setup_centos.sh` to `Dockerfile` RUN (that is executed on `docker build`'s timing).
We can not start mysqld daemon service in the process of `docker  build`. (Can not access the service).
So, I moved only the part to install dependency to the Dockerfile RUN part.
I add the modification as another commit to see the change easily for now.
How?
Thank you for the checking.

junaruga:
@sodabrew I modified it!

sodabrew:
Looks good to me! Are you ready for merge, or still hacking on this PR?

junaruga:
@sodabrew I am ready for merge! I did squash the 2 commits to 1 commit now.

junaruga:
Thank you! :smile:

PR #1003, opened 2018-09-17
added #empty? to MySql2::Result

walf443:
```ruby
result = client.qeury("SELECT 1 FROM mysql2_test")
unless result.empty? # currently it need to write `if result.count > 0`
   # do something with `result`
end
```

MySql2::Result include Enumerable. If it added `#empty?` method, I'm happy to write a code with more rubyish style.


sodabrew:
#empty? Isn’t a usual part of Enumerable though: https://ruby-doc.org/core-2.5.1/Enumerable.html

But Array and Hash do have it. So I can see the utility in the convenience method.

Let me think a bit before adding this to the api.

sodabrew:
Accepted for next major.

walf443:
oh I see. thanks.

PR #1011, opened 2018-10-31
Add support for returning result rows as structs. (:as => :struct)

joe07734:
This patch adds a new query option to return result rows as structs.

At [Bandcamp](https://bandcamp.com) we've been using this patch (and the other) in our fork of the mysql2 gem since the beginning. We think dot notation is cleaner and more object-oriented-looking, and wanted to use it in our code that processes query results. I believe it can also make the code faster if it's doing a lot of column accesses.

This is a default for all our queries:

```
opts = Mysql2::Client.default_query_options
opts[:as] = :struct
```
And query results can be used like this:

```
band = query(<<-SQL
    SELECT band.name, bio.bio
    FROM bands AS band
    LEFT JOIN bios AS bio ON bio.band_id = band.id
    WHERE band.id = #{band_id}
    SQL
).first

puts "#{band.name} - #{band.bio}"
```

I've been meaning to make these two pull requests since forever. We've had years of experience with them and think they're useful. If you have any questions about how Bandcamp's Linux/MySQL/Ruby stack works, just ask.

### DIFF
--- a/ext/mysql2/result.c
+++ b/ext/mysql2/result.c
@@ -304,7 +304,7 @@ static VALUE cast_row_as_struct(VALUE self, VALUE rowVal, mysql2_result_wrapper 
     for (i = 0; i < wrapper->numberOfFields; i++) {
       argv_fields[i] = rb_mysql_result_fetch_field(self, i, 1);
     }
-    wrapper->rowStruct = rb_funcallv(rb_cStruct, intern_new, (int) wrapper->numberOfFields, argv_fields);
+    wrapper->rowStruct = rb_funcall2(rb_cStruct, intern_new, (int) wrapper->numberOfFields, argv_fields);
   }
 
   return rb_struct_alloc(wrapper->rowStruct, rowVal); 
--- a/ext/mysql2/result.c
+++ b/ext/mysql2/result.c
@@ -861,7 +861,6 @@ static VALUE rb_mysql_result_each(int argc, VALUE * argv, VALUE self) {
   }
 
   symbolizeKeys = RTEST(rb_hash_aref(opts, sym_symbolize_keys));
-  rowsAs        = AS_HASH;
   castBool      = RTEST(rb_hash_aref(opts, sym_cast_booleans));
   cacheRows     = RTEST(rb_hash_aref(opts, sym_cache_rows));
   cast          = RTEST(rb_hash_aref(opts, sym_cast));
@@ -872,6 +871,8 @@ static VALUE rb_mysql_result_each(int argc, VALUE * argv, VALUE self) {
   } else if (as_opt == sym_struct) {
     rowsAs = AS_STRUCT;
     symbolizeKeys = 1;  /* force */
+  } else {
+    rowsAs = AS_HASH;
   }
 
   if (wrapper->is_streaming && cacheRows) {
--- a/ext/mysql2/result.c
+++ b/ext/mysql2/result.c
@@ -722,7 +722,6 @@ static VALUE rb_mysql_result_fetch_row(VALUE self, MYSQL_FIELD * fields, const r
   return rowVal;
 }
 
-
 static VALUE rb_mysql_result_fetch_fields(VALUE self) {
   unsigned int i = 0;
   short int symbolizeKeys = 0;
--- a/ext/mysql2/result.c
+++ b/ext/mysql2/result.c
@@ -48,7 +48,6 @@ static VALUE sym_struct;
 #define AS_ARRAY  1
 #define AS_STRUCT 2
 
-
 /* Mark any VALUEs that are only referenced in C, so the GC won't get them. */
 static void rb_mysql_result_mark(void * wrapper) {
   mysql2_result_wrapper * w = wrapper;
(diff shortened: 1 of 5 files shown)


joe07734:
Fixed for those failed tests, now going to update the unit tests for statements, like in the other PR.

ext/mysql2/result.c
```diff
@@ -826,11 +861,19 @@ static VALUE rb_mysql_result_each(int argc, VALUE * argv, VALUE self) {
   }
 
   symbolizeKeys = RTEST(rb_hash_aref(opts, sym_symbolize_keys));
-  asArray       = rb_hash_aref(opts, sym_as) == sym_array;
+  rowsAs        = AS_HASH;
```
sodabrew: Nit: make this the `else` case in the block below. (or the `default` case, and convert to a `switch` statement)

ext/mysql2/result.c
```diff
@@ -464,13 +488,17 @@ static VALUE rb_mysql_result_fetch_row_stmt(VALUE self, MYSQL_FIELD * fields, co
...
-    } else {
+    if (args->rowsAs == AS_HASH) {
       rb_hash_aset(rowVal, field, val);
+    } else /* array or struct */ {
+      rb_ary_push(rowVal, val);
     }
   }
 
+  if (args->rowsAs == AS_STRUCT) {
+    rowVal = cast_row_as_struct(self, rowVal, wrapper);
```
sodabrew: This is still using an intermediary array in the row fetch loop above. Is that better/more efficient to create the struct once than to incrementally add members to the struct in the same way elements are pushed into the array or added into the hash? I think the latter would have more clarity / all three options being handled as cases in the same loop.

ext/mysql2/result.c
```diff
@@ -498,10 +526,12 @@ static VALUE rb_mysql_result_fetch_row(VALUE self, MYSQL_FIELD * fields, const r
...
   }
-  if (args->asArray) {
-    rowVal = rb_ary_new2(wrapper->numberOfFields);
-  } else {
+  
+  if (args->rowsAs == AS_HASH) {
     rowVal = rb_hash_new();
+  } else /* array or struct */ {
+    /* struct uses array as an intermediary */
+    rowVal = rb_ary_new2(wrapper->numberOfFields);
```
sodabrew: These don't get GC marked, why does the struct variant need to be GC marked? (`w->rowStruct` above) (Or conversely, should these also get GC marked?)

sodabrew:
This is also awesome. Few comments. Thank you for contributing upstream!

ext/mysql2/result.c
joe07734: Done.

ext/mysql2/result.c
joe07734: I was trying to avoid modifying the the main loop for clarity! It also allowed me to factor the "cast_row_as_struct" code with a simple parameter list. The statement support added a lot of duplicate code and I really wanted to not contribute to that. Also, the struct.new native call has a nice array form.

joe07734: Happy to rethink this if you think it's critical.

ext/mysql2/result.c
joe07734: I admit I'm hazy on when something needs to be gc-marked and when it doesn't. The "prototype" struct needs to be marked so that it'll stick around while control hops between the gem and ruby, while enumerating the results. Each row result struct, tho', is an instance of this prototype struct and is made right here, just like the hash and array, and returned to the client asap and will be gc marked there. That's my understanding anyway.

sodabrew:
Closing -- the core approach (returning rows as `Struct`s) has real correctness problems beyond needing a rebase, and I don't think it's mergeable as designed.

`Struct.new` doesn't validate member names, so any column named after a method `Struct`/`Object`/`Kernel` already defines gets silently shadowed rather than rejected. I checked this empirically:

```ruby
S = Struct.new(:class, :name)
row = S.new("Users", "Alice")
row.class   # => "Users"   -- not the Ruby class
```

This isn't a rare edge case -- `class`, `id`, `type`, `name`, `hash`, `send`, `count`, `size` are all completely ordinary column names. And it fails silently: no error, no warning, just a wrong-typed value flowing into any code that calls `.class`/`.hash`/etc. on the row (serializers, `pp`/`inspect`, type-dispatch, memoization, debuggers) -- all of which assume that invariant unconditionally because in ordinary Ruby it's always safe.

Two more failure modes, also verified:

- **Duplicate names** (the common case in joins, e.g. two tables both having an `id` column) raise outright: `Struct.new(:id, :id) # ArgumentError: duplicate member: id`. Hash mode today just lets the second key win; Struct mode is strictly more fragile for one of the most ordinary query shapes there is.
- **Unaliased expressions** (`SELECT COUNT(*)`, `SELECT 1+1`) produce field names MySQL sends literally as the expression text. `Struct.new(:"COUNT(*)")` succeeds, but the member is then unreachable via the dot syntax that's the entire point of this feature -- `row.COUNT(*)` parses as a different, nonexistent method call and raises `NoMethodError`. So for any non-trivial query, the feature's core benefit silently stops working and you're back to `row[:"COUNT(*)"]`, no better than a Hash, while still carrying the two risks above.

A rework would need to actually solve the name-safety problem (validate and raise clearly, or fall back to hash mode) rather than just rebase the diff -- that's real design work, not a mechanical fix.

Related: #1499 adds `Result#tables`/`Result#dbs` for the table/join-disambiguation use case this PR's join example was partly aiming at, without touching row shape at all.

PR #1022, opened 2019-02-01
Use vio_is_connected to check connected state

csfrancis:
### Problem

The `CONNECTED` define which is used by the `closed?` method, does not detect if the server has remotely closed the connection. In this case, should the remote side terminate the connection, the call to `closed?` will incorrectly return `true`. As a result, it's possible for subsequent query to raise an exception once the client actually detects that the connection has been closed.

### Solution

Use the `vio_is_connected` function to detect if the connection has been closed. The implementation will perform a non-blocking read on the socket to determine if the connection has been closed.

This function has been available since at least MySQL 5.6.x (from what I've Googled), so should be reliable. The symbol is globally visible as part of libmysqlclient shared library, but is not declared in the libmysqlclient headers (which is why I had to declare it inline). It does not seem as though the API has changed between 5.6 -> 8.0, so I feel safe using it here.


ext/mysql2/client.c
```diff
@@ -26,7 +26,8 @@ static ID intern_brackets, intern_merge, intern_merge_bang, intern_new_with_args
   }
 
 #if defined(HAVE_MYSQL_NET_VIO) || defined(HAVE_ST_NET_VIO)
-  #define CONNECTED(wrapper) (wrapper->client->net.vio != NULL && wrapper->client->net.fd != -1)
+  my_bool vio_is_connected(Vio *vio);
+  #define CONNECTED(wrapper) (wrapper->client->net.vio != NULL && wrapper->client->net.fd != -1 && vio_is_connected(wrapper->client->net.vio))
 #elif defined(HAVE_MYSQL_NET_PVIO) || defined(HAVE_ST_NET_PVIO)
   #define CONNECTED(wrapper) (wrapper->client->net.pvio != NULL && wrapper->client->net.fd != -1)
```
sodabrew: Please add the `vio_is_connected` call here as well, or a reason why it's not there (such as, and I don't know the truth of this, if net.pvio is older versions that don't have this function in the library at all).

spec/mysql2/client_spec.rb
```diff
@@ -573,7 +586,7 @@ def run_gc
       end
       expect do
         @client.query("SELECT SLEEP(1)")
-      end.to raise_error(Mysql2::Error, /Lost connection to MySQL server/)
+      end.to raise_error(Mysql2::Error, 'MySQL client is not connected')
```
sodabrew: Why the different error text?

ext/mysql2/client.c
```diff
@@ -26,7 +26,8 @@ static ID intern_brackets, intern_merge, intern_merge_bang, intern_new_with_args
   }
 
 #if defined(HAVE_MYSQL_NET_VIO) || defined(HAVE_ST_NET_VIO)
-  #define CONNECTED(wrapper) (wrapper->client->net.vio != NULL && wrapper->client->net.fd != -1)
+  my_bool vio_is_connected(Vio *vio);
```
sodabrew: Please move this to a header file and add an autoconf test that checks if the header is available in the given client library (there are more client libraries than just libmysqlclient).

spec/mysql2/client_spec.rb
csfrancis: I think it's because we're validating the connection before executing the query (and it's now failing) and it raises with this error text.

ext/mysql2/client.c
csfrancis: 👍 Yeah, there's some symbol resolution issues I'm seeing in CI. I'm going to add some autoconf-style checks to clean this up.

csfrancis:
Thanks for the quick review @sodabrew! Looking at CI, there are a bunch of compatibility issues between versions which I'm going to investigate. I'll ping once I'm ready for another look.

sodabrew:
Thank you! There's an older PR that was also going to make a vio call, and I'd been trying to avoid that, but you're making me rethink. See #807.

csfrancis:
Can you take another look @sodabrew? 

I'm not super happy with this, but I can't really come up with a better way. The problem is that `vio_is_connected` is not always exported (it depends on the libmysqlclient package being installed). What I don't like about this is that the behaviour of `closed?` is going to depend on whether or not the function is available. I added a private `_has_vio_is_connected?` method that can be used to determine if the function is available, but it still feels a bit gross to me.

Another option would be to duplicate the `st_vio` structure from `violite.h` in the gem, as the `is_connected` function available as a member on the structure. I don't like this as much because it's very brittle -- if the underlying `struct` changes there's no easy way to tell.

A couple of the CI test environments are failing, but the failures don't seem related to these changes.

kirs:
Thinking about `_has_vio_is_connected?`, should it be a class level method so you can do `Mysql2::Client._has_vio_is_connected?` before the connection is established? Alternatively could it be the part of `Mysql2::Client.info` hash?

csfrancis:
@kirs I think removing `_has_vio_is_connected?` as a private instance method and making it available from the class is a good idea. I prefer the approach of adding it to the `Mysql2::Client.info` hash, rather than polluting the class with a separate method. I really hate having to expose this implementation detail to the client, but I think it's better to be explicit and make the information available, rather than try to have the client infer it in some other way.

csfrancis:
@sodabrew Would you be able to take another look?

PR #1025, opened 2019-02-21
AWS Aurora Fast Failover support.

BupycHuk:


sodabrew:
Thanks! Apologies for the long delay to provide comments.

* It's important to me that all of the unit tests are running in CI and locally. I think this means mocking the Aurora behavior most of the time, based on a spec of how it behaves in the wild. I won't make the tests of this gem dependent on an AWS account.
* Given the proposed new class structure, there's good reason _not_ to be part of this gem. Rails won't know to use the different class name for the same gem name, so the way to structure this would be a new gem (`mysql2-aurora` for example) and then a new Rails ActiveRecord adapter that knows to use the new class. I figure most users asking for this feature are going to be using it for Rails, so this is a key consideration.
* I've been trying to think of another way to structure this. The thing that comes to mind is turning the `query` wrapper around. Rather than wrapping `query`, I'm thinking to pass in a rescue function, so `reconnect_with_readonly` passed in, and called when the query fails. Where I get stuck on this line of thinking is again with Rails, how to specify this behavior from the database config yaml.

sodabrew:
While researching #1294, I found that MariaDB's Connector/C has an Aurora failover plugin! https://mariadb.com/kb/en/mariadb-connectorc-plugins/

```
mysql_optionsv(mysql, MARIADB_OPT_CONNECTION_HANDLER, (void *)"aurora");
```

Because many distributions are always switching from MySQL 8 client libraries to MariaDB Connector/C client libraries, I think this is the best option to implement Aurora support going forward.

Time permitting, I will open a new PR that allows setting this option.

PR #1038, opened 2019-04-19
Debian 9 released with MariaDB as the only MySQL variant since 2017

qudongfang:
[Debian 9 released with MariaDB as the only MySQL variant](https://mariadb.org/debian-9-released-mariadb-mysql-variant/), and `apt-get install libmysqlclient-dev` doesn't work anymore.
We'd better update the document a little bit.

### DIFF
--- a/ext/mysql2/extconf.rb
+++ b/ext/mysql2/extconf.rb
@@ -7,7 +7,7 @@ def asplode(lib)
   elsif RUBY_PLATFORM =~ /darwin/
     abort "-----\n#{lib} is missing. You may need to 'brew install mysql' or 'port install mysql', and try again.\n-----"
   else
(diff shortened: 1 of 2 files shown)


ext/mysql2/extconf.rb
```diff
@@ -7,7 +7,7 @@ def asplode(lib)
   elsif RUBY_PLATFORM =~ /darwin/
     abort "-----\n#{lib} is missing. You may need to 'brew install mysql' or 'port install mysql', and try again.\n-----"
   else
-    abort "-----\n#{lib} is missing. You may need to 'apt-get install libmysqlclient-dev' or 'yum install mysql-devel', and try again.\n-----"
+    abort "-----\n#{lib} is missing. You may need to 'sudo apt-get install libmariadb-dev' or 'sudo yum install mysql-devel', and try again.\n-----"
```
sodabrew: Please keep libmysqlclient-dev as a suggestion here.

qudongfang: Thanks for the review, I have changed it.

PR #1040, opened 2019-04-27
Fix MySQL Encoding Tables

felixbuenemann:
Due to a bug in the generator script, missing MySQL encoding mappings caused empty strings instead of NULL mappings to be generated.

This PR fixes the mapping error and generates a WARNING to STDERR for missing mappings.

It also adds the two missing mappings that caused the problem and regenerates the encoding list based on the latest MySQL 8.0.16 community server on Ubuntu Xenial.

To avoid CI failures an update to the expired GPG key in `support/5072E1F5.asc` is also included and was copied from https://dev.mysql.com/doc/refman/8.0/en/checking-gpg-signature.html.


felixbuenemann:
I pushed another unrelated CI fix to use Xenial distribution for the `DB=mysql80` matrix, because it was actually installing MySQL 5.5 due to missing MySQL 8.0 packages for Trusty.

felixbuenemann:
CI errors are unrelated and already fixed by #1040.

sodabrew:
Thank you!

PR #1041, opened 2019-04-27
Fix fragile and crashing specs on macOS

felixbuenemann:
* Use monotonic time if possible to guard against clock skew
* Relax time comparisons to make test failures under load less likely
* Fix memory corruption segfaults on macOS due to mixing Threads and Timeout

Without the timeout thread safety fix, the specs always crash in the
"threaded queries should be supported" spec for various reasons on
macOS 10.14.4 with ruby 2.6.1 / MySQL 8.0.15.

Some observed errors:

    malloc: *** error for object 0x7fcbba90ef80: pointer being freed was not allocated
    malloc: *** set a breakpoint in malloc_error_break to debug

    lib/mysql2/client.rb:131:in `_query': Bad file descriptor (Errno::EBADF)

Now specs run fine, even during Prime95 load test on i9-8950HK (6c/12t).

### DIFF
--- a/spec/mysql2/client_spec.rb
+++ b/spec/mysql2/client_spec.rb
@@ -607,21 +607,21 @@ RSpec.describe Mysql2::Client do
       # XXX this test is not deterministic (because Unix signal handling is not)
       # and may fail on a loaded system
       it "should run signal handlers while waiting for a response" do
-        kill_time = 0.1
-        query_time = 2 * kill_time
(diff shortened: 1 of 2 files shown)


sodabrew:
I think you meant comment https://github.com/brianmario/mysql2/pull/1040#issuecomment-487253755 on this PR - merging in PR order.

PR #1044, opened 2019-05-15
Remove test files from bundled gem

ikottman:
RubyGems has [deprecated](https://github.com/rubygems/guides/issues/90) the `test_files` option, so it is unnecessary to include spec files in the gem anymore.


ikottman:
Thanks @sodabrew!

PR #1046, opened 2019-05-17
Improve performance to call Mysql2::Result#each and Mysql2::Result#fields

Watson1978:
This PR will reduce the time to call Mysql2::Result#each and Mysql2::Result#fields.

If table records are small enough, the performance will improve as following.

```
Mysql2::Result#each : 901.633k -> 1.550M (1.71x faster)
Mysql2::Result#fields : 3.779M -> 6.233M (1.65x faster)
```

1. Use rb_ivar_get/rb_ivar_set to improve performance
  * By convert C-string to Ruby string ID each time, it will take a slightly time to call method each time.
This patch will catch the Ruby string ID to cut off the time.
2. Check only whether block was given
  * `rb_scan_args(argc, argv, "01&", ...)` will generate `Proc` object from block. However, the object has used to only check whether block was given. To remove redundant object generating, this patch will use `rb_block_given_p()` to check whether block was given.

### Environment
- CPU : Intel(R) Core(TM) i5-3210M CPU @ 2.50GHz
- OS : Ubuntu 18.10
- Compiler : gcc version 8.3.0

### Before changing
```
Warming up --------------------------------------
               query   886.000  i/100ms
                each    67.339k i/100ms
              fields   195.612k i/100ms
Calculating -------------------------------------
               query      9.385k (± 4.5%) i/s -     46.958k in   5.016539s
                each    901.633k (± 0.2%) i/s -      4.512M in   5.003951s
              fields      3.779M (± 0.2%) i/s -     18.974M in   5.021533s
```

### After changing
```
Warming up --------------------------------------
               query   864.000  i/100ms
                each   106.916k i/100ms
              fields   251.255k i/100ms
Calculating -------------------------------------
               query      9.457k (± 3.8%) i/s -     47.520k in   5.032949s
                each      1.550M (± 0.3%) i/s -      7.805M in   5.037029s
              fields      6.233M (± 0.1%) i/s -     31.407M in   5.039049s
```

### Benchmarks
```ruby
require 'benchmark/ips'
require 'mysql2'

def setup_database(client)
  # client.query('create database benchmark')
  sql = <<SQL
create table benchmark.users (
  id int NOT NULL AUTO_INCREMENT PRIMARY KEY,
  name text,
  email text,
  created_at datetime,
  updated_at datetime
)
SQL
  client.query(sql)

  5.times do |i|
    client.query("insert into benchmark.users (name, email, created_at, updated_at) values ('foo', 'foo@example.com', '#{Time.now.strftime("%Y-%m-%d %H:%M:%S")}', '#{Time.now.strftime("%Y-%m-%d %H:%M:%S")}')")
  end
end


client = Mysql2::Client.new(
  :host     => "127.0.0.1",
  :username => "root",
  :password => "",
  :database => "benchmark"
)

# setup_database(client)

Benchmark.ips do |x|
  query = 'select * from users'.freeze
  x.report "query" do
    client.query(query)
  end

  results = client.query(query)
  x.report "each" do
    results.each {}
  end

  x.report "fields" do
    results.fields
  end
end
```

### DIFF
--- a/ext/mysql2/result.c
+++ b/ext/mysql2/result.c
@@ -30,7 +30,7 @@ typedef struct {
   int streaming;
   ID db_timezone;
   ID app_timezone;
-  VALUE block_given;
+  int block_given; /* boolean */
 } result_each_args;
 
 extern VALUE mMysql2, cMysql2Client, cMysql2Error;
@@ -741,7 +741,7 @@ static VALUE rb_mysql_result_each_(VALUE self,
         row = fetch_row_func(self, fields, args);
         if (row != Qnil) {
           wrapper->numberOfRows++;
-          if (args->block_given != Qnil) {
+          if (args->block_given) {
             rb_yield(row);
           }
         }
@@ -791,7 +791,7 @@ static VALUE rb_mysql_result_each_(VALUE self,
           return Qnil;
         }
 
-        if (args->block_given != Qnil) {
+        if (args->block_given) {
           rb_yield(row);
         }
       }
@@ -809,7 +809,7 @@ static VALUE rb_mysql_result_each_(VALUE self,
 
 static VALUE rb_mysql_result_each(int argc, VALUE * argv, VALUE self) {
   result_each_args args;
-  VALUE defaults, opts, block, (*fetch_row_func)(VALUE, MYSQL_FIELD *fields, const result_each_args *args);
+  VALUE defaults, opts, (*fetch_row_func)(VALUE, MYSQL_FIELD *fields, const result_each_args *args);
   ID db_timezone, app_timezone, dbTz, appTz;
   int symbolizeKeys, asArray, castBool, cacheRows, cast;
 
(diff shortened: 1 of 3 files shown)


ext/mysql2/result.c
```diff
@@ -821,12 +821,17 @@ static VALUE rb_mysql_result_each(int argc, VALUE * argv, VALUE self) {
...
+  if (rb_scan_args(argc, argv, "01", &opts) == 1) {
     opts = rb_funcall(defaults, intern_merge, 1, opts);
   } else {
     opts = defaults;
   }
 
+  block = Qfalse;
+  if (rb_block_given_p()) {
+    block = Qtrue;
+  }
```
sodabrew: Great explanation that actually capturing the block will generate a Proc object that is never called!

To really clarify this code, I suggest a deeper change:

1. Add a comment above `rb_scan_args` that says something like "A block can be passed to this method, but since we don't call the block directly from C, we don't need to capture it into a variable here with the "&" scan arg"

2. Remove the `block` variable entirely.

3. Replace `args.block_given = block;` on line 890 with `args.block_given = rb_block_given_p()`

(Adjust the data type of `args.block_given` to match the return value of `rb_block_given_p()` if needed)

sodabrew:
This is great work, thank you! One comment in the review.

ext/mysql2/result.c
Watson1978: @sodabrew Thank you for your review!! I fixed a patch with your idea at https://github.com/brianmario/mysql2/pull/1046/commits/f7b0db4574b2c3b8a92aa298da52a8209c013d8d

Issue #584, opened 2015-02-20
undefined method chars for ["P"]:Array - mysql2 0.3.18, Win7 x64, x64_mingw32


ABaumgaertner:
simple test fails: mysql.rb 
gem install mysql2 --> mysql2-0.3.18-x64-mingw32

    more mysql.rb
    require 'mysql2'

    ruby mysql.rb
    C:/apps/redmine_dev/ruby/lib/ruby/2.2.0/Win32API.rb:18:in `initialize': undefined method `chars' for ["P"]:Array (NoMethodError)
        from C:/apps/redmine_dev/ruby/lib/ruby/gems/2.2.0/gems/mysql2-0.3.18-x64-mingw32/lib/mysql2.rb:23:in `new'
        from C:/apps/redmine_dev/ruby/lib/ruby/gems/2.2.0/gems/mysql2-0.3.18-x64-mingw32/lib/mysql2.rb:23:in `<top (required)>'
        from C:/apps/redmine_dev/ruby/lib/ruby/2.2.0/rubygems/core_ext/kernel_require.rb:128:in `require'
        from C:/apps/redmine_dev/ruby/lib/ruby/2.2.0/rubygems/core_ext/kernel_require.rb:128:in `rescue in require'
        from C:/apps/redmine_dev/ruby/lib/ruby/2.2.0/rubygems/core_ext/kernel_require.rb:39:in `require'
        from mysql.rb:1:in `<main>'

    RubyGems Environment:
      - RUBYGEMS VERSION: 2.4.5
      - RUBY VERSION: 2.2.0 (2015-02-20 patchlevel 69) [x64-mingw32]
      - INSTALLATION DIRECTORY: C:/apps/redmine_dev/ruby/lib/ruby/gems/2.2.0
      - RUBY EXECUTABLE: C:/apps/redmine_dev/ruby/bin/ruby.exe
      - EXECUTABLE DIRECTORY: C:/apps/redmine_dev/ruby/bin
      - SPEC CACHE DIRECTORY: C:/Users/abaumgaertner/.gem/specs
      - SYSTEM CONFIGURATION DIRECTORY: C:/ProgramData
      - RUBYGEMS PLATFORMS:
        - ruby
        - x64-mingw32
      - GEM PATHS:
         - C:/apps/redmine_dev/ruby/lib/ruby/gems/2.2.0
         - C:/Users/abaumgaertner/.gem/ruby/2.2.0
      - GEM CONFIGURATION:
[...]

PR #1053, opened 2019-06-15
Replace Win32API with Fiddle, update appveyor.yml

MSP-Greg:
My fork's Appveyor build passed and is at:
https://ci.appveyor.com/project/MSP-Greg/mysql2/builds/25306388

### DIFF
--- a/lib/mysql2.rb
+++ b/lib/mysql2.rb
@@ -20,9 +20,12 @@ if RUBY_PLATFORM =~ /mswin|mingw/
   end
 
   if dll_path
-    require 'Win32API'
-    LoadLibrary = Win32API.new('Kernel32', 'LoadLibrary', ['P'], 'I')
-    if LoadLibrary.call(dll_path).zero?
+    require 'fiddle'
+    kernel32 = Fiddle.dlopen 'kernel32'
+    load_library = Fiddle::Function.new(
+      kernel32['LoadLibraryW'], [Fiddle::TYPE_VOIDP], Fiddle::TYPE_INT,
+    )
+    if load_library.call(dll_path.encode('utf-16le')).zero?
       abort "Failed to load libmysql.dll from #{dll_path}"
     end
   end
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -25,6 +25,10 @@ on_failure:
   - find tmp -name "*.log" | xargs cat
 environment:
   matrix:
+    - ruby_version: "26-x64"
+      MINGW_PACKAGE_PREFIX: "mingw-w64-x86_64"
+    - ruby_version: "25-x64"
+      MINGW_PACKAGE_PREFIX: "mingw-w64-x86_64"
     - ruby_version: "24-x64"
       MINGW_PACKAGE_PREFIX: "mingw-w64-x86_64"
     - ruby_version: "24"


sodabrew:
I remember doing it this way because Win32API is in core. Is it deprecated/removed and is Fiddle bundled with core now? At which Ruby does this change? (In other words what revisions are being left behind?)

MSP-Greg:
> I remember doing it this way because Win32API is in core

I think it's always been part of the std-lib.

>Is it deprecated/removed

See [`ext/win32/lib/Win32API.rb:5`](https://github.com/ruby/ruby/blob/11f8c89171ff0b260891dec064cf69d9400a0704/ext/win32/lib/Win32API.rb#L5).  I haven't gone all the way back, but also see https://github.com/ruby/ruby/commit/dd3aec25d63e865ce0e90b2ff870d7bcc39b1a56.

Then again, I've joked about it, given that it's been deprecated for a very long time.

Win32API uses Fiddle.  Really, I'm just tired of seeing warning messages...

lib/mysql2.rb
```diff
@@ -20,9 +20,12 @@
   end
 
   if dll_path
-    require 'Win32API'
-    LoadLibrary = Win32API.new('Kernel32', 'LoadLibrary', ['P'], 'I')
-    if LoadLibrary.call(dll_path).zero?
+    require 'fiddle'
+    kernel32 = Fiddle.dlopen 'kernel32'
```
sodabrew: I think this should still be initial cap `Kernel32` ?

sodabrew:
Looks like Fiddle is core as far back as Ruby 2.0: https://ruby-doc.org/stdlib-2.0.0/libdoc/fiddle/rdoc/Fiddle.html

Meanwhile, I knew there was some history in my memory banks around Win32API: https://github.com/ruby/ruby/pull/835

See also https://github.com/brianmario/mysql2/issues/584

Since the issue came up in the mysql2 0.3.x series, continuing to use `Win32API` must have been to continue supporting Ruby 1.8/1.9 (without Fiddle in core). Since those versions are no longer supported I see no reason not to switch.

_I am reminded to get a release out..._

lib/mysql2.rb
MSP-Greg: Umm, not reusing it are we?

MSP-Greg:
Note to self - ALWAYS USE 'W' VERSIONS...

Some users may have UTF-8 characters in their user names...  Updated.  [Passing Appveyor build](https://ci.appveyor.com/project/MSP-Greg/mysql2/builds/25343815)

lib/mysql2.rb
sodabrew: I mean the library:`Win32API.new('Kernel32', ...`
vs. `Fiddle.dlopen 'kernel32'`

MSP-Greg: Sorry. between multi-tasking & bouncing between languages, I often miss the obvious.

Actually, the file is `kernel32.dll`, and as you know, Windows file system is case-insensitive (which can cause issues when comparing paths, use `casecmp`...)

Also, ruby/reline essentially created its own `Win32API` class to avoid the deprecation warnings, and it's all lower case.  See [lib/reline/windows.rb
](https://github.com/ruby/ruby/blob/9b18cb341f08c725d74f667ef3724c4cc852fdbd/lib/reline/windows.rb#L55-L62)

Long story short, I think we're ok...

PR #1060, opened 2019-08-08
Remove test_files (i.e. specs, examples) from gem

nbeyer:
I'd like to propose removing the RSpecs and Examples from the packaged gem that's distributed. There doesn't seem to be any practical use of these files within a release version of the gem file. Another set of "test" files, benchmarks, is already not included. The benefit of removing these files would be a reduced gem file size as well as eliminating false positive vulnerabilities from security scanners (the test files include private certificates).


sodabrew:
Resolved by #1044

Issue #1067, opened 2019-09-08
Feature Request: get field types from query result


danhuynhdev:
This feature would be very helpful for me. And since mysql c-api does support this capability, this feature should be relatively straight forward to implement.

PR #1068, opened 2019-09-08
Implementing type reflection from mysql result

danhuynhdev:
Add `Mysql2::Result#field_types` method to get field types from result.

Resolve: #1067

### DIFF
--- a/ext/mysql2/result.c
+++ b/ext/mysql2/result.c
@@ -17,6 +17,20 @@ static rb_encoding *binaryEncoding;
  */
 #define MYSQL2_MIN_TIME 2678400ULL
 
+#define MYSQL2_MAX_BYTES_PER_CHAR 3
+
+/* From Mysql documentations:
+ *   To distinguish between binary and nonbinary data for string data types,
+ *   check whether the charsetnr value is 63. If so, the character set is binary,
+ *   which indicates binary rather than nonbinary data. This enables you to distinguish BINARY
+ *   from CHAR, VARBINARY from VARCHAR, and the BLOB types from the TEXT types.
+ */
+#define MYSQL2_BINARY_CHARSET 63
+
+#ifndef MYSQL_TYPE_JSON
+#define MYSQL_TYPE_JSON 245
+#endif
+
 #define GET_RESULT(self) \
   mysql2_result_wrapper *wrapper; \
   Data_Get_Struct(self, mysql2_result_wrapper, wrapper);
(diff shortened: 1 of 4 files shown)


ext/mysql2/result.c
```diff
@@ -169,6 +169,160 @@ static VALUE rb_mysql_result_fetch_field(VALUE self, unsigned int idx, int symbo
...
+static VALUE rb_mysql_result_fetch_field_type(VALUE self, unsigned int idx) {
+  VALUE rb_field_type;
+  GET_RESULT(self);
+
+  if (wrapper->fieldTypes == Qnil) {
+    wrapper->numberOfFields = mysql_num_fields(wrapper->result);
+    wrapper->fieldTypes = rb_ary_new2(wrapper->numberOfFields);
+  }
+
+  unsigned long mb_per_char = 3;
```
danhuynhdev: This is fixed for now. Could be doing something like https://github.com/mysql/mysql-connector-j/blob/16a712ddb3f826a1933ab42b0039f7fb9eebc6ec/src/main/core-api/java/com/mysql/cj/CharsetMapping.java for correctness.

ext/mysql2/result.c
```diff
@@ -169,6 +169,160 @@ static VALUE rb_mysql_result_fetch_field(VALUE self, unsigned int idx, int symbo
...
+      case MYSQL_TYPE_NEWDECIMAL:   // char[]
+        rb_field_type = rb_sprintf("decimal(%ld,%d)", field->length - 2, field->decimals);
+        break;
+      case MYSQL_TYPE_STRING:       // char[]
+        if (field->flags & ENUM_FLAG) {
+          rb_field_type = rb_str_new_cstr("enum");
+        } else if (field->flags & SET_FLAG) {
+          rb_field_type = rb_str_new_cstr("set");
+        } else {
+          if (field->charsetnr == 63) {
```
danhuynhdev: From the mysql documentation. Will change this to a constant later.
> To distinguish between binary and nonbinary data for string data types, check whether the charsetnr value is 63. If so, the character set is binary, which indicates binary rather than nonbinary data. This enables you to distinguish BINARY from CHAR, VARBINARY from VARCHAR, and the BLOB types from the TEXT types.

ext/mysql2/result.c
```diff
@@ -169,6 +169,160 @@ static VALUE rb_mysql_result_fetch_field(VALUE self, unsigned int idx, int symbo
...
+        break;
+      case MYSQL_TYPE_DATETIME:     // MYSQL_TIME
+        rb_field_type = rb_str_new_cstr("datetime");
+        break;
+      case MYSQL_TYPE_TIMESTAMP:    // MYSQL_TIME
+        rb_field_type = rb_str_new_cstr("timestamp");
+        break;
+      case MYSQL_TYPE_DECIMAL:      // char[]
+      case MYSQL_TYPE_NEWDECIMAL:   // char[]
+        rb_field_type = rb_sprintf("decimal(%ld,%d)", field->length - 2, field->decimals);
```
danhuynhdev: Not sure why the length has 2 extra bytes. I couldn't find any documentation about this. Appreciate any help here :thinking:

ext/mysql2/result.c
sodabrew: Please either add this comment into the code here, or break out `63` as a constant and put the comment next to it.

ext/mysql2/result.c
sodabrew: Please make this a constant, since there's no other values for mb_per_char at this time.

sodabrew:
This looks good! Few smaller comments, and please rebase onto current master. (It appears there are no conflicts, but we'll get confirmation of unit tests.)

danhuynhdev:
I've resolved the comments and squash the commits. Please feedback if any change is needed. cc: @sodabrew

ext/mysql2/result.c
```diff
@@ -2,7 +2,6 @@
 
 #include "mysql_enc_to_ruby.h"
 #define MYSQL2_CHARSETNR_SIZE (sizeof(mysql2_mysql_enc_to_rb)/sizeof(mysql2_mysql_enc_to_rb[0]))
-
```
simi: :scissors: ?

ext/mysql2/result.c
```diff
@@ -169,9 +178,161 @@ static VALUE rb_mysql_result_fetch_field(VALUE self, unsigned int idx, int symbo
...
+      case MYSQL_TYPE_SET:          // char[]
+        rb_field_type = rb_str_new_cstr("set");
+        break;
+      case MYSQL_TYPE_ENUM:         // char[]
+        rb_field_type = rb_str_new_cstr("enum");
+        break;
+      case MYSQL_TYPE_GEOMETRY:     // char[]
+        rb_field_type = rb_str_new_cstr("geometry");
+        break;
+      default:
```
simi: If I understand it well it will return `nil` for unknown type, right?

ext/mysql2/result.c
sodabrew: Please note this oddity in a comment.

ext/mysql2/result.c
sodabrew: Yes, please avoid unrelated whitespace changes outside the area of code being modified.

ext/mysql2/result.c
```diff
@@ -182,7 +343,7 @@ static VALUE mysql2_set_field_string_encoding(VALUE val, MYSQL_FIELD field, rb_e
     int enc_index;
 
     enc_name = (field.charsetnr-1 < MYSQL2_CHARSETNR_SIZE) ? mysql2_mysql_enc_to_rb[field.charsetnr-1] : NULL;
-    
+
```
sodabrew: Nit: unrelated whitespace

simi:
:thinking: I was looking for any chance to keep the list of types updated when new type would be introduced (by exposing `enum_field_types` and ensuring we handle all supported types), but looking into `enum_field_types` history, it is rarely touched and seems pretty stable.

ext/mysql2/result.c
sodabrew: I wonder how we can force a unit test down the no-match path?

ext/mysql2/result.c
danhuynhdev: Mysql do the same in their code I've add a comment to that section
https://github.com/mysql/mysql-server/blob/ea7d2e2d16ac03afdd9cb72a972a95981107bf51/sql/field.cc#L2246

ext/mysql2/result.c
danhuynhdev: If we know a unknown to test for it, would it be unknown anymore :laughing:?

danhuynhdev: Anyway I've made it return 'unknown' for this case.

simi: I was looking into this @sodabrew and I have found out there's no easy way how to introduce custom column type in MySQL :/

But maybe there's some way how to create table including all MySQL data types (by some meta programming) and it would be possible to check if we cover all cases there.

danhuynhdev:
Seems like `mysql2` cannot handle JSON type in query result. I've attempted to add JSON type to the list but it just failed right away.

ext/mysql2/result.c
danhuynhdev: It will fail if we include `MYSQL_TYPE_JSON` though. I cannot find any code to handle `MYSQL_TYPE_JSON`.

sodabrew: See #693 for the history on how JSON is supported; basically it's a fall-through to `char[]`.

It seems to me that the type table should be updated to explicitly include JSON so that it's not "accidentally" supported but is explictly routed through the `char[]` path of the result handler.

What is the failure you're seeing if `MYSQL_TYPE_JSON` is included in this table?

sodabrew:
The JSON type should be supported, what error are you seeing?

danhuynhdev:
This is actually not an error on our side but on mysql's json support. I'm using mariadb on my local machine and they don't support `JSON` type. Also `MYSQL_TYPE_JSON` only exist on mysql 8.0 onward. Attempt to add it into the table will result in this error:
```
../../../../ext/mysql2/result.c:320:12: error: ‘MYSQL_TYPE_JSON’ undeclared (first use in this function)
       case MYSQL_TYPE_JSON:     // char[]
```

danhuynhdev:
Not sure if we should add something like this for `MYSQL_TYPE_JSON`
```
#ifdef MYSQL_TYPE_JSON
      case MYSQL_TYPE_JSON:     // json
        rb_field_type = rb_str_new_cstr("json");
        break;
#endif
```
or just declare it if not exist
```
#ifnotdef MYSQL_TYPE_JSON
#define MYSQL_TYPE_JSON 245
#endif 
```
I'm inclined to do latter since it doesn't require user to install the correct header version on the client side.

sodabrew:
I was going to say we should do the first one, but you're right that some clients may have been compiled with an older library, and the column type number is should always be the same for compatibility reasons.

PR #1083, opened 2019-11-21
The taint mechanism will be deprecated in Ruby 2.7

kamipo:
The Ruby core team decided to deprecate the taint mechanism in Ruby 2.7
and will remove that in Ruby 3.

https://bugs.ruby-lang.org/issues/16131
https://github.com/ruby/ruby/pull/2476

In Ruby 2.7, `Object#{taint,untaint,trust,untrust}` and related
functions in the C-API no longer have an effect (all objects are always
considered untainted), and are now warned deprecation message.

### DIFF
--- a/ext/mysql2/client.c
+++ b/ext/mysql2/client.c
@@ -171,7 +171,7 @@ static void rb_mysql_client_mark(void * wrapper) {
 
 static VALUE rb_raise_mysql2_error(mysql_client_wrapper *wrapper) {
   VALUE rb_error_msg = rb_str_new2(mysql_error(wrapper->client));
-  VALUE rb_sql_state = rb_tainted_str_new2(mysql_sqlstate(wrapper->client));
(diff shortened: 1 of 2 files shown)


sodabrew:
I'm not sure yet I understand the impact for older Rubies to use the non-taint API calls; it sounds like most code in practice is not using taint hints to protect against untrusted inputs, and so in practice there won't be any impact to "most" real user code?

But I wonder if it wouldn't make sense to wrap these in #ifdef until the older Rubies are no longer supported by this gem.

kamipo:
IMO if we marked `error.message` as tainted, I think there is room to carefully consider removing the tainted flag, but `error.message` is always marked as trusted string, even though it may have an user input.

OTOH `error.sql_state` is just an error code, it is obviously safe than `error.message`, but it is not marked as safe (trusted) string unlike `error.message`.

I believe the inconsistency is not intended.

So I'd say yes for the answer "it sounds like most code in practice is not using taint hints to protect against untrusted inputs, and so in practice there won't be any impact to "most" real user code?".

PR #1092, opened 2019-12-11
Client session tracking

insom:
This adds support for Oracle/Percona MySQL 5.7+ session tracking to the client.

Docs on the function: https://dev.mysql.com/doc/refman/5.7/en/mysql-session-track-get-first.html

At Shopify, we wanted to be able to tell which GTID was generated by a given `COMMIT` so that when we are reading from delayed replicas we can be sure that a given replica has the data that we want on it. This is pretty much the exact flow that's described by the [Oracle worklog entry describing the implementation of the feature](https://dev.mysql.com/worklog/task/?id=6128)

Example usage:

```ruby
require 'mysql2'
c = Mysql2::Client.new(:host => "127.0.0.1", :username => "root", :password => "test1", :flags => "SESSION_TRACK", :database => "test");
c.query("SET SESSION session_track_gtids=OWN_GTID")
c.query("INSERT INTO test VALUES (1)");
gtid = c.session_track(Mysql2::Client::SESSION_TRACK_GTIDS)[0]

c2 = Mysql2::Client.new(:host => "replica.host")
c2.query("SELECT WAIT_FOR_EXECUTED_GTID_SET('#{gtid}')")
# waits forever, by default, probably not what people want
c2.query("SELECT * FROM test")
```

I mirrored the implementation of `client.last_id` -- I noticed that wasn't documented in the `README.md` so I hadn't added this, but if you have any suggestions about how or where to document this feature I'd appreciate that.

I believe that all the `ifdef`ing is right to not compile this feature for MariaDB (which doesn't support it).

### DIFF
--- a/README.md
+++ b/README.md
@@ -197,6 +197,23 @@ statement = @client.prepare("SELECT * FROM users WHERE last_login >= ? AND locat
 result = statement.execute(1, "CA", :as => :array)
 ```
 
+Session Tracking information can be accessed with
+
+```ruby
+c = Mysql2::Client.new(
+  host: "127.0.0.1",
+  username: "root",
+  flags: "SESSION_TRACK",
+  init_command: "SET @@SESSION.session_track_schema=ON"
+)
+c.query("INSERT INTO test VALUES (1)")
+session_track_type = Mysql2::Client::SESSION_TRACK_SCHEMA
+session_track_data = c.session_track(session_track_type)
+```
+
+The types of session track types can be found at
+[https://dev.mysql.com/doc/refman/5.7/en/mysql-session-track-get-first.html](https://dev.mysql.com/doc/refman/5.7/en/mysql-session-track-get-first.html)
+
 ## Connection options
 
 You may set the following connection options in Mysql2::Client.new(...):
--- a/ext/mysql2/client.c
+++ b/ext/mysql2/client.c
@@ -1642,6 +1642,13 @@ void init_mysql2_client() {
 
 #ifdef CLIENT_SESSION_TRACK
   rb_const_set(cMysql2Client, rb_intern("SESSION_TRACK"), INT2NUM(CLIENT_SESSION_TRACK));
+  /* From mysql_com.h -- but stable from at least 5.7.4 through 8.0.20 */
+  rb_const_set(cMysql2Client, rb_intern("SESSION_TRACK_SYSTEM_VARIABLES"), INT2NUM(SESSION_TRACK_SYSTEM_VARIABLES));
+  rb_const_set(cMysql2Client, rb_intern("SESSION_TRACK_SCHEMA"), INT2NUM(SESSION_TRACK_SCHEMA));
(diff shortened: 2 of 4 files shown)


sodabrew:
Thank you!

Few things needed:
- Unit tests (which should skip if MySQL < 5.7)
- README updates to add a section about how to use this feature
- Example code should use the init_command feature because it is works with `reconnect` to ensure the initial SQL is executed even if the connection drops and is transparently reconnected.

```
require 'mysql2'

c = Mysql2::Client.new(
    host: "127.0.0.1",
    username: "root",
    password: "test1",
    database: "test",
    flags: "SESSION_TRACK",
    init_command: "SET SESSION session_track_gtids=OWN_GTID")
c.query("INSERT INTO test VALUES (1)")
gtid = c.session_track(SESSION_TRACK_GTIDS).first
```

ext/mysql2/client.c
```diff
@@ -1613,6 +1641,21 @@ void init_mysql2_client() {
       INT2NUM(0));
 #endif
 
+#ifdef CLIENT_SESSION_TRACK
+  rb_define_method(cMysql2Client, "session_track", rb_mysql_client_session_track, 1);
```
sodabrew: The pattern I'd like to follow is to always define this method on the object, and within the method, if the feature is unavailable, to return Qnil.

bibstha:
@sodabrew I've made the following changes
1. Always defined `session_track` method and conditionally returns nil.
2. Updated readme with information on how to call `session_track` method.

bibstha:
Also seems like the failing test is due to network failure. I'm unable to rerun the failing build, hum.

bibstha:
Hi @sodabrew, do you have some time to review this PR?

README.md
```diff
@@ -197,6 +197,23 @@ statement = @client.prepare("SELECT * FROM users WHERE last_login >= ? AND locat
 result = statement.execute(1, "CA", :as => :array)
 ```
 
+Session Tracking information can be accessed with
+
+```ruby
+c = Mysql2::Client.new(
+  host: "127.0.0.1",
+  username: "root",
+  flags: "SESSION_TRACK",
```
kirs: I think this example might be wrong? Should it be `flags: Mysql2::Client::SESSION_TRACK`?

insom: You can pass all sorts of things to `flags` 😁 -- https://github.com/brianmario/mysql2/blob/master/lib/mysql2/client.rb#L56L57 -- so this will work. It might be better to encourage the use of the constant for style reasons though? The other examples do.

jeremy:
This is a great feature. Thank you for contributing it.

spec/mysql2/client_spec.rb
```diff
@@ -1,6 +1,6 @@
 require 'spec_helper'
 
-RSpec.describe Mysql2::Client do
+RSpec.describe Mysql2::Client do # rubocop:disable Metrics/BlockLength
```
simi: Is there any chance to disable/extend this rubocop rule in specs only? rspec is usually bunch of huge blocks by design.

sodabrew:
Thanks again for this the support for this feature!

PR #1097, opened 2019-12-19
Fixed segfault by `Mysql2::Client.allocate`

nobu:

PR #1101, opened 2020-01-04
Fix crash if a Mysql2::Client object is allocated but never connected

sodabrew:
Reproducible with any older version of the mysql2 gem:

        ruby -r mysql2 -I lib -e 'Mysql2::Client.allocate.close'

Before this fix, the line above would crash out with a memory error:

        malloc: *** error for object 0x...: pointer being freed was not allocated
        malloc: *** set a breakpoint in malloc_error_break to debug
        Abort trap: 6

Alternative to #1097


sodabrew:
@nobu I want to check with you if this resolves the same crash at #1097, or if you saw a different crash (and how to reproduce it?)

nobu:
Thank you, I confirmed that it worked fine.

PR #1112, opened 2020-03-09
Avoid a hash object allocation per each `query`/`execute` call

kamipo:
This is a same optimization approach with https://github.com/ruby/ruby/pull/2393.


sodabrew:
Thank you!

PR #1115, opened 2020-04-01
Register C global variables to Ruby GC to avoid problems with GC.compact

casperisfine:
Since Ruby 2.7, C global variables holding references to Ruby objects must be declared to the GC even if they are also held from the Ruby side, otherwise a call to GC.compact might move them to another location.

For instance here's an exception we've seen caused by this:

```
bundler/gems/mysql2-d3c815c68f22/lib/mysql2/client.rb:131:in `_query': NoMethodError: undefined method `new_with_args' for "2":String (ActiveRecord::StatementInvalid)
    from bundler/gems/mysql2-d3c815c68f22/lib/mysql2/client.rb:131:in `block in query'
    from bundler/gems/mysql2-d3c815c68f22/lib/mysql2/client.rb:130:in `handle_interrupt'
    from bundler/gems/mysql2-d3c815c68f22/lib/mysql2/client.rb:130:in `query'
```

In this instance `Mysql2::Error` was moved to another location, and `cMysql2Error` was now pointing to an instance of `"2"`.

We had a similar issue in https://github.com/Shopify/liquid-c/pull/55, however here I tried to reproduce it in the test suite by repeatedly calling `GC.compact`, but without any success. Maybe the mysql2 test suite is not producing enough GC objects to demonstrate the bug.

cc @csfrancis @XrXr @rafaelfranca

### DIFF
--- a/ext/mysql2/client.c
+++ b/ext/mysql2/client.c
@@ -1453,6 +1453,7 @@ void init_mysql2_client() {
   mMysql2      = rb_define_module("Mysql2"); Teach RDoc about Mysql2 constant.
 #endif
   cMysql2Client = rb_define_class_under(mMysql2, "Client", rb_cObject);
+  rb_global_variable(&cMysql2Client);
 
   rb_define_alloc_func(cMysql2Client, allocate);
 
--- a/ext/mysql2/mysql2_ext.c
+++ b/ext/mysql2/mysql2_ext.c
@@ -4,9 +4,14 @@ VALUE mMysql2, cMysql2Error, cMysql2TimeoutError;
 
 /* Ruby Extension initializer */
 void Init_mysql2() {
-  mMysql2      = rb_define_module("Mysql2");
+  mMysql2 = rb_define_module("Mysql2");
+  rb_global_variable(&mMysql2);
+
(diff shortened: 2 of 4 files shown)


sodabrew:
Thank you!

casperisfine:
Welcome. Let me know if there's anything needed to get this merged.

I tried some more tricks to reproduce the bug (generate a lot of garbage before loding the extension & then compact), but still no luck.

I wish there was a way to force the GC to move a particular object, that would be very useful to test this kind of stuff.

sodabrew:
Do all of the static variables need to be flagged against movement as well? For example
```
  intern_usec = rb_intern("usec");
```

casperisfine:
@sodabrew I'll let @XrXr correct me if I say something wrong.

But I didn't mark these symbols, because AFAIK they aren't true objects. Meaning there's nothing more than the content of the `VALUE`. There's no associated object slot that can be moved.

Same for Integer and Float etc. But other than that, yes all static `VALUE` should be marked.

sodabrew:
I'm not terribly familiar with these changes in Ruby 2.7, so I'll take your testing-word for it. If the other variables should also be marked that's easy enough to add in a subsequent PR!

casperisfine:
Thanks!

PR #1126, opened 2020-06-27
0.2.x update for ruby 2.5

pavelvasev:
Mysql2 0.2.24 does not compile in ruby 2.5 due to extra check this ruby performs on rb_funcall.
In particular, file result.c had a bug - wrong number (6) of arguments on line 300.
* This pull request fixes number of arguments to valid value of 7.
* Also, to publish these change as a gem, a gem version is updated to 0.2.25.

I ask to merge this changeset in master repository and also ask to build and publish gem version 0.2.25.

### DIFF
--- a/ext/mysql2/result.c
+++ b/ext/mysql2/result.c
@@ -297,7 +297,7 @@ static VALUE rb_mysql_result_fetch_row(VALUE self, ID db_timezone, ID app_timezo
             break;
           }
           msec = msec_char_to_uint(msec_char, sizeof(msec_char));
-          val = rb_funcall(rb_cTime, db_timezone, 6, opt_time_year, opt_time_month, opt_time_month, UINT2NUM(hour), UINT2NUM(min), UINT2NUM(sec), UINT2NUM(msec));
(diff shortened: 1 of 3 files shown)


pavelvasev:
Anybody alive here?..

sodabrew:
Can you use a newer version of the gem to that supports current versions of Ruby?

The mysql2 0.2 series is rather out of date.

pavelvasev:
> Can you use a newer version of the gem to that supports current versions of Ruby?
> 
> The mysql2 0.2 series is rather out of date.

Thank you for your reply! I support Rails 2.3 application. We try to switch to Rails 2.3 LTS + Ruby 2.5, and it forces us to use mysql2. And as I understand, 0.2.x is the only branch compatible with ActiveRecord from Rails 2.3..

sodabrew:
What is Rails 2.3 LTS?

pavelvasev:
> What is Rails 2.3 LTS?

https://railslts.com/

pavelvasev:
Here is their recommendation to use mysql2. https://makandracards.com/railslts/47192-ruby-2-5-support-for-rails-2-3-lts
Actually they motivate it because mysql gem does not compile in ruby 2.5 at all.

Issue #950, opened 2018-03-22
"Specified 'mysql2' for database adapter, but the gem is not loaded" when upgrading to 0.5.0


roberts1000:
I just tried upgrading from 0.4.10 to 0.5.0 in a Rails app and got the following error when starting the server in development:

```bash
=> Booting Puma
=> Rails 5.1.5 application starting in development 
=> Run `rails server -h` for more startup options
Exiting
Traceback (most recent call last):
   ....
   5: from /home/dev/.rvm/gems/ruby-2.5.0/gems/activesupport-5.1.5/lib/active_support/lazy_load_hooks.rb:69:in `instance_eval'
   4: from /home/dev/.rvm/gems/ruby-2.5.0/gems/activerecord-5.1.5/lib/active_record/railtie.rb:124:in `block (2 levels) in <class:Railtie>'
   3: from /home/dev/.rvm/gems/ruby-2.5.0/gems/activerecord-5.1.5/lib/active_record/connection_handling.rb:58:in `establish_connection'
   2: from /home/dev/.rvm/gems/ruby-2.5.0/gems/activerecord-5.1.5/lib/active_record/connection_adapters/abstract/connection_pool.rb:880:in `establish_connection'
   1: from /home/dev/.rvm/gems/ruby-2.5.0/gems/activerecord-5.1.5/lib/active_record/connection_adapters/connection_specification.rb:185:in `spec'
/home/dev/.rvm/gems/ruby-2.5.0/gems/activerecord-5.1.5/lib/active_record/connection_adapters/connection_specification.rb:188:in `rescue in spec': Specified 'mysql2' for database adapter, but the gem is not loaded. Add `gem 'mysql2'` to your Gemfile (and ensure its version is at the minimum required by ActiveRecord). (Gem::LoadError)
```

I simply changed the version in the Gemfile to 0.5.0 and rebundled.  If I downgrade to 0.4.10, everything works fine.

I'm using:

**Ruby:** 2.5.0
**Rails:** 5.15
**Ubuntu:** 16.04
**mysql:** mysql  Ver 14.14 Distrib 5.7.21, for Linux (x86_64) using  EditLine wrapper

sodabrew:
This is expected until the next minor revisions of Rails that enable mysql2 0.5.0 to load.

bverem:
Is mysql2 0.5.0 expected to work with Ruby 2.5.0?

sodabrew:
Yes, Ruby 2.5 is fully supported.

[...]

PR #1131, opened 2020-08-05
Clarify Rails 4.2.11 support in Readme

muzfuz:
The README states that `rails 4.2` does not support version `0.5` of the gem, however, as indicated by [this comment](https://github.com/brianmario/mysql2/issues/950#issuecomment-442228062) and [this release](https://github.com/rails/rails/compare/v4.2.10...v4.2.11), Rails `4.2.11` does actually support `0.5`.

This caused a bit of confusion at our end during the upgrade of a legacy app, as all tests seemed to pass even though the docs indicated otherwise.

I hope this small change is useful for someone in a similar situation.  Thank you :smile:

### DIFF
--- a/README.md
+++ b/README.md
@@ -572,7 +572,7 @@ This gem is tested with the following MySQL and MariaDB versions:
 
 ### Ruby on Rails / Active Record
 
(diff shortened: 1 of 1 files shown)


sodabrew:
Thank you!

PR #1135, opened 2020-09-07
Improve Homebrew compatibility

stefansundin:
Hi!

When I develop on my Mac, I usually install the Homebrew package `mysql-client` instead of `mysql`, since I don't really need the server (I usually run that in Docker). But unfortunately mysql2 doesn't compile out of the box.

I have worked around it by first running:
```
bundle config build.mysql2 --with-mysql-dir="/usr/local/opt/mysql-client"
```

But now I decided to look into contributing a fix.

Hopefully this will be accepted and included in the next version. Thanks!

### DIFF
--- a/ext/mysql2/extconf.rb
+++ b/ext/mysql2/extconf.rb
@@ -21,6 +21,11 @@ def add_ssl_defines(header)
   $CFLAGS << ' -DNO_SSL_MODE_SUPPORT' if has_no_support
 end
 
+# Homebrew openssl
+if RUBY_PLATFORM =~ /darwin/
+  $LDFLAGS << ' -L/usr/local/opt/openssl/lib'
+end
+
 # 2.1+
 have_func('rb_absint_size')
 have_func('rb_absint_singlebit_p')
--- a/ext/mysql2/extconf.rb
+++ b/ext/mysql2/extconf.rb
@@ -47,7 +47,9 @@ dirs = ENV.fetch('PATH').split(File::PATH_SEPARATOR) + %w[
   /usr/local/mysql-*
   /usr/local/lib/mysql5*
   /usr/local/opt/mysql5*
+  /usr/local/opt/mysql@*
   /usr/local/opt/mysql-client
+  /usr/local/opt/mysql-client@*
 ].map { |dir| dir << '/bin' }
 
 # For those without HOMEBREW_ROOT in PATH
--- a/ext/mysql2/extconf.rb
+++ b/ext/mysql2/extconf.rb
@@ -22,9 +22,7 @@ def add_ssl_defines(header)
 end
 
 # Homebrew openssl
-if RUBY_PLATFORM =~ /darwin/
-  $LDFLAGS << ' -L/usr/local/opt/openssl/lib'
-end
+$LDFLAGS << ' -L/usr/local/opt/openssl/lib' if RUBY_PLATFORM =~ /darwin/
 
 # 2.1+
 have_func('rb_absint_size')
--- a/ext/mysql2/extconf.rb
+++ b/ext/mysql2/extconf.rb
@@ -42,6 +42,7 @@ dirs = ENV.fetch('PATH').split(File::PATH_SEPARATOR) + %w[
   /usr/local/mysql-*
   /usr/local/lib/mysql5*
   /usr/local/opt/mysql5*
+  /usr/local/opt/mysql-client
 ].map { |dir| dir << '/bin' }
 
 # For those without HOMEBREW_ROOT in PATH


sodabrew (approved):
Looks good to me!

sodabrew:
Where does the mysql-server package install in Homebrew? The list of paths may be a bit out of date at this point, given MySQL 8 being the primary version to use now.

stefansundin:
It looks like with `mysql`, it finds `/usr/local/bin/mysql_config` and uses that. I am going to run some more testing to try to make this gem more compatible with different macOS setups. There isn't a separate `mysql-server` package for the server, so the `mysql` package installs both the server and the client.

For example, I don't have the full Xcode installed, so I was having some issues with linking openssl with the `mysql` package.

```
ld: library not found for -lssl
```

I was able to fix it by adding `$LDFLAGS << " -L/usr/local/opt/openssl/lib"` to `extconf.rb`. Not sure if there's a better way. Not sure why this didn't happen with `mysql-client`.

I also realized that there's a `mysql-client@5.7` package. There's also `mysql@5.6` and `mysql@5.7`. I can add more search paths until it works with all of the combinations.

```
$ brew search mysql | cat
==> Formulae
automysqlbackup
mysql
mysql++
mysql-client
mysql-client@5.7
mysql-connector-c++
mysql-connector-c++@1.1
mysql-sandbox
mysql-search-replace
mysql@5.6
mysql@5.7
mysqltuner
==> Casks
mysql-connector-python
mysql-shell
mysql-utilities
mysqlworkbench
navicat-for-mysql
sqlpro-for-mysql
```

(used `cat` to just get one entry per line, like `ls -1`)

I will do some more testing and report back.

stefansundin:
This explains the `-lssl` issue. For some reason there are `libssl` files that comes with `mysql-client` but not `mysql`.

```
$ /usr/local/bin/mysql_config --libs
-L/usr/local/Cellar/mysql/8.0.21_1/lib -lmysqlclient -lssl -lcrypto
$ /usr/local/opt/mysql-client/bin/mysql_config --libs
-L/usr/local/opt/mysql-client/lib -lmysqlclient -lssl -lcrypto
$ ls -1 /usr/local/Cellar/mysql/8.0.21_1/lib
libmysqlclient.21.dylib
libmysqlclient.a
libmysqlclient.dylib
libmysqlharness.1.dylib
libmysqlrouter.1.dylib
libmysqlrouter_http.1.dylib
libmysqlrouter_http_auth_backend.1.dylib
libmysqlrouter_http_auth_realm.1.dylib
libmysqlservices.a
mysqlrouter
pkgconfig
plugin
$ ls -1 /usr/local/opt/mysql-client/lib
libcrypto.1.1.dylib
libcrypto.dylib
libmysqlclient.21.dylib
libmysqlclient.a
libmysqlclient.dylib
libssl.1.1.dylib
libssl.dylib
pkgconfig
plugin
```

It may be because `mysql` sets [`-DWITH_SSL=#{Formula["openssl@1.1"].opt_prefix}`](https://github.com/Homebrew/homebrew-core/blob/9e4e4f9c53e83d5ff555f6faaa2afc7f72266f13/Formula/mysql.rb#L57) while `mysql-client` sets [`-DWITH_SSL=yes`](https://github.com/Homebrew/homebrew-core/blob/9e4e4f9c53e83d5ff555f6faaa2afc7f72266f13/Formula/mysql-client.rb#L41).

If you are ok with it, I think we should add something like this:
```
if RUBY_PLATFORM =~ /darwin/
  $LDFLAGS << ' -L/usr/local/opt/openssl/lib'
end
```

I imagine that plenty of Rubyists may be installing full Xcode to get this gem to work.

stefansundin:
I added two more commits. Let me know if you just want me to change `/usr/local/opt/mysql@*` to `/usr/local/opt/mysql*` instead, as that should catch a lot of paths. And let me know if you want me to squash the commits (or you can do it when you merge).

Thanks!

sodabrew:
All good, will squash and merge!

olivierlacan:
Hey @stefansundin. Unless I'm mistaken there's no `openssl` recipe for Homebrew, only https://formulae.brew.sh/formula/openssl@1.1#default which doesn't install in `/openssl/` anymore: 

```
$ brew --prefix openssl
/opt/homebrew/opt/openssl@1.1
```

Homebrew itself recommends this flag: 
```
For compilers to find openssl@1.1 you may need to set:
  export LDFLAGS="-L/opt/homebrew/opt/openssl@1.1/lib"
  export CPPFLAGS="-I/opt/homebrew/opt/openssl@1.1/include"
```

So at the risk of slowing things down a bit it might be good to ask Homebrew if it's installed and then ask it where openssl is installed if at all. Something like: 

```ruby
brew_installed = system("command -v brew") 
if brew_installed # because brew --prefix would explode otherwise
  openssl_location = `brew --prefix openssl`.strip # because output includes \n character
  $LDFLAGS << " -L#{openssl_location}/lib" if RUBY_PLATFORM =~ /darwin/ && openssl_location
end
```

I'll gladly submit a follow-up PR if this seems reasonable.

stefansundin:
@olivierlacan Sounds reasonable to me!

PR #1142, opened 2020-10-15
Updated README SSL documentation

ultratin:
Updated the README to better reflect the update SSL options that Mysql supports for versions >= 5.7.11.

I was struggling with understanding the difference between `verify_ssl` and `ssl_mode` and felt that clarifying this in the README would help.

### DIFF
--- a/README.md
+++ b/README.md
@@ -236,7 +236,6 @@ Mysql2::Client.new(
   :reconnect = true/false,
   :local_infile = true/false,
   :secure_auth = true/false,
-  :ssl_mode = :disabled / :preferred / :required / :verify_ca / :verify_identity,
   :default_file = '/path/to/my.cfg',
   :default_group = 'my.cfg section',
   :default_auth = 'authentication_windows_client'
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -263,8 +262,9 @@ Setting any of the following options will enable an SSL connection, but only if
 your MySQL client library and server have been compiled with SSL support.
 MySQL client library defaults will be used for any parameters that are left out
 or set to nil. Relative paths are allowed, and may be required by managed
-hosting providers such as Heroku. Set `:sslverify => true` to require that the
-server presents a valid certificate.
+hosting providers such as Heroku. For MySQL versions > `5.7.11` do not use `:sslverify`
```
sodabrew, line 265: I suggest to rephrase this in the positive, how about:
> For MySQL versions 5.7.11 and higher, use `:ssl_mode` to set a preferred or required SSL connections and certificate validation level. For earlier versions of MySQL, use the `:sslverify` boolean.

sodabrew: Off the top of my head, I don't recall the specific behavior if the user set `:sslverify` when connecting to MySQL 8. Based on the deprecation and now removal of this option, I think the mysql2 gem should enforce migrating users to :ssl_mode by way of warnings and failing the connection.

ultratin: Alright, changing it to sound more positive makes sense. Regarding the behaviour, I can open up an issue to track it.

sodabrew:
Cherry-picked as bab4969, belated thank you for this PR!

PR #1154, opened 2021-01-11
Add GitHub Actions.

junaruga:
This is a PR to start GitHub Actions. It's a minimal start to have the compatible or covered cases on GitHub Actions from Travis eventually. This is my first experience to work on GitHub Actions.

I find GitHub Actions specific limitations.

* I am using https://github.com/ruby/setup-ruby common library to install Ruby and Bundler, and `bundle install` that is commonly used in Ruby applications. However while the current setup needs to run the `bundle install --without benchmarks development` on Travis, I could not find the way to inject it to the common library on GitHub Actions. So, I modified `Gemfile` to update the groups to be optional as `bundle install` worked. You can see the detail [here](https://bundler.io/v2.2/guides/groups.html). Without the modification, the `bundle install` fails on the environment.
* According to [the GitHub Actions available ubuntu environment](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions), there are only 3 available environments, `ubuntu-20.04` (focal), `ubuntu-18.04` (bionic), `ubuntu-16.04` (xenial), while the trusty and xenial are used in Travis. See [the Ubuntu version history](https://en.wikipedia.org/wiki/Ubuntu_version_history).

I saw the following error on the cases on GitHub. Do you have any idea to fix this? This blocks the PR.

```
$ bash .travis_setup.sh
...
 + mysql -u root -e 'CREATE DATABASE /*M!50701 IF NOT EXISTS */ test'
ERROR 1045 (28000): Access denied for user 'root'@'localhost' (using password: NO)
Error: Process completed with exit code 1.
```

Thanks.

### DIFF
--- a/.travis_setup.sh
+++ b/.travis_setup.sh
@@ -2,6 +2,30 @@
 
 set -eux
 
+CHANGED_PWD=false
+# Change the password recreating the root user on mariadb < 10.2
+# where ALTER USER is not available.
+# https://stackoverflow.com/questions/56052177/
+CHANGED_PWD_BY_RECREATE=false
+
+# Install the default used DB if DB is not set.
+if [[ -n ${GITHUB_ACTION-} && -z ${DB-} ]]; then
+  if command -v lsb_release > /dev/null; then
+    case "$(lsb_release -cs)" in
+    xenial | bionic)
+      sudo apt-get install -qq mysql-server-5.7 mysql-client-core-5.7 mysql-client-5.7
+      CHANGED_PWD=true
+      ;;
+    focal)
+      sudo apt-get install -qq mysql-server-8.0 mysql-client-core-8.0 mysql-client-8.0
+      CHANGED_PWD=true
+      ;;
+    *)
+      ;;
+    esac
+  fi
+fi
+
 # Install MySQL 5.5 if DB=mysql55
 if [[ -n ${DB-} && x$DB =~ ^xmysql55 ]]; then
   sudo bash .travis_mysql55.sh
@@ -10,27 +34,45 @@ fi
 # Install MySQL 5.7 if DB=mysql57
 if [[ -n ${DB-} && x$DB =~ ^xmysql57 ]]; then
   sudo bash .travis_mysql57.sh
+  CHANGED_PWD=true
 fi
 
 # Install MySQL 8.0 if DB=mysql80
 if [[ -n ${DB-} && x$DB =~ ^xmysql80 ]]; then
   sudo bash .travis_mysql80.sh
+  CHANGED_PWD=true
 fi
 
 # Install MariaDB client headers after Travis CI fix for MariaDB 10.2 broke earlier 10.x
 if [[ -n ${DB-} && x$DB =~ ^xmariadb10.0 ]]; then
-  sudo apt-get install -y -o Dpkg::Options::='--force-confnew' libmariadbclient-dev
+  if [[ -n ${GITHUB_ACTION-} ]]; then
+    sudo apt-get install -y -o Dpkg::Options::='--force-confnew' mariadb-server mariadb-server-10.0 libmariadb2
+    CHANGED_PWD_BY_RECREATE=true
+  else
+    sudo apt-get install -y -o Dpkg::Options::='--force-confnew' libmariadbclient-dev
+  fi
 fi
 
 # Install MariaDB client headers after Travis CI fix for MariaDB 10.2 broke earlier 10.x
 if [[ -n ${DB-} && x$DB =~ ^xmariadb10.1 ]]; then
-  sudo apt-get install -y -o Dpkg::Options::='--force-confnew' libmariadbclient-dev
+  if [[ -n ${GITHUB_ACTION-} ]]; then
+    sudo apt-get install -y -o Dpkg::Options::='--force-confnew' mariadb-server mariadb-server-10.1 libmariadb-dev
+    CHANGED_PWD_BY_RECREATE=true
+  else
+    sudo apt-get install -y -o Dpkg::Options::='--force-confnew' libmariadbclient-dev
+  fi
 fi
 
 # Install MariaDB 10.2 if DB=mariadb10.2
 # NOTE this is a workaround until Travis CI merges a fix to its mariadb addon.
 if [[ -n ${DB-} && x$DB =~ ^xmariadb10.2 ]]; then
   sudo apt-get install -y -o Dpkg::Options::='--force-confnew' mariadb-server mariadb-server-10.2 libmariadbclient18
+fi
+
+# Install MariaDB 10.3 if DB=mariadb10.3
+if [[ -n ${GITHUB_ACTION-} && -n ${DB-} && x$DB =~ ^xmariadb10.3 ]]; then
+  sudo apt-get install -y -o Dpkg::Options::='--force-confnew' mariadb-server mariadb-server-10.3 libmariadb-dev
+  CHANGED_PWD=true
 fi
 
 # Install MySQL if OS=darwin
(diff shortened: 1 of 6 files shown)


junaruga:
> I saw the following error on the cases on GitHub. Do you have any idea to fix this? This blocks the PR.

Here is the error log on my forked repository.
https://github.com/junaruga/mysql2/runs/1683056833?check_suite_focus=true#step:5:32

junaruga:
I referred the following GitHub Actions config files.

* https://github.com/puma/puma/blob/master/.github/workflows/mri.yml
* https://github.com/rails/rails/blob/master/.github/workflows/rubocop.yml
* https://github.com/ruby/ruby/blob/master/.github/workflows/ubuntu.yml

junaruga:
> Here is the error log on my forked repository.
> https://github.com/junaruga/mysql2/runs/1683056833?check_suite_focus=true#step:5:32

It seems there is no mysql 5.6 packages such as `mysql-server-5.6` on ubuntu-16.04 xenial. But there is mysql 5.7 packages on it.

Running the following commands, those packages were already installed.

```
sudo apt-get -yq install mysql-server-5.7 mysql-client-core-5.7 mysql-client-5.7
```

And I see the following error about the password is on the mysql 5.7.

```
+ mysqld --version
mysqld  Ver 5.7.32-0ubuntu0.16.04.1 for Linux on x86_64 ((Ubuntu))
+ mysql -u root -e 'CREATE DATABASE /*M!50701 IF NOT EXISTS */ test'
ERROR 1045 (28000): Access denied for user 'root'@'localhost' (using password: NO)
Error: Process completed with exit code 1.
```

In `.travis_setup.sh`, the used package `libmariadbclient-dev` is not avalable on ubuntu-16.04 xenial. We need to find an alternative libmariadbclient-dev package to be used in xmariadb10.0 and xmariadb10.1 cases.

sodabrew:
This is great! Thank you! The `.travis_mysql57` / `.travis_mysql80.sh` scripts set the root password and should work for this.

sodabrew:
I think it would be reasonable to move all the tests up to Ubuntu 20.04. The most common problem I'm seeing in the issues tracker these days are out of date Rails versions on up to date MySQL versions. This is unexpected, in my experience I've always updated my application code long before the database engine!

junaruga:
> I think it would be reasonable to move all the tests up to Ubuntu 20.04.

I can agree on it. But there might be older mysql/mariadb versions that are not available on Ubuntu 20.04 but available on Ubuntu 18.04 or Ubuntu-16.04. So, I think we can "basically" use Ubuntu 20.04 for the cases as much as possible, but if the tested mysql/mariadb version's packages do not exist on Ubuntu 20.04, we can try to use another old Ubuntu environment.

Thanks for the info to fix the issues. Tomorrow I will take a look at it!

junaruga:
Rebased this PR.

> The .travis_mysql57 / .travis_mysql80.sh scripts set the root password and should work for this.

I still see the error by the SQL command to set the root password. Do you have any idea to fix it?

https://github.com/junaruga/mysql2/runs/1689931827?check_suite_focus=true#step:5:23

```
+ mysql -u root -e 'ALTER USER '\''root'\''@'\''localhost'\'' IDENTIFIED WITH mysql_native_password BY '\'''\'''
ERROR 1045 (28000): Access denied for user 'root'@'localhost' (using password: NO)
```

junaruga:
As a reference, when I tested the behavior on ubuntu focal container, it was succeeded by root user, though I actually used `podman` command rather than `docker` command.

```
$ docker run -it ubuntu:focal bash
# apt-get update -qq
# apt-get install -qq mysql-server-8.0 mysql-client-8.0
# service mysql start

# mysqld --version
/usr/sbin/mysqld  Ver 8.0.22-0ubuntu0.20.04.3 for Linux on x86_64 ((Ubuntu))

# mysql -u root -e "ALTER USER 'root'@'localhost' IDENTIFIED WITH mysql_native_password BY ''"
# mysql -u root -e 'CREATE DATABASE /*M!50701 IF NOT EXISTS */ test'
```

junaruga:
Rebased.

On mysql 8.0 case, there is one error.
https://github.com/junaruga/mysql2/runs/1696242902?check_suite_focus=true#step:7:535

```
  1) Mysql2::Client session_track returns multiple session track type values when available
     Failure/Error: expect(res).to eq(["________"])

       expected: ["________"]
            got: nil

       (compared using ==)
     # ./spec/mysql2/client_spec.rb:1053:in `block (3 levels) in <top (required)>'
```

On mariadb 10.0 case, I see a clue [here](https://stackoverflow.com/questions/44298160/mysql-mariadb-10-0-29-set-root-password-but-still-can-login-without-asking-p). We might do `sudo mysql ...` or the following SQL? I am not sure.

```
mysql -u root <<SQL
USE mysql;
desc mysql.user;
SELECT User, Host, Password, plugin, password FROM mysql.user;
SET PASSWORD FOR 'root'@localhost = PASSWORD("");
UPDATE mysql.user SET plugin = '' WHERE user = 'root' AND host = 'localhost';
FLUSH PRIVILEGES;
SELECT User, Host, Password, plugin, password FROM mysql.user;
SQL
```

junaruga:
I need someone's help to fix it.

junaruga:
Rebased. I think I am ready for the review now on this PR now.

Here is the result of GitHub Actions.
https://github.com/junaruga/mysql2/actions/runs/579780022

* Some job are set as `allow-failure: true` using [continue-on-error syntax](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstepscontinue-on-error). I noted the reason of the failures in `.github/workflows/test.yml` matrix part.
* Added the Ubuntu cases as much as possible.  Ruby `2.0.0` is not available on GitHub Actions' [ruby/setup-ruby@v1](https://github.com/ruby/setup-ruby) library. I found the way to run `bundle install` by ourselves. When we do not set `bundler-cache: true` on the library, we can do it. I could not add `mariadb10.1`, `mariadb10.2`, `mysql55` cases that exists on Travis CI, as I have no idea how to add it.
* The MacOSX and CentOS case with container are not included on this PR. I plan I will work on the CentOS case later for another PR.
* There is a logic such as `add-apt-repository 'http://repo.mysql.com/apt/ubuntu mysql-5.7'` to set the repository from the mysql.com. The problem of `add-apt-repository` is even when the repository does not exist on the target ubuntu version, the command finishes successfully returning the exit status 0. This causes installing unintended version's mysql-server. To solve this problem I added `wget -q --spider http://repo.mysql.com/apt/ubuntu/dists/$(lsb_release -cs)/mysql-5.7` to verify the actual repository URL.
* I was struggled to set up the config to run `mysql -u root` from regular user, especially in the cases using MySQL and MariaDB installed from Ubuntu official deb package. Finally I added the full grant user `root-test` such as `root` user.

@sodabrew Could you review? If it is okay, please merge the PR and enable GitHub Actions on this repository. Thanks.

.travis_setup.sh
```diff
@@ -2,6 +2,12 @@
 
 set -eux
 
+# Install the default used DB if DB is not set.
+if [[ -z ${DB-} ]]; then
+  sudo apt-get update -qq
+  sudo apt-get install -qq mysql-server-5.7 mysql-client-core-5.7 mysql-client-5.7
```
junaruga, line 8: I see this line causes an install error on the current Travis CI case on Ubuntu trusty.
https://travis-ci.org/github/brianmario/mysql2/jobs/759562170#L1200

If we still like to keep the Travis trusty case, maybe if changing `mysql-server-5.7 mysql-client-core-5.7 mysql-client-5.7` to `mysql-server mysql-client-core mysql-client`, we might avoid the error. I wanted to add the version string of the mysql as much as possible to avoid installing the unintended version.

junaruga: Rebased. The Github Actions log is [here](https://github.com/junaruga/mysql2/actions/runs/580986457).

junaruga:
Rebased to pass the Travis on this PR. ([Github Actions log](https://github.com/junaruga/mysql2/actions/runs/581837849), [Travis log](https://travis-ci.org/github/junaruga/mysql2/builds/759643864))

* Add the command to create `root-test` DB user on CentOS case in `.travis_setup_centos.sh`. Without this logic, the CentOS case fails when accessing to the server with the not existing `root-test` DB user.
* Remove the CI cases on Travis migrated to GitHub Actions.

junaruga:
> I was struggled to set up the config to run mysql -u root from regular user, especially in the cases using MySQL and MariaDB installed from Ubuntu official deb package. Finally I added the full grant user root-test such as root user.

I tried to find a way to use the `root` user as a db system user instead of `root-test` user adding [this commit](https://github.com/junaruga/mysql2/commit/116fced5dfece991e8ee50c7986e0ec63eb9e04a) on another branch in my forked repository.

The result is [here](https://github.com/junaruga/mysql2/runs/1945605981?check_suite_focus=true#step:8:212). The following error happens with mariadb 10.0, as maybe the `mysql_native_password` is not recognized as a syntax on the old DB.

```
+ sudo mysql --defaults-extra-file=/etc/mysql/debian.cnf -u root -e 'ALTER USER '\''root'\''@'\''localhost'\'' IDENTIFIED WITH mysql_native_password BY '\'''\'''
ERROR 1064 (42000) at line 1: You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near 'USER 'root'@'localhost' IDENTIFIED WITH mysql_native_password BY ''' at line 1
```

I think the merit of way of creating `root-test` user is more consistent for any testing environment than the way using the `root` user. Some contributors trying to run the tests also might not want to change the db system user `root`'s settings on their local environment. In other words, this new way using `root-test` might confuse the contributors. Because when the granted `root-test` user does not exist on DB, the test failures are not obvious. Here is [the result](https://travis-ci.org/github/brianmario/mysql2/jobs/759611596#L3664). I think the solution is to add a rake task to verify the test environment to `tasks/rspec.rake` before running actual tests. For example running the following command in the rake task.

```
$ mysql -u root-test -e 'SHOW DATABASES;'
```

junaruga:
Rebased. I was able to pass the tests without creating `root-test` user. [GitHub Actions log](https://github.com/junaruga/mysql2/actions/runs/588985561), [Travis log](https://travis-ci.org/github/brianmario/mysql2/builds/759981753).

sodabrew:
This looks amazing. I've read it closely, I don't see anything that jumps out wrong at me. Merging now, and we can break-fix if needed!

junaruga:
Thanks for your review and merging!! I am glad for that! Now we can ask people sending PRs for this repository to rebase the PR to check if the PR passes. When you see the PR passing, you see the PR to be merged.

I am working to migrate the CentOS case to GitHub Actions now.

junaruga:
> we can break-fix if needed!

Yeah, it's time to do it!

Issue #1149, opened 2020-12-10
Rethinking CI?


junaruga:
Do you know Travis's following announcements about their changes?

https://blog.travis-ci.com/oss-announcement
https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing
> The upcoming pricing change will affect those of you who are:
> Building on the macOS environment
> Building with more than 10 concurrent builds
> Building on a public repositories only

And I feel Travis on my forked repository is very slow now.

It seems that the Ruby project recently migrated the most cases to GitHub Actions remaining a few non-x86_64 jobs in Travis.
https://github.com/ruby/ruby/blob/f3c19ddfbcfde63505a34b7a5b3a640f23b22dc0/.travis.yml#L199-L204

So, how do you think of migrating Travis to GitHub Actions or an alternative CI?

sodabrew:
Yes, Travis is done. I want to migrate several projects to GitHub Actions. I would gladly take a PR before I have time to do it myself!

gautamkrishnar:
I Will give it a try during my free time

simi:
@gautamkrishnar I think it could be done in iterative way. It could be possible to get up quickly simple pipeline using linux + default mariadb in there and we can add the rest in following PRs. Feel free to ping me if any help is needed.

junaruga:
OK. Thanks for your help, guys!

junaruga:
@gautamkrishnar I sent the PR #1154 now, but I think you can also sent your PR.

gautamkrishnar:
@junaruga please feel free to work on it, I was busy with my work didn't get a chance to look into it.

junaruga:
I am waiting someone's review for my PR https://github.com/brianmario/mysql2/pull/1154 to migrate majority of cases in Travis to GitHub Actions. Anyone, could you review to merge it? Thanks.

sodabrew:
Thank you @junaruga !!

junaruga:
I opened a new PR https://github.com/brianmario/mysql2/pull/1168 to migrate centos case from Travis to GitHub Actions, and add fedora cases here.

junaruga:
Now I am working on migrating Mac OSX case in Travis to GitHub Actions.

junaruga:
I opened a new PR #1170 to migrate macOS case from Travis to GitHub Actions.

junaruga:
Note Travis-ci.org shut down will occur on May 31, 2021.
[...]

PR #1155, opened 2021-01-25
Run GC.verify_compaction_references on CI

casperisfine:
This is a followup to https://github.com/brianmario/mysql2/pull/1115

In 3.0 a new API was introduced that ask the GC to move all objects. This is useful to test for bugs like the ones that were fixed in #1115. 

cc @sodabrew

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -4,6 +4,12 @@ require 'timeout'
 require 'yaml'
 DatabaseCredentials = YAML.load_file('spec/configuration.yml')
 
+if GC.respond_to?(:verify_compaction_references)
+  # This method was added in Ruby 3.0.0. Calling it this way asks the GC to
+  # move objects around, helping to find object movement bugs.
+  GC.verify_compaction_references(double_heap: true, toward: :empty)
+end
+
 RSpec.configure do |config|
   config.disable_monkey_patching!
 


junaruga:
Thanks for the PR! I see the following error on Ruby 3.0 case.

https://travis-ci.org/github/brianmario/mysql2/jobs/756060952

```
/home/travis/.rvm/scripts/functions/cli: line 243: return: _ret: numeric argument required

The command "rvm get head" failed and exited with 2 during .
```

Perhaps, it's not `'3.0'` but `"3.0"` in the YAML syntax?

casperisfine:
> Perhaps, it's not `'3.0'` but `"3.0"` in the YAML syntax?

No, that's the same thing in YAML. It might be a travis bug.

Either way travis is shutting down soon, might be a good idea to migrate to GitHub Actions.

junaruga:
> No, that's the same thing in YAML. It might be a travis bug.

OK.

> Either way travis is shutting down soon, might be a good idea to migrate to GitHub Actions.

There has already been the issue ticket #1149 and the PR #1154 to migrate to GitHub Actions. However the PR is still working in progress.

sodabrew:
Since #1154 just landed to switch from Travis to Github Actions, could you rebase so we can see this run on GHA?

casperisfine:
sure

casperisfine:
Ok, same failures than on master, which I suppose is a success?

sodabrew:
Yep! Looks like the expected-to-pass tests pass and the expected-to-fail ones fail. Looks good to merge!

PR #1168, opened 2021-02-23
GitHub Actions: Add CentOS/Fedora cases.

junaruga:
## Summary

* Migrate CentOS case in Travis to GitHub Actions.
* Add Fedora cases as allow failure.

## Notes

* I added `.github/workflows/container.yml` as a workflow to test on a container. This file includes the migrated centos case and newly added fedora case. For now, I'd like to add the fedora cases because it has the test failure reported on the issue ticket. I'd like to share the reproductive cases here as I need someone's help to fix it. After fixing it, it might be dropped again if someone does not like it.
* I renamed `test.yml` to `ubuntu.yml`. The file name might be renamed to such as `ubuntu_osx.yml` in near future if the MacOSX case will be added to this file. But I think `ubuntu.yml` is accurate name for now. 
* `.dockerignore`: I added this file to avoid the unintended file is copied to the container from host environment such as `spec/configuration.yml` especially in the use case debugging on local environment.


junaruga:
The CIs are OK. GitHub Actions log [ubuntu](https://github.com/brianmario/mysql2/actions/runs/593739328), [container](https://github.com/brianmario/mysql2/actions/runs/593739327), [Travis log](https://travis-ci.org/github/junaruga/mysql2/builds/760209741) on my forked repository.

sodabrew:
Continues to look really good!

junaruga:
Thanks for reviewing and merging!

Issue #1165, opened 2021-02-12
mysql80: Mysql2::Client session_track fails on mysql >= 8.0.20 or 8.0.21?


junaruga:
I like to open the ticket of this test failure on mysql80 case, coming from https://github.com/brianmario/mysql2/pull/1154#issuecomment-759568657 . Actually I see this issue happens on the current Travis too.

```
Failures:
  1) Mysql2::Client session_track returns multiple session track type values when available
     Failure/Error: expect(res).to eq(["________"])

       expected: ["________"]
            got: nil
     
       (compared using ==)
     # ./spec/mysql2/client_spec.rb:1053:in `block (3 levels) in <top (required)>'
```

Seeing [the history of builds](https://travis-ci.org/github/brianmario/mysql2/builds).

* With mysql-server 8.0.19: ok : [Travis log](https://travis-ci.org/github/brianmario/mysql2/jobs/671107235#L764).
* With mysql-server 8.0.21: failure : [Travis log](https://travis-ci.org/github/brianmario/mysql2/jobs/725350411#L686).
* With mysql-server 8.0.23: failure : [Travis log](https://travis-ci.org/github/brianmario/mysql2/jobs/758232282#L980).

junaruga:
The changes in the mysql 8.0.20 or 8.0.21 might cause this issue.
https://dev.mysql.com/doc/relnotes/mysql/8.0/en/news-8-0-20.html
https://dev.mysql.com/doc/relnotes/mysql/8.0/en/news-8-0-21.html

junaruga:
You can see the reproducer of this issue on the GitHub Actions - ubuntu-20.04 ruby 2.4 mysql80 case here.
https://github.com/brianmario/mysql2/runs/1973366419?check_suite_focus=true

sodabrew:
Specifically this release note looks like the culprit:
```
The Performance Schema collected session-related statistics for errors that can occur only globally and not per session. This is no longer done, reducing memory overhead for error instrumentation. Additionally, rows for global errors are no longer included in error summaries reported per thread, account, user, or host. (Bug #30311574)
```

junaruga:
It seems it was fixed. See https://github.com/brianmario/mysql2/issues/1220#issuecomment-982961410 .

Faq:
As this is closed, noone updated those comments: https://github.com/brianmario/mysql2/blob/890342a5aa2554530299ccfa37bcc2eae28e4d02/.github/workflows/build.yml#L14

Issue #1152, opened 2021-01-07
tests failing with mariadb10.5: Mysql2::Statement


pvalena:
```
Failures:

  1) Mysql2::Error encoding returns error messages as UTF-8 by default
     Failure/Error: expect(bad_err.message).to include("??}\u001F")
       expected "You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near '??}\\001F' at line 1" to include "??}\u001F"
     # ./spec/mysql2/error_spec.rb:59:in `block (4 levels) in <top (required)>'
     # ./spec/spec_helper.rb:17:in `with_internal_encoding'
     # ./spec/mysql2/error_spec.rb:52:in `block (3 levels) in <top (required)>'

  2) Mysql2::Statement should create a statement
     Failure/Error: expect { statement = @client.prepare 'SELECT 1' }.to change(&method(:stmt_count)).by(1)
       expected result to have changed by 1, but was changed by 0
     # ./spec/mysql2/statement_spec.rb:18:in `block (2 levels) in <top (required)>'

  3) Mysql2::Statement close should free server resources
     Failure/Error: expect { stmt.close }.to change(&method(:stmt_count)).by(-1)
       expected result to have changed by -1, but was changed by 0
     # ./spec/mysql2/statement_spec.rb:702:in `block (3 levels) in <top (required)>'

Finished in 11.72 seconds (files took 0.21288 seconds to load)
322 examples, 3 failures

Failed examples:

rspec ./spec/mysql2/error_spec.rb:51 # Mysql2::Error encoding returns error messages as UTF-8 by default
rspec ./spec/mysql2/statement_spec.rb:16 # Mysql2::Statement should create a statement
rspec ./spec/mysql2/statement_spec.rb:699 # Mysql2::Statement close should free server resources
```
As those tests are old and simple I do not have a clue what is causing this.
Exactly same code builds / passes tests in our Fedora Stable (mariadb: 10.4.17).

The change may have been caused by update to openssl version: 1.1.1i
(According to one of our tools, which might be wrong.)
_ _ _ _

Ruby version: both 2.7 and 3.0 are failing
mariadb-server version: 10.5.8
gem version: 0.5.3

junaruga:
This error started on Fedora at the timing [1] of that the following dependency packages to build rubygem-mysql2 were changed. I assume it might be the gcc 11 specific issue.

gcc 10.2.1 => 11.0.0
[...]

Issue #965, opened 2018-04-27
Test failures on MariaDB performance_schema: OFF


junaruga:
I would report as I got 2 test failures on my local environment.

I tried it on current master latest branch (4a20d58a5a5b8096da3a220a3994853ca26c79c7).
Do you know what's wrong on my environment?

Thanks!

OS: Fedora 27
MariaDB 10.2
Ruby: 2.5.1

```
$ cat /etc/fedora-release 
Fedora release 27 (Twenty Seven)

$ rpm -q mariadb-devel mariadb-server
mariadb-devel-10.2.12-5.fc27.x86_64
mariadb-server-10.2.12-5.fc27.x86_64
```

Start MariaDB daemon.

```
$ sudo systemctl start mariadb
```

```
$ which mysql
/usr/bin/mysql

$ rpm -qf /usr/bin/mysql
mariadb-10.2.12-5.fc27.x86_64

$ mysql -u root -e 'CREATE DATABASE /*M!50701 IF NOT EXISTS */ test'

$ which ruby
/usr/local/ruby-2.5.1/bin/ruby

$ ruby -v
ruby 2.5.1p57 (2018-03-29 revision 63029) [x86_64-linux]

$ which bundle
/usr/local/ruby-2.5.1/bin/bundle

$ bundle -v
Bundler version 1.16.1

$ bundle install --path vendor/bundle --without benchmarks development

$ bundle list
Gems included by the bundle:
  * ast (2.4.0)
  * bundler (1.16.1)
  * diff-lcs (1.3)
  * eventmachine (1.2.5)
  * mysql2 (0.5.1)
  * parallel (1.12.1)
  * parser (2.5.1.0)
  * powerpack (0.1.1)
  * rainbow (2.2.2)
  * rake (10.4.2)
  * rake-compiler (1.0.4)
  * rspec (3.7.0)
  * rspec-core (3.7.1)
  * rspec-expectations (3.7.0)
  * rspec-mocks (3.7.0)
  * rspec-support (3.7.1)
  * rubocop (0.50.0)
  * ruby-progressbar (1.9.0)
  * unicode-display_width (1.3.2)
```

Run unit tests.

```
$ bundle exec rake spec
mkdir -p tmp/x86_64-linux/mysql2/2.5.1
cd tmp/x86_64-linux/mysql2/2.5.1
/usr/local/ruby-2.5.1/bin/ruby -I. ../../../../ext/mysql2/extconf.rb
checking for rb_absint_size()... yes
[...]

PR #1170, opened 2021-02-24
GitHub Actions: Add macOS cases.

junaruga:
* Migrate macOS cases in Travis to GitHub Actions.
* Increase the expected time in the test to avoid the following random failures
  on macOS.

  ```
  1) Mysql2::Client#query threaded queries should be supported
     Failure/Error: expect(stop - start).to be_within(0.1).of(sleep_time)
       expected 0.632204999999999 to be within 0.1 of 0.5
     # ./spec/mysql2/client_spec.rb:711:in `block (3 levels) in <top (required)>'
  ```

  ```
  3) Mysql2::Client#query should run signal handlers while waiting for a response
     Failure/Error: expect(mark.fetch(:QUERY_END) - mark.fetch(:QUERY_START)).to be_within(0.1).of(query_time)
       expected 1.1042130000000157 to be within 0.1 of 1.0
     # ./spec/mysql2/client_spec.rb:632:in `block (3 levels) in <top (required)>'
  ```

## Notes

* I renamed `.github/workflows/ubuntu.yml` to `build.yml` again, after adding the macOS cases to the file. I considered the name `ubuntu_macos.yml`. But thinking about we want to keep a same name for a long time, and considering the name (`name: Build`) is used as a build badge name when we will add it later. The name `Build` makes sense to me.

* In Travis macOS case, the brew package `mariadb-connector-c` was installed. So I always installed the `mariadb-connector-c` in GitHub Actions macOS case too. However when we do not actually install it on the `db: 'mysql@5.6'` case of the macOS, `rake compile` can be compiled, and there are some test failures. In `db: 'mariadb@10.4'` case, `rake compile` shows the following library missing error when `mariadb-connector-c` is not installed.
 
  ```
  checking for -lmysqlclient... no
  mysql client is missing. You may need to 'brew install mysql' or 'port install mysql', and try again.
  ```

  We might be able to install `mariadb-connector-c` conditionally on macOS case like this.

  ```
  brew install "${DB}"
  if [[ -n ${DB-} && x$DB =~ ^xmariadb ]]; then
    brew install mariadb-connector-c
  fi
  ```

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,4 +1,4 @@
-name: Ubuntu
+name: Build
 on: [push, pull_request]
 jobs:
   build:
@@ -38,6 +38,12 @@ jobs:
           # `service mysql restart` fails.
           - {os: ubuntu-20.04, ruby: 2.4, db: mysql80, allow-failure: true}
           - {os: ubuntu-18.04, ruby: 'head', db: '', allow-failure: true}
+          # db: the DB's brew package name in macOS case.
+          # Allow failure due to the following test failures.
+          # https://github.com/brianmario/mysql2/issues/965
+          - {os: macos-latest, ruby: 2.4, db: 'mariadb@10.4', allow-failure: true}
+          # Allow failure due to Mysql2::Error: Unknown system variable 'session_track_system_variables'.
+          - {os: macos-latest, ruby: 2.4, db: 'mysql@5.6', allow-failure: true}
       # On the fail-fast: true, it cancels all in-progress jobs
       # if any matrix job fails unlike Travis fast_finish.
       fail-fast: false
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -38,12 +38,16 @@ jobs:
           # `service mysql restart` fails.
           - {os: ubuntu-20.04, ruby: 2.4, db: mysql80, allow-failure: true}
           - {os: ubuntu-18.04, ruby: 'head', db: '', allow-failure: true}
-          # db: the DB's brew package name in macOS case.
+          # db: A DB's brew package name in macOS case.
+          #     Set a name "db: 'name@X.Y'" when using an old version.
+          # MariaDB lastet version
           # Allow failure due to the following test failures.
           # https://github.com/brianmario/mysql2/issues/965
-          - {os: macos-latest, ruby: 2.4, db: 'mariadb@10.4', allow-failure: true}
-          # Allow failure due to Mysql2::Error: Unknown system variable 'session_track_system_variables'.
-          - {os: macos-latest, ruby: 2.4, db: 'mysql@5.6', allow-failure: true}
+          # https://github.com/brianmario/mysql2/issues/1152
+          - {os: macos-latest, ruby: 2.4, db: mariadb, allow-failure: true}
+          # MySQL latest version
+          # Allow failure due to the issue #1165.
+          - {os: macos-latest, ruby: 2.4, db: mysql, allow-failure: true}
       # On the fail-fast: true, it cancels all in-progress jobs
       # if any matrix job fails unlike Travis fast_finish.
       fail-fast: false
--- a/.travis_setup.sh
+++ b/.travis_setup.sh
@@ -9,7 +9,7 @@ CHANGED_PWD=false
 CHANGED_PWD_BY_RECREATE=false
 
 # Install the default used DB if DB is not set.
-if [[ -n ${GITHUB_ACTION-} && -z ${DB-} ]]; then
+if [[ -n ${GITHUB_ACTIONS-} && -z ${DB-} ]]; then
   if command -v lsb_release > /dev/null; then
     case "$(lsb_release -cs)" in
     xenial | bionic)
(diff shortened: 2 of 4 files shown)


.travis_setup.sh
```diff
@@ -2,23 +2,24 @@
...
+CHANGED_PASSWORD=false
+# Change the password to be empty, recreating the root user on mariadb < 10.2
 # where ALTER USER is not available.
 # https://stackoverflow.com/questions/56052177/
-CHANGED_PWD_BY_RECREATE=false
+CHANGED_PASSWORD_BY_RECREATE=false
 
 # Install the default used DB if DB is not set.
-if [[ -n ${GITHUB_ACTION-} && -z ${DB-} ]]; then
+if [[ -n ${GITHUB_ACTIONS-} && -z ${DB-} ]]; then
```
junaruga, line 13: * I renamed the environment variable `GITHUB_ACTION` to `GITHUB_ACTIONS`. I think the `GITHUB_ACTIONS` showing always `true` on GitHub Actions environment is a right environment variable, and what I wanted to use. See https://docs.github.com/en/actions/reference/environment-variables .

.travis_setup.sh
```diff
@@ -2,23 +2,24 @@
...
 
-CHANGED_PWD=false
-# Change the password recreating the root user on mariadb < 10.2
+# Change the password to be empty.
+CHANGED_PASSWORD=false
+# Change the password to be empty, recreating the root user on mariadb < 10.2
 # where ALTER USER is not available.
 # https://stackoverflow.com/questions/56052177/
-CHANGED_PWD_BY_RECREATE=false
+CHANGED_PASSWORD_BY_RECREATE=false
```
junaruga, line 10: * I renamed `CHANGED_PWD` to `CHANGED_PASSWORD`, because `PWD` makes me imagine `pwd` (print working directory) command. Sorry it was my typo.

.travis_setup.sh
```diff
@@ -105,25 +116,20 @@ else
...
+  sudo mysql ${MYSQL_OPTS} -u "${DB_SYS_USER}" <<SQL
 DROP USER 'root'@'localhost';
 CREATE USER 'root'@'localhost' IDENTIFIED BY '';
 GRANT ALL PRIVILEGES ON *.* TO 'root'@'localhost' WITH GRANT OPTION;
 FLUSH PRIVILEGES;
 SQL
-  fi
-
-  # IF NOT EXISTS is mariadb-10+ only - https://mariadb.com/kb/en/mariadb/comment-syntax/
-  mysql -u root -e 'CREATE DATABASE /*M!50701 IF NOT EXISTS */ test'
```
junaruga, line 128: The new macOS `db: 'mysql@5.6'` case showed the error when running `mysql -u root -e 'CREATE DATABASE /*M!50701 IF NOT EXISTS */ test'`, because it was recognized as `mysql -u root -e 'CREATE DATABASE test'` and there was `test` database already. As every used database supports `IF NOT EXISTS` in CIs, I think we can remove the `/*M!50701 ... */`.

.github/workflows/build.yml
```diff
@@ -38,6 +38,12 @@ jobs:
           # `service mysql restart` fails.
           - {os: ubuntu-20.04, ruby: 2.4, db: mysql80, allow-failure: true}
           - {os: ubuntu-18.04, ruby: 'head', db: '', allow-failure: true}
+          # db: the DB's brew package name in macOS case.
+          # Allow failure due to the following test failures.
+          # https://github.com/brianmario/mysql2/issues/965
+          - {os: macos-latest, ruby: 2.4, db: 'mariadb@10.4', allow-failure: true}
+          # Allow failure due to Mysql2::Error: Unknown system variable 'session_track_system_variables'.
+          - {os: macos-latest, ruby: 2.4, db: 'mysql@5.6', allow-failure: true}
```
junaruga, line 46: While there was only one case mysql 5.6 in macOS, I added one more diverse case mariadb 10.4 that looks the latest version on the macOS environment, seeing `brew search mariadb`. It is to test the macOS case and the testing logic on GitHub Actions with diverse cases. It seems that macOS is slower than Ubuntu in GitHub Actions. But I think having the 2 macOS cases is still okay.

junaruga:
The CIs are okay for both GitHub Actions and Travis CI. I noticed the "Checks" tab showing the result of GitHub Actions. It looks convenient.

.github/workflows/build.yml
sodabrew: I think MySQL 5.6 is pretty old at this point. MySQL 8.0 is the way to go.

junaruga: @sodabrew OK I will change it to MySQL 8.0. By the way, I noticed that the brew package `mysql` without `@X.Y` is MYSQL 8.0, and the brew package `mariadb` is 10.5. These are the latest version on the macOS environment. I will change MariaDB to MariaDB 10.5 too.

```
mysql # => This is 8.0
mysql@5.6
mysql@5.7
mariadb => This is 10.5
mariadb@10.1
mariadb@10.2
mariadb@10.3
mariadb@10.4
```

junaruga:
I added one more commit for the review. Now [MariaDB 10.5 (10.5.9)](https://github.com/brianmario/mysql2/pull/1170/checks?check_run_id=1974205505#step:8:68) and [MySQL 8.0 (8.0.23)](https://github.com/brianmario/mysql2/pull/1170/checks?check_run_id=1974205524#step:8:68) are set on the macOS cases. I removed the `mariadb-connector-c` brew package of installing, because it was conflicted with the `mariadb` brew package. And both latest MariaDB and MySQL works without the `mariadb-connector-c`.

* On MariaDB 10.5 (10.5.9), the tests fail with #965, #1152.
* On MySQL 8.0 (8.0.23), the tests fail with #1165.

junaruga:
I did split the previous 2 commits to 4 commits, splitting the previous 1st commit about migrating macOS and renaming variables. I think it's better to see the commit about migrating macOS. The CIs looks okay. ([Travis log](https://travis-ci.org/github/junaruga/mysql2/builds/760570858) on my forked repository).

sodabrew:
Looks good!

PR #1171, opened 2021-03-02
Rename .travis_* files to ci/*.

junaruga:
The "ci" is the commonly used keyword not depending on the specific CI services.

Because for example the environment variable `CI=true` used in both GitHub Actions and Travis.
https://docs.github.com/en/actions/reference/environment-variables
https://docs.travis-ci.com/user/environment-variables/


sodabrew:
Yes! Great cleanup. I suggest this instead: new directory ci/ with each setup file moved there. No more dotfiles :)

junaruga:
Good idea! Let me change it on your suggested way.

junaruga:
Rebased. I changed the commit message too. The CIs look good.

PR #1173, opened 2021-03-02
Make the benchmarks group optional.

junaruga:
I have a suggestion that is to make the benchmark group optional. Right now the installation of the benchmark group by `bundle install` fails on Ruby  >= 2.4. I think this error might prevent a contributor from contributing. I think fixing this issue leads to the contributor's smooth onboarding. 

Here is [the current GitHub Actions log](https://github.com/junaruga/mysql2/runs/1977334233?check_suite_focus=true) when trying to install the benchmarks group. You see it fails on Ruby >= 2.4.

```
$ bundle install --without development
...
Gem::Ext::BuildError: ERROR: Failed to build gem native extension.
...
 current directory:
/opt/hostedtoolcache/Ruby/2.4.10/x64/lib/ruby/gems/2.4.0/gems/mysql-2.9.1/ext/mysql_api
make "DESTDIR="
compiling mysql.c
mysql.c: In function ‘stmt_bind_result’:
mysql.c:1320:74: error: ‘rb_cFixnum’ undeclared (first use in this function);
did you mean ‘rb_isalnum’?
else if (argv[i] == rb_cNumeric || argv[i] == rb_cInteger || argv[i] ==
rb_cFixnum)
^~~~~~~~~~
rb_isalnum
...
```

After making the benchmark optional, we can install the benchmarks group like this when we need it.

```
$ bundle install --with benchmarks
```

I am also thinking if we can make the development group optional too. The pros is we can test the `bundle install` without command line options on the supported platforms in CI. The cons is we have to run `bundle install --with development` when we need the gems in the development group.

What do you think?

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -58,7 +58,7 @@ jobs:
         with:
           ruby-version: ${{ matrix.ruby }}
       - run: ruby -v
-      - run: bundle install --without benchmarks development
+      - run: bundle install --without development
       - if: matrix.db != ''
         run: echo 'DB=${{ matrix.db }}' >> $GITHUB_ENV
       - run: sudo echo "127.0.0.1 mysql2gem.example.com" | sudo tee -a /etc/hosts
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 sudo: required
 dist: trusty
 language: ruby
-bundler_args: --without benchmarks development
+bundler_args: --without development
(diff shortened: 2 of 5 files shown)


junaruga:
This PR is also related to #931 . I think making the benchmarks group and the `mysql` gem optional is more acceptable than removing it, as it still works on Ruby < 2.4.

junaruga:
@sodabrew ping.

junaruga:
Thanks!

PR #1174, opened 2021-03-04
Verify the testing database before running tests.

junaruga:
## Motivation

Right now when the testing database is not running or has an error to the testing database, we see a lot of test failures on the testing result. By verifying the testing database connection before running the tests, we can guide users by the simple and proper message.

## New error message

The actual error message will be like this by this PR when there is an error while connecting to the testing database.
For example, this message is when the database server is not running.

```
$ bundle exec rake
...
Randomized with seed 54589

An error occurred while connecting to the testing database server.
Make sure that the database server is running.
Make sure that `mysql -u root [options] test` succeeds by the root user config in spec/configuration.yml.
Make sure that the testing database 'test' exists. If it does not exist, create it.

An error occurred in a `before(:suite)` hook.
Failure/Error: connect user, pass, host, port, database, socket, flags, conn_attrs

  Can't connect to local MySQL server through socket '/var/lib/mysql/mysql.sock' (2)
# ./lib/mysql2/client.rb:90:in `connect'
# ./lib/mysql2/client.rb:90:in `initialize'
# ./spec/spec_helper.rb:31:in `new'
# ./spec/spec_helper.rb:31:in `new_client'
# ./spec/spec_helper.rb:62:in `block (2 levels) in <top (required)>'

Finished in 0.03324 seconds (files took 0.17416 seconds to load)
0 examples, 0 failures, 1 error occurred outside of examples
```

I was inspired for the message content by Bundler error message guiding the actual command line `gem install ...` to make user verify it.

```
$ bundle install
...
An error occurred while installing mysql (2.9.1), and Bundler cannot continue.
Make sure that `gem install mysql -v '2.9.1' --source 'https://rubygems.org/'` succeeds before bundling.
...
```

Here are also other cases I tested on CI.

* `Mysql2::Error::ConnectionError: / Access denied for user 'root'@'localhost' (using password: NO)`: [the log](https://github.com/junaruga/mysql2/runs/2030679897?check_suite_focus=true#step:9:98).
* `Mysql2::Error: / Unknown database 'test'`: [the log](https://travis-ci.org/github/junaruga/mysql2/jobs/761427182#L766)

## Showing the message.

I wanted to use the following heredoc to show the message. But I faced the rubocop error on it, I am using `%( ... )` syntax to avoid it.

```
      message = <<-MESSAGE
foo
MESSAGE
```

## Sorting the before after blocks

I also sorted the before after blocks in the following actually called order.

```
before suite
before all
before each
after  each
after  all
after  suite
```

You can see the called order on [the rspec 2.99 document](https://relishapp.com/rspec/rspec-core/v/2-99/docs/hooks/before-and-after-hooks). As the used keywords `:all`, `:each` is old unlike the `:context` and `:example` on [the rspec 3 document](https://relishapp.com/rspec/rspec-core/v/3-10/docs/hooks/before-and-after-hooks). I will send another PR to replace the keywords later.

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -57,12 +57,21 @@ RSpec.configure do |config|
     end
   end
 
-  config.before :each do
-    @client = new_client
-  end
-
-  config.after :each do
-    @clients.each(&:close)
+  config.before(:suite) do
+    begin
+      new_client
+    rescue Mysql2::Error => e
+      username = DatabaseCredentials['root']['username']
+      database = DatabaseCredentials['root']['database']
+      message = %(
+An error occurred while connecting to the testing database server.
+Make sure that the database server is running.
+Make sure that `mysql -u #{username} [options] #{database}` succeeds by the root user config in spec/configuration.yml.
+Make sure that the testing database '#{database}' exists. If it does not exist, create it.
+)
+      warn message
+      raise e
+    end
   end
 
   config.before(:all) do
@@ -125,5 +134,13 @@ RSpec.configure do |config|
         )
       ]
     end
+  end
+
+  config.before(:each) do
+    @client = new_client
+  end
+
+  config.after(:each) do
+    @clients.each(&:close)
   end
 end


junaruga:
@sodabrew ping.

junaruga:
Thanks!

PR #1176, opened 2021-03-22
Refactor CI yaml files

junaruga:
There are 3 commits on this PR to refactor the CI yaml files.

* Fix wrong comments.
* GitHub Actions: Set the default value to allow-failure.
* Migrate Ruby 2.0.0 case in Travis to GitHub Actions.

## Fix wrong comments.

The comments for the 2 cases were actually opposite.

## GitHub Actions: Set the default value to allow-failure.

There is a way to omit the `allow-failure` when the case passes setting the default value. The tip is inspired by [the rspec-core's GitHub Actions yaml file](https://github.com/rspec/rspec-core/blob/ea8554afd1a2b63677c6593059fa8f2476181deb/.github/workflows/ci.yml#L57).

## Migrate Ruby 2.0.0 case in Travis to GitHub Actions.

We have actually Ruby 2.0.0 case on CentOS 7 case. I know the system Ruby in CentOS 7 is 2.0.0 until the CentOS 7 will be end of life. The tip using `name_extra` also comes from [the rspec-core's GitHub Actions yaml file](https://github.com/rspec/rspec-core/blob/ea8554afd1a2b63677c6593059fa8f2476181deb/.github/workflows/ci.yml#L20).

I thought we could remove the Ruby 2.0.0 case in Travis CI. In other words, we may still want to keep the Ruby 2.0.0 in Travis for some reasons.

What do you think?

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -7,7 +7,7 @@ jobs:
     # Run all the tests on the new environment as much as possible.
     # https://docs.github.com/en/free-pro-team@latest/actions/reference/specifications-for-github-hosted-runners
     runs-on: ${{ matrix.os }}
-    continue-on-error: ${{ matrix.allow-failure }}
+    continue-on-error: ${{ matrix.allow-failure || false }}
     strategy:
       matrix:
         os:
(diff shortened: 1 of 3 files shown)


junaruga:
@sodabrew Could you take a look this PR?

junaruga:
My concern is the Ruby 2.0.0 installed on Travis ubuntu by RVM, and installed as a RPM package on CentOS 7 are possibly a little different.  At least I can predict that Ruby 2.0.0 on CentOS is applying many patch files in it. I am a little negative to remove Ruby 2.0.0 on Travis now.

junaruga:
Thanks for checking and merging!

PR #1181, opened 2021-04-02
Make Result#fields return interned strings in Ruby 3+

casperisfine:
### Context

These strings are very likely to be used for hash keys. And when using a mutable string as a hash key, ruby will duplicate it and freeze the copy:

```ruby
>> k = "foo" + "bar"
=> "foobar"
>> k.object_id
=> 260
>> h = {k => 1}
=> {"foobar"=>1}
>> h.keys.first.object_id
=> 280
```

This leads to performance patches like this one: https://github.com/rails/rails/commit/a46dcb7454b56c979cded85f2f4f875dcd2cfdf0

Additionally, on recent rubies, when using a frozen string as a key, Ruby will intern the string to save on memory: https://github.com/msgpack/msgpack-ruby/commit/e8dedef58886578110887abba166632c08ee60c6

And finally, since Ruby 3.0, `rb_enc_interned_str` allow to directly retrieve an existing interned string without allocating any object. 

### This patch

Based on the three statements above, I believe that it would be preferable to directly intern the field string if possible, and if not to simply freeze them.

Of course this could cause some minor compatibility issues, but they quite unlikely and trivial to fix.

### DIFF
--- a/ext/mysql2/extconf.rb
+++ b/ext/mysql2/extconf.rb
@@ -31,6 +31,8 @@ have_func('rb_absint_singlebit_p')
 # Missing in RBX (https://github.com/rubinius/rubinius/issues/3771)
 have_func('rb_wait_for_single_fd')
 
+have_func("rb_enc_interned_str", "ruby.h")
+
 # borrowed from mysqlplus
 # http://github.com/oldmoe/mysqlplus/blob/master/ext/extconf.rb
 dirs = ENV.fetch('PATH').split(File::PATH_SEPARATOR) + %w[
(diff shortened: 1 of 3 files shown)


sodabrew:
This generally makes sense to me. Interned strings are not garbage collected until Ruby 2.2 and higher, however this gem supports Ruby 2.0 and higher. I would be concerned if the same string is interned multiple times, does it reuse the pre-existing intern or create another one? I'd want to be sure it's not going to create a memory leak for applications.

That said, if this does result in a memory leak but only for latest mysql2 gem releases on Ruby 2.0 and 2.1, then it would be sensible to set Ruby 2.2 as the new minimum version.

casperisfine:
> Interned strings are not garbage collected until Ruby 2.2 and higher

The API used here is only available in 3.0 and higher. For older rubies there basically no change.

> does it reuse the pre-existing intern or create another one? 

Always re-use yes.

sodabrew:
The subtle change is freezing the strings on older Rubies. It's an important relnote.

casperisfine:
> The subtle change is freezing the strings on older Rubies

Indeed, I forgot to point it out, this is for the interface to be consistent with newer rubies (because interned string implies frozen).

Thanks a lot for the merge!

Issue #1184, opened 2021-04-10
GitHub Actions: ubuntu: focal mariadb10.3 not starting.


junaruga:
This ticket is to manage the issue on the GitHub Actions: Ubuntu focal (20.04) mariadb (>= 10.3) case.

```
- {os: ubuntu-20.04, ruby: 2.4, db: mariadb10.3, allow-failure: true}
```

Today I debugged this environment. [the CI result with debugging code](https://github.com/junaruga/mysql2/runs/2313176749?check_suite_focus=true), [the working branch on my repo](https://github.com/junaruga/mysql2/tree/wip/ubuntu-focal-mariadb-10.3).

Right now I see the mariadb-server simply does not start correctly. It's not related to SSL.

Does anyone succeeded to install and start MariaDB server >= 10.3 on Ubuntu focal? 

So far I saw some issues, and here is my working log as a reference.

## apparmor issue

I saw this error log by `journalctl` on the CI environment.

```
$ journalctl -xe -n 100
...
Warning: :59:43 fv-az206-236 mysqld[2662]: 2021-04-10  8:59:43 0 [Warning] Could not increase number of max_open_files to more than 16384 (request: 32184)
Apr 10 08:59:44 fv-az206-236 audit[2662]: AVC apparmor="DENIED" operation="sendmsg" info="Failed name lookup - disconnected path" error=-13 profile="/usr/sbin/mysqld" name="run/systemd/notify" pid=2662 comm="mysqld" requested_mask="w" denied_mask="w" fsuid=115 ouid=0
Apr 10 08:59:44 fv-az206-236 kernel: audit: type=1400 audit(1618045184.198:34): apparmor="DENIED" operation="sendmsg" info="Failed name lookup - disconnected path" error=-13 profile="/usr/sbin/mysqld" name="run/systemd/notify" pid=2662 comm="mysqld" requested_mask="w" denied_mask="w" fsuid=115 ouid=0
...
```
It seems that when the called apparmor's profile for the mysqld (that is MySQL, not MariaDB) is enabled, then the mariadb-server is installed, this error happens. The solution is after disabling or removing the profile or removing, installing the mariadb-server.
Here is the detail.
https://mariadb.com/kb/en/the-community-mariadb-troubles-only-running-after-reboot-times-out-when-try/

## Warning: Could not increase number of max_open_files ...

I am not sure this is the direct cause of the error.

```
[Warning] Could not increase number of max_open_files to more than 16384 (request: 32184)
```

[...]

Issue #1182, opened 2021-04-09
Can't enable SSL with MariaDB driver library


vakuum:
It is currently not possible to enable SSL through the ssl_mode parameter when using the MariaDB driver library.

This is critical because you can't connect to a MySQL server with a user that requires SSL.

# How to reproduce?

Versions:

```
$ lsb_release -d
Description:    Ubuntu 20.04.2 LTS

$ ruby --version
ruby 2.6.6p146 (2020-03-31 revision 67876) [x86_64-linux]

$ gem --version
3.0.9

$ bundle --version
Bundler version 1.17.3

$ dpkg --list mysql-server-8.0
ii  mysql-server-8.0 8.0.23-0ubuntu0.20.04.1 amd64        MySQL database server binaries and system database setup
```

Create a MySQL database user with "REQUIRE SSL":

```
> CREATE USER IF NOT EXISTS test@127.0.0.1
IDENTIFIED WITH mysql_native_password BY 'test'
REQUIRE SSL;
```

Create a test script:

```
$ vim Gemfile
source 'https://rubygems.org'
gem 'mysql2'

$ vim test.rb
require 'mysql2'
client = Mysql2::Client.new(host: '127.0.0.1', username: 'test', password: 'test', ssl_mode: :required)
puts client.query('SHOW STATUS LIKE "Ssl_cipher"').first.inspect
```

## 🔴 Test with the MariaDB driver library

Install the driver library and the mysql2 gem:

```
$ sudo apt-get install libmariadb-dev-compat

$ dpkg --list libmariadb-dev-compat
...
ii  libmariadb-dev-compat:amd64 1:10.3.25-0ubuntu0.20.04.1 amd64        MariaDB Connector/C, compatibility symlinks

$ bundle exec gem uninstall mysql2
Successfully uninstalled mysql2-0.5.3

$ bundle
...
Installing mysql2 0.5.3 with native extensions
...

$ ldd $(bundle show mysql2)/lib/mysql2/mysql2.so
	...
	libmariadb.so.3 => /usr/lib/x86_64-linux-gnu/libmariadb.so.3 (0x00007efe06293000)
	...
```

Run the test script:

```
$ bundle exec ruby test.rb 
/home/clemens/.rvm/gems/ruby-2.6.6/gems/mysql2-0.5.3/lib/mysql2/client.rb:51: warning: Your mysql client library does not support ssl_mode as expected.
Traceback (most recent call last):
        3: from test.rb:2:in `<main>'
[...]

PR #1185, opened 2021-04-10
GitHub Actions: Comment out mariadb server 10.3 case on Ubuntu.

junaruga:
Shall we comment out the case for now?
Even the mariadb 10.3 server does not start on the environment.

I debugged it today related to #1184 . But it still does not work.


junaruga:
@sodabrew Could you review this PR?

junaruga:
I would close this PR as the MariaDB 10.3 case would work by the PR #1182 .

PR #1183, opened 2021-04-10
Can't enable SSL with MariaDB driver library. (#1182)

vakuum:
Please see #1182 for details.

PR #1186, opened 2021-04-10
WIP: GitHub Actions: Add DB server MySQL, DB client MariaDB case.

junaruga:
Related to #1182 , #1183 , I want to add the case: DB server mysql, DB client: mariadb.

### DIFF
--- a/ci/setup.sh
+++ b/ci/setup.sh
@@ -9,6 +9,8 @@ CHANGED_PASSWORD=false
 # https://stackoverflow.com/questions/56052177/
 CHANGED_PASSWORD_BY_RECREATE=false
 
+echo "DB: ${DB-}, DB_CLIENT: ${DB_CLIENT-}"
+
 # Install the default used DB if DB is not set.
 if [[ -n ${GITHUB_ACTIONS-} && -z ${DB-} ]]; then
   if command -v lsb_release > /dev/null; then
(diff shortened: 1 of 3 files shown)


ci/setup.sh
```diff
@@ -41,6 +43,12 @@ fi
 # Install MySQL 8.0 if DB=mysql80
 if [[ -n ${DB-} && x$DB =~ ^xmysql80 ]]; then
   sudo bash ci/mysql80.sh
+  if [[ -n ${DB_CLIENT-} && x$DB_CLIENT =~ ^xmariadb ]]; then
+    sudo apt-get install -qq libmariadb-dev-compat
+  else
+    sudo apt-get install -qq libmysqlclient-dev
+  fi
+
```
junaruga, line 51: I moved the logic to install libmysqlclient-dev or libmariadb-dev-compat from `mysql80.sh` to here. Because the environment variable `DB_CLIENT` is not available in the `sudo bash`.

simi:
@junaruga I'm just wondering, is this use-case even officially supported? Maybe it will be easier to put some kind of warning for this use-case instead.

junaruga:
@simi Great point. I am not sure about it. And maybe it's unlikely that each upstream MariaDB / MySQL project supports this case. Originally I wanted to have the CI case to verify the reported issue #1182. But we might be able to find another way to verify it without using this CI case.

junaruga:
I would close this PR for now.

PR #1187, opened 2021-04-11
CI: Set the verbose option in the Makefile to print compiling commands.

junaruga:
This PR is to see the compiling command lines including the used compiler, compiler flags and macros in the CI logs.

For example on this ticket https://github.com/brianmario/mysql2/issues/1182#issuecomment-817171322 , I think we want to know which CI case  has the defined `HAVE_CONST_MYSQL_OPT_SSL_ENFORCE` to check if  the logic `warning: Your mysql client library does not support ssl_mode as expected.` is tested.

Set the environment variable `MAKE="make V=1"` to print command lines to compile on CI.
For the `MAKE` environment, See https://github.com/rake-compiler/rake-compiler/blob/v1.1.1/lib/rake/extensiontask.rb#L477

After following my PR is merged, we can set this logic in `tasks/compile.rake`.
https://github.com/rake-compiler/rake-compiler/pull/192

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -63,6 +63,8 @@ jobs:
         run: echo "127.0.0.1 mysql2gem.example.com" | tee -a C:/Windows/System32/drivers/etc/hosts
       - run: echo 'DB=${{ matrix.db }}' >> $GITHUB_ENV
       - run: bash ci/setup.sh
+      # Set the verbose option in the Makefile to print compiling command lines.
+      - run: echo "MAKEFLAGS=V=1" >> $GITHUB_ENV
       - if: matrix.ssl
         run: echo "rake_spec_opts=--with-openssl-dir=$(brew --prefix ${{ matrix.ssl }})" >> $GITHUB_ENV
       - run: bundle exec rake spec -- $rake_spec_opts


junaruga:
* GitHub Actions - Build: [the log](https://github.com/brianmario/mysql2/pull/1187/checks?check_run_id=2317179300#step:9:36).
* GitHub Actions - Container: [the log](https://github.com/brianmario/mysql2/runs/2317179221?check_suite_focus=true#step:4:166).
* Travis: [the log](https://travis-ci.org/github/junaruga/mysql2/jobs/766661833#L1426) (on my forked repo).

junaruga:
@sodabrew Could you review this PR?

junaruga:
I heard we can use `MAKEFLAGS="V=1"` on the `rake-compiler` project. This is a better alternative of current way of this PR.

junaruga:
Later I will rebase this PR, using `MAKEFLAGS="V=1"` in `tasks/compile.rake`.

junaruga:
@sodabrew ping. I rebased. Could you review? Thanks.

junaruga:
@sodabrew I really want to see this PR will be merged. If you have a concern, please let me know. This PR is to make the debugging easier to fix issues.

If you are not comfortable for the `make -v` already executed, I can suggest that we can run the `make -v` only if the environment variable `RUBY_MYSQL2_DEBUG` is enabled. I see `RUBY_MYSQL2_` prefix is used in this repo already.

```
$ grep -r RUBY_MYSQL2_ lib/ ext/
lib/mysql2.rb:  dll_path = if ENV['RUBY_MYSQL2_LIBMYSQL_DLL']
lib/mysql2.rb:    ENV['RUBY_MYSQL2_LIBMYSQL_DLL']
```

And the approach for creating `rake debug` task to enable `RUBY_MYSQL2_DEBUG` is used in the nokogiri gem project. I like this approach.

https://github.com/sparklemotion/nokogiri/blob/18f6a3b9c2c385c0a1baef6459a563590989d1b7/rakelib/debug.rake#L1-L6

junaruga:
I tested the `rake debug` task by myself expecting `rake debug compile` works giving `MAKEFLAGS` for the `compiler` task. But it did not work because the compiler task is always called earlier than debug task.

So, I want to suggest the following way to set `MAKEFLAGS` if you have a concern about `MAKEFLAGS` to be always set.

```
$ RUBY_MYSQL2_DEBUG=true bundle exec rake
```

tasks/compile.rake

```
Rake::ExtensionTask.new("mysql2", Mysql2::GEMSPEC) do |ext|
  if ENV['RUBY_MYSQL2_DEBUG']
    # Set MAKEFLAGS for GNU/BSD make to print compiling commands in verbose mode.
    # https://github.com/rake-compiler/rake-compiler/pull/192
    ENV['MAKEFLAGS'] = 'V=1' if ENV['MAKEFLAGS'].nil?
  end
...
```

junaruga:
> So, I want to suggest the following way to set MAKEFLAGS if you have a concern about MAKEFLAGS to be always set.

Rebased the PR to set env MAKEFLAGS for verbose mode if env RUBY_MYSQL2_DEBUG is set. The `RUBY_MYSQL2_DEBUG` is set in the each CIs.  A main purpose of this PR is it's useful to see the compile flags for debugging when we find an issue on CI.

sodabrew:
Catching up on this older PR -- would the simplest option be to set the environment variable `MAKEFLAGS="V=1"` in CI with no other code changes

junaruga:
That's a good idea! I will rebase this PR when I have spare time this week.

junaruga:
The ruby/ruby is using GNUMAKEFLAGS ratehr than MAKEFLAGS. I am not sure which is the best in this case.

https://github.com/ruby/ruby/blob/1e2523fad7c03dc4f3b298d224548abd07374761/.github/workflows/compilers.yml#L226-L228

Here is the document.
https://www.gnu.org/software/make/manual/html_node/Options_002fRecursion.html
> If you’d like to run other implementations of make in addition to GNU make, and hence do not want to add GNU make-specific flags to the MAKEFLAGS variable, you can add them to the GNUMAKEFLAGS variable instead. This variable is parsed just before MAKEFLAGS, in the same way as MAKEFLAGS. When make constructs MAKEFLAGS to pass to a recursive make it will include all flags, even those taken from GNUMAKEFLAGS. As a result, after parsing GNUMAKEFLAGS GNU make sets this variable to the empty string to avoid duplicating flags during recursion.
>
> It’s best to use GNUMAKEFLAGS only with flags which won’t materially change the behavior of your makefiles. If your makefiles require GNU make anyway then simply use MAKEFLAGS. Flags such as ‘--no-print-directory’ or ‘--output-sync’ may be appropriate for GNUMAKEFLAGS.

junaruga:
Rebased the PR, updating the content and commit message.

I noticed that the `GNUMAKEFLAGS` doesn't work for the make used in the CI MacOS cases [in my experiment on my forked repo](https://github.com/junaruga/mysql2/actions/runs/3986546498/jobs/6835254878#step:10:16
). Maybe it's not GNU make.

Below is the updated commit message. @sodabrew review please.

---

CI: Set the verbose option in the Makefile to print compiling commands.

Enable the verbose option in the Makefile used to compile the extension to print the compiling command lines in the log.

Note that we use MAKEFLAGS rather than GNUMAKEFLAGS[1], because the GNUMAKEFLAGS doesn't work for the make used in CI MacOS cases.

We don't need to set the verbose option to the CI Fedora and CentOS cases (`.github/workflows/container.yml`). Because the Rubies from the Ruby RPM package used in Fedora and CentOS cases are already enabling the verbose option.

[1] https://www.gnu.org/software/make/manual/html_node/Options_002fRecursion.html

junaruga:
I wanted to make the MAKEFLAGS flag only affect to the mysql2 building process, and not affect to the other dependency software's building process. So, I added the MAKEFLAGS before the `bundle exec rake spec` rather than setting in the `env` element part.

sodabrew:
Looks good, merged!

PR #1192, opened 2021-04-15
Support ruby GC compaction

fatkodima:
ruby 2.7+ introduced GC compaction via `GC.compact`. This PR helps this gem to be compaction friendly.


sodabrew:
Thank you very much for this! I will read up on compaction to learn more about it! Very likely I want to delay landing this until after the pending 0.5.x release so that there's a new stable version folks could fall back to in case of any instability caused by this change.

sodabrew:
@tenderlove This looks good to me! Any review notes at your end?

sodabrew:
Sorry about the very long delay. Would you mind rebasing since merging the #1284 made some minor conflicting changes? (If you don't have time I'll swing back to this and adjust in a fixit commit.)

fatkodima:
@sodabrew Rebased.

PR #1205, opened 2021-08-27
Allow setting VERIFY_IDENTITY for MariaDB

dbussink:
This adds support for setting the VERIFY_IDENTITY mode with MariaDB. On MariaDB, the `MYSQL_OPT_SSL_VERIFY_SERVER_CERT` option is available which is equivalent to `VERIFY_IDENTITY`.

Many containers with Ruby apps are based on Debian where MariaDB is the standard provided, so this improves support there significantly.

### DIFF
--- a/ext/mysql2/client.c
+++ b/ext/mysql2/client.c
@@ -71,6 +71,10 @@ static ID intern_brackets, intern_merge, intern_merge_bang, intern_new_with_args
 /*
  * compatibility with mysql-connector-c 6.1.x, and with MySQL 5.7.3 - 5.7.10.
  */
+#ifdef HAVE_CONST_MYSQL_OPT_SSL_VERIFY_SERVER_CERT
+  #define SSL_MODE_VERIFY_IDENTITY 5
+  #define HAVE_CONST_SSL_MODE_VERIFY_IDENTITY
+#endif
 #ifdef HAVE_CONST_MYSQL_OPT_SSL_ENFORCE
   #define SSL_MODE_DISABLED 1
   #define SSL_MODE_REQUIRED 3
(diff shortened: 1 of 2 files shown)


ext/mysql2/client.c
```diff
@@ -1676,10 +1689,15 @@ void init_mysql2_client() {
   rb_const_set(cMysql2Client, rb_intern("SSL_MODE_REQUIRED"), INT2NUM(SSL_MODE_REQUIRED));
   rb_const_set(cMysql2Client, rb_intern("SSL_MODE_VERIFY_CA"), INT2NUM(SSL_MODE_VERIFY_CA));
   rb_const_set(cMysql2Client, rb_intern("SSL_MODE_VERIFY_IDENTITY"), INT2NUM(SSL_MODE_VERIFY_IDENTITY));
-#elif defined(HAVE_CONST_MYSQL_OPT_SSL_ENFORCE) // MySQL 5.7.3 - 5.7.10
+#else
+#ifdef HAVE_CONST_MYSQL_OPT_SSL_VERIFY_SERVER_CERT // MySQL 5.7.3 - 5.7.10
```
sodabrew, line 1693: `and MariaDB xx .. [yy? and above?]`

sodabrew (commented):
This looks excellent. Thank you for taking the time to carefully work these out. I have just one comment about modifying a comment

dbussink:
> I think long-term that supporting the brief set of versions MySQL 5.7.3 - 5.7.10 should just be dropped. Unless that same pattern is actually what MariaDB needs long-term, in which case I'd like to get the comments right to guide future-us correctly when it's deprecation time!

It is what MariaDB needs long-term, they are still using those options for the latest versions so I don't think they can be removed / deprecated unless MariaDB compatibility is explicitly dropped. I'll update the comments accordingly as well.

ext/mysql2/client.c
```diff
@@ -121,19 +125,27 @@ static VALUE rb_set_ssl_mode_option(VALUE self, VALUE setting) {
     rb_warn( "Your mysql client library does not support setting ssl_mode; full support comes with 5.7.11." );
     return Qnil;
   }
-#ifdef HAVE_CONST_MYSQL_OPT_SSL_ENFORCE
+#if defined(HAVE_CONST_MYSQL_OPT_SSL_VERIFY_SERVER_CERT) || defined(HAVE_CONST_MYSQL_OPT_SSL_ENFORCE)
   GET_CLIENT(self);
   int val = NUM2INT( setting );
-  // Either MySQL 5.7.3 - 5.7.10, or Connector/C 6.1.3 - 6.1.x, or MariaDB 10.x
-  if ((version >= 50703 && version < 50711) || (version >= 60103 && version < 60200) || (version >= 100000 && version < 110000)) {
```
dbussink, line 128: I have updated the comments and also dropped the `< 110000` here since there's no indication that MariaDB will change here for any future version (there's no 11.0 yet though, but I think it would likely break if they'd release such a version).

dbussink:
@sodabrew Alright, updated things here. I was on vacation for a bit so only came back to it now :smile:.

PR #1206, opened 2021-08-27
Setup default CA path if not provided

dbussink:
This adds setup of a default CA path if there's no path provided by the user. This enables easier configuration of system level CA validation if the MySQL server has a certificate signed by a system root.

On more and more cloud based MySQL platforms system signed CA certificates are used and this hides the issue of selecting the appropriate path from the user.

The real longer term answer here is that this is a default that changes in libmysqlclient itself. The current situation here is mixed. When using MariaDB (including the changes in #1205), the default system roots are already loaded and used if no CA is provided.

On MySQL itself on the other hand, a CA path is required today. I have also opened a PR to improve that, see
https://github.com/mysql/mysql-server/pull/358 & https://bugs.mysql.com/bug.php?id=104649.

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -17,7 +17,7 @@ Layout/IndentHeredoc:
 
 # Offense count: 2
 Metrics/AbcSize:
-  Max: 90
+  Max: 91
 
 # Offense count: 31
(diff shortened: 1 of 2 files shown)


sodabrew:
This is excellent, thank you for the PR!

Issue #1208, opened 2021-09-02
tests: Invalid DATETIME test fails with mysql >= 8.0.24


paride:
The test suite is failing when running against MySQL >= 8.0.24 with:

```
Failures:

  1) Mysql2::Result row data type mapping should raise an error given an invalid DATETIME
     Failure/Error:
       expect { @client.query("SELECT CAST('1972-00-27 00:00:00' AS DATETIME) as bad_datetime").each }.to \
         raise_error(Mysql2::Error, "Invalid date in field 'bad_datetime': 1972-00-27 00:00:00")

       expected Mysql2::Error with "Invalid date in field 'bad_datetime': 1972-00-27 00:00:00" but nothing was raised
     # ./spec/mysql2/result_spec.rb:321:in `block (3 levels) in <top (required)>'

  2) Mysql2::Statement row data type mapping should raise an error given an invalid DATETIME
     Failure/Error:
       expect { @client.query("SELECT CAST('1972-00-27 00:00:00' AS DATETIME) as bad_datetime").each }.to \
         raise_error(Mysql2::Error, "Invalid date in field 'bad_datetime': 1972-00-27 00:00:00")

       expected Mysql2::Error with "Invalid date in field 'bad_datetime': 1972-00-27 00:00:00" but nothing was raised
     # ./spec/mysql2/statement_spec.rb:478:in `block (3 levels) in <top (required)>'
```

This is because according to the release notes (https://dev.mysql.com/doc/relnotes/mysql/8.0/en/news-8-0-24.html):
> Using `CAST()` on a `DATE` or `DATETIME` value in an `INSERT` statement raised a warning for `"0000-00-00"` and `"2000-02-31"`, but not for `"2000-01-00"` or `"2000-00-01"`. Now a warning is shown in each of these cases also.

The `statement_spec.rb` and `result_spec.rb` files need to be updated with a bad date that still generates an error, e.g. `2000-01-00`.

casperisfine:
This is tricky, because it seems that casting to datetime pretty much never error out anymore, but just shows a warning:

```
mysql> SELECT CAST('WAT' AS DATETIME); SHOW WARNINGS;
+-------------------------+
| CAST('WAT' AS DATETIME) |
+-------------------------+
| NULL                    |
+-------------------------+
1 row in set, 1 warning (0.00 sec)

+---------+------+---------------------------------+
| Level   | Code | Message                         |
+---------+------+---------------------------------+
[...]

PR #1221, opened 2021-11-26
Update DATETIME casting tests for mysql 8.0

casperisfine:
Fix: https://github.com/brianmario/mysql2/issues/1208

With Mysql 5, the invalid datetime was returned.

However with 8.0 the returned row is `NULL`, so we can no longer detect this case and raise.

NB: I chose not to change the behavior on 5.x, but we could also chose to change it to match 8.0 behavior. However I don't see any way to bring back that behavior on 8.0 because it's impossible to distinguish between a `NULL` datetime and an invalid one.

### DIFF
--- a/spec/mysql2/result_spec.rb
+++ b/spec/mysql2/result_spec.rb
@@ -410,8 +410,13 @@ RSpec.describe Mysql2::Result do
     end
 
     it "should raise an error given an invalid DATETIME" do
-      expect { @client.query("SELECT CAST('1972-00-27 00:00:00' AS DATETIME) as bad_datetime").each }.to \
-        raise_error(Mysql2::Error, "Invalid date in field 'bad_datetime': 1972-00-27 00:00:00")
+      if @client.info[:version] < "8.0"
+        expect { @client.query("SELECT CAST('1972-00-27 00:00:00' AS DATETIME) as bad_datetime").each }.to \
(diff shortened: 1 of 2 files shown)


sodabrew:
Thanks!

spec/mysql2/result_spec.rb
```diff
@@ -410,8 +410,13 @@
     end
 
     it "should raise an error given an invalid DATETIME" do
-      expect { @client.query("SELECT CAST('1972-00-27 00:00:00' AS DATETIME) as bad_datetime").each }.to \
-        raise_error(Mysql2::Error, "Invalid date in field 'bad_datetime': 1972-00-27 00:00:00")
+      if @client.info[:version] < "8.0"
```
junaruga, line 413: I seems that this part and another part on this PR to compare the MySQL client version will not work when the MySQL 10.0 is released in the future. The following way could be the solution.

```
if (@client.info[:version].split('.').map(&:to_i) <=> [8, 0]) < 0
```

The version is about MySQL not about Gem's, but the following way could be also possible technically.

```
if Gem::Version.new(@client.info[:version]) < Gem::Version.new("8.0")-
```

I have not tested it by myself.

casperisfine: To best honest it's test code, so it's not like it's gonna break apps updating or anything and we can always update it when we add MySQL 10 to the CI in 4 or 5 years.

But yeah, feel free to PR the `Gem::Version` one.

PR #1225, opened 2021-12-01
Move rubocop to a separate CI job

casperisfine:
We use a very old version of rubocop to support ruby 2.0, however that version is not compatible with ruby-head.

Either way there's not point running rubocop on every ruby version a distinct CI job is more efficient.


casperisfine:
ruby-head is now green. cc @tenderlove

Issue #1179, opened 2021-03-24
Could we have a new release? (0.5.4)


casperisfine:
Sorry for asking this, I know it's a bit annoying for maintainers. But there's a bunch of important fixes on master and there wasn't any release in over a year. 

We've been using `mysql2` as a git gem for a long time, and I'd love if we could finally go back to a published version.

casperisfine:
@sodabrew sorry to insist, but it's really annoying to have to use the git gem on all new projects.

TomA-R:
This would be really useful, maybe @brianmario or @tenderlove could help? Thanks a lot in advance

tenderlove:
I have access to do a release, but I haven't committed to this project in quite a while.  @sodabrew do you have time to ship a release?  If not, I can take care of it (but I don't want to step on your toes 🙇🏻‍♂️)

sodabrew:
Thanks for the nudge, I really do need to do this!

listerino:
hey @sodabrew is this something we can help with?

listerino:
Happy 2 year anniversary to the last release! 🎉  57 commits and 1300 lines of change later, it seems like cutting a new release is simply too much effort. Can we get another maintainer @sodabrew? Otherwise maybe @tenderlove can take this task over?

simi:
I think we should fix master CI first. @listerino I think you're more than welcomed to help on that. I can try to spend some time in that during the weekend as well.

listerino:
That's a fair point, @simi! I'll have a look into it, thank you for also volunteering, it's appreciated

casperisfine:
> I think we should fix master CI first.

There are currently 3 issues on CI:

  - One client session tracking test fail against mysql 8: https://github.com/brianmario/mysql2/issues/1220 I pinged the original author and he's a college so I'll make sure to followup on that one.
  - There's a small Ruby 3.1 change, I submitted a PR for it: https://github.com/brianmario/mysql2/pull/1219
  - Then there's two date casting test that need to be updated for mysql 8 https://github.com/brianmario/mysql2/issues/1208, I'll submit a PR too.

casperisfine:
For the datetime one: https://github.com/brianmario/mysql2/pull/1221

So only the session tracking one has no fix yet.

listerino:
You're a machine, thank you, Jean. I was planning to set my weekend aside to investigate what took you ~2 hours - fingers crossed your fixes get reviewed&merged!

[...]

Issue #1194, opened 2021-04-16
macOS MariaDB 10.5 test failures


junaruga:
This issue ticket is to manage the test failures on GitHub Actions macOS MariaDB 10.5 test failures.

https://github.com/brianmario/mysql2/pull/1193/checks?check_run_id=2363155750#step:9:536

```
  1) Mysql2::EM::Client should support async queries
     Failure/Error: expect(results[0].keys).to include("second_query")
       expected ["first_query"] to include "second_query"
     # ./spec/em/em_spec.rb:26:in `block (2 levels) in <top (required)>'

  2) Mysql2::EM::Client should not raise error when closing client with no query running
     Failure/Error: expect(callbacks_run).to eq([:callback])

       expected: [:callback]
            got: []

       (compared using ==)

       Diff:
       @@ -1 +1 @@
       -[:callback]
       +[]
     # ./spec/em/em_spec.rb:124:in `block (4 levels) in <top (required)>'
     # ./spec/em/em_spec.rb:114:in `block (2 levels) in <top (required)>'

  3) Mysql2::Client#query should be impervious to connection-corrupting timeouts in #execute
     Failure/Error: expect { Timeout.timeout(0.1) { stmt.execute(0.2) } }.to raise_error(Timeout::Error)
       expected Timeout::Error but nothing was raised
     # ./spec/mysql2/client_spec.rb:659:in `block (3 levels) in <top (required)>'

Finished in 13.92 seconds (files took 0.9592 seconds to load)
333 examples, 3 failures, 3 pending
```

The failures **rarely** happens.
For example here is the success case on my forked repository.
https://github.com/junaruga/mysql2/runs/2363101517?check_suite_focus=true

junaruga:
Another case for the failure. It's not rare but sometimes happens.

https://github.com/brianmario/mysql2/runs/2365625101?check_suite_focus=true#step:9:535

```
Failures:

  1) Mysql2::EM::Client should support async queries
     Failure/Error: expect(results[0].keys).to include("second_query")
       expected ["first_query"] to include "second_query"
     # ./spec/em/em_spec.rb:26:in `block (2 levels) in <top (required)>'

Finished in 15.98 seconds (files took 0.76309 seconds to load)
333 examples, 1 failure, 3 pending
```

junaruga:
It seems this issue happens on MacOS mysql case too.
Here is the log on the latest master branch.
[...]

PR #1226, opened 2021-12-01
Fix compilation on macOS CI

casperisfine:
Ref: https://github.com/brianmario/mysql2/issues/1179

The mysql & mariadb formulas depend on an older openssl and leave
`brew --prefix openssl` pointing to an empty directory.

This fixes the `ld: library not found for -lssl` error on CI.

NB: There's one test failing afterward, but I'll start digging into it too.

cc @tenderlove

### DIFF
--- a/ci/setup.sh
+++ b/ci/setup.sh
@@ -99,9 +99,14 @@ if [[ x$OSTYPE =~ ^xdarwin ]]; then
 
   mysql.server start
   CHANGED_PASSWORD_BY_RECREATE=true
+
+  # the mysql & mariadb formulas depend on an older openssl and leave
+  # `brew --prefix openssl` pointing to an empty directory.
+  # So we install it again.
+  brew install openssl
 fi
 
-# TODO: get SSL working on OS X in Travis
+# TODO: get SSL working on OS X
 if ! [[ x$OSTYPE =~ ^xdarwin ]]; then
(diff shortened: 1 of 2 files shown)


casperisfine:
The two remaining failing test on macOS/mysql:

```
  1) Mysql2::Client#query evented async queries should be supported
     Failure/Error: expect(loops >= 1).to be true

       expected true
            got false
     # ./spec/mysql2/client_spec.rb:756:in `block (3 levels) in <top (required)>'

  2) Mysql2::EM::Client should not raise error when closing client with no query running
     Failure/Error: expect(callbacks_run).to eq([:callback])

       expected: [:callback]
            got: []

       (compared using ==)

       Diff:
       @@ -1 +1 @@
       -[:callback]
       +[]
     # ./spec/em/em_spec.rb:124:in `block (4 levels) in <top (required)>'
     # ./spec/em/em_spec.rb:114:in `block (2 levels) in <top (required)>'
```

casperisfine:
Hu, I can't reproduce these errors locally :/

tenderlove:
@casperisfine I think I fixed this in #1224 

Can you check?

tenderlove:
Also it looks like the EM test might be flaky, see: https://github.com/brianmario/mysql2/issues/1194

PR #1236, opened 2022-01-17
Undefine T_DATA allocators for Ruby 3.2 compatibility

casperisfine:
Ref: https://bugs.ruby-lang.org/issues/18007

cc @tenderlove 

```
lib/mysql2/em.rb:17: warning: undefining the allocator of T_DATA class Mysql2::Result
spec/mysql2/statement_spec.rb:56: warning: undefining the allocator of T_DATA class Mysql2::Statement
```

### DIFF
--- a/ext/mysql2/result.c
+++ b/ext/mysql2/result.c
@@ -1174,6 +1174,7 @@ void init_mysql2_result() {
   rb_global_variable(&cDateTime);
 
   cMysql2Result = rb_define_class_under(mMysql2, "Result", rb_cObject);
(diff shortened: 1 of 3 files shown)


sodabrew:
This appears to be fully backwards-compatible to earlier versions of Ruby as well?

casperisfine:
> This appears to be fully backwards-compatible to earlier versions of Ruby as well?

Yes, this isn't a new API, it just start to be slowly enforced.

> if you don't have any further changes to propose.

Nope all good for me.

PR #1244, opened 2022-02-23
Expose `db` attribute of `MYSQL` client struct

richardlarocque:
Adds an accessor method that returns the current value of the `MSYQL`
client struct's `db` attribute.

The `MYSQL` client struct includes a field `char *db`.  When the
`session_track_schema` setting is enabled, this field will be updated
using information from server-provided "OK" packets, keeping it in sync
as the client switches between databases.

### DIFF
--- a/ext/mysql2/client.c
+++ b/ext/mysql2/client.c
@@ -1260,6 +1260,22 @@ static VALUE rb_mysql_client_encoding(VALUE self) {
 }
 
 /* call-seq:
+ *    client.db
+ *
+ * Returns the currently selected db.
+ */
+static VALUE rb_mysql_client_db(VALUE self) {
+  GET_CLIENT(self);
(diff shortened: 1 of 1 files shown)


richardlarocque:
Sorry, I did not mean to open this against brianmario's repo at this time.  I'll close for now.  Apologies for the noise.

sodabrew:
This looks useful, I'd be glad to review and merge.

richardlarocque:
Thanks, @sodabrew!   I'm glad you think so.

I'm still figuring out how this works by reading source code.  I haven't found a lot of official documentation.

I think this will break if session tracking is turned off.  If so, this method needs a warning label.  I want to test and document it properly.

I'm hoping to come back with a PR that includes better documentation and unit tests in a few days.

PR #1245, opened 2022-02-24
Expose `db` attribute of `MYSQL` client struct

richardlarocque:
Adds an accessor method that returns the current value of the `MSYQL`
client struct's `db` attribute.

The `MYSQL` client struct includes a field `char *db`.  When the
`session_track_schema` setting is enabled, this field will be updated
using information from server-provided "OK" packets, keeping it in sync
as the client switches between databases.

### DIFF
--- a/ext/mysql2/client.c
+++ b/ext/mysql2/client.c
@@ -1332,6 +1332,26 @@ static VALUE rb_mysql_client_encoding(VALUE self) {
 }
 
 /* call-seq:
+ *    client.database
+ *
+ * Returns the currently selected database.
+ *
+ * The result may be stale if `session_track_schema` is disabled.  Read
+ * https://dev.mysql.com/doc/refman/5.7/en/session-state-tracking.html for more
+ * information.
+ */
+static VALUE rb_mysql_client_database(VALUE self) {
+  GET_CLIENT(self);
+
+  char *db = wrapper->client->db;
+  if (!db) {
+    return Qnil;
+  }
+
+  return rb_str_new_cstr(wrapper->client->db);
+}
+
+/* call-seq:
  *    client.automatic_close?
  *
  * @return [Boolean]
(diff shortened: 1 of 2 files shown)


richardlarocque:
I'll admit I'm not very familiar with the MySQL client libraries or Ruby extensions, so noob mistakes are a strong possibility here.

I started working on this when I saw https://github.com/brianmario/mysql2/pull/1092 didn't work the way I expected.  The call to `c.session_track(Mysql2::Client::SESSION_TRACK_SCHEMA)` would return the current schema if it was run immediately after a `USE` statement, but would otherwise return `nil`.  

I dug into the client code on the MySQL side and found this:
https://github.com/mysql/mysql-server/blob/0cd98bdf981583a1cf4cb526581fc16e23bb839b/sql-common/client.c#L929-L942

It's updating the `db` attribute whenever it gets back a `SESSION_TRACK_SCHEMA` change notification.  Unlike the "info" that only lasts until the next "OK" packet, this information sticks around.

Hence this PR to expose the `db` field.

Now, does the MySQL client library *guarantee* that it keeps this `db` field up to date and it's safe to take a dependency on it?  I'm not sure.  I wasn't able to turn up any docs that say so.  I'm just expecting this to work based on my reading of the code.

richardlarocque:
Hi @sodabrew.  This is a friendly reminder of PR.  You mentioned in an [earlier iteration](https://github.com/brianmario/mysql2/pull/1244) that you'd be willing to review.  If you can take a look at this one (now with tests!) I'd appreciate it.

Thanks!

sodabrew:
Thanks for the nudge!

sodabrew:
Catching up on very old PRs! This looks OK, but it should be named `database` to be consistent with the constructor:
https://github.com/brianmario/mysql2/blob/58f8d009a0d107776dd6d24c9906426fe0f7b856/lib/mysql2/client.rb#L168-L169

richardlarocque:
Hi @sodabrew,

Sorry I'm late to notice your comment.  @stefansundin reminded me that this exists and provided a patch that implements your suggestion.

I've updated the PR accordingly.

PR #1267, opened 2022-05-28
Added accessor methods for MYSQL_FIELD char * table and char *db

rbur004:
Code to add methods to result.c to access the query result's array of  MYSQL_FIELDs, to access the MYSQL_FIELD.table and MYSQL_FIELD.db as Ruby Arrays of Strings. 

The code allows introspection of a field's table and database in multi-database and/or multi-table queries, when returning the fields as an :array.  Currently, prior knowledge of the SELECT query fields has been required to distinguish fields of the same name, but from different tables, or when a SELECT is made between a table and an alias to the same table (e.g. FROM atable AS t1, atable AS t2).  The only alternative has been to alias each field, with a distinct name, with AS, to get a flat name space.


rbur004:
If I understand your tests correctly, then the simple test would be replicas of the 'fields' tests.
This would test for the simple case of a single table, single database query.  This is probably good enough.
To get a more comprehensive test of the result.tables, the test database would need a second table added to the test DB. To get a more comprehensive test of the result.dbs, a second test database would need to be added.

e.g. the simple test case, testing table and database names of the fields.
```
context "#tables" do
  let(:test_result) { @client.query("SELECT * FROM mysql2_test ORDER BY id DESC LIMIT 1") }

  it "method should exist" do
    expect(test_result).to respond_to(:tables)
  end

  it "should return an array of field names in proper order" do
    result = @client.query("SELECT id, bit_test, single_bit_test FROM mysql2_test ORDER BY id DESC LIMIT 1")
    expect(result.tables).to eql(%w[mysql2_test mysql2_test mysql2_test])
  end

  it "should return an array of frozen strings" do
    result = @client.query("SELECT * FROM mysql2_test ORDER BY id DESC LIMIT 1")
    result.tables.each do |f|
      expect(f).to be_frozen
    end
  end
end

context "#dbs" do
  let(:test_result) { @client.query("SELECT * FROM mysql2_test ORDER BY id DESC LIMIT 1") }

  it "method should exist" do
    expect(test_result).to respond_to(:dbs)
  end

  it "should return an array of field names in proper order" do
    db = DatabaseCredentials['root']['database']
    result = @client.query( "SELECT id, bit_test, single_bit_test FROM mysql2_test ORDER BY id DESC LIMIT 1" )
    expect(result.dbs).to eql([db,db,db])
  end

  it "should return an array of frozen strings" do
    result = @client.query ""SELECT * FROM mysql2_test ORDER BY id DESC LIMIT 1""
    result.dbs.each do |f|
      expect(f).to be_frozen
    end
  end
end

```

PR #1256, opened 2022-04-19
Suppress errors on RuboCop 0.50.0 & Ruby 3.1+

mishina2228:
This PR suppress errors on RuboCop 0.50.0 & Ruby 3.1+.

RuboCop raises errors in `fedora fedora:rawhide` workflow.
https://github.com/brianmario/mysql2/runs/6070946568?check_suite_focus=true

```
Running RuboCop...
Inspecting 36 files
...............An error occurred while Lint/BlockAlignment cop was inspecting /build/lib/mysql2/console.rb:3:20.
To see the complete backtrace run rubocop -d.
..................An error occurred while Lint/BlockAlignment cop was inspecting /build/tasks/generate.rake:1:0.
To see the complete backtrace run rubocop -d.
...

36 files inspected, no offenses detected

2 errors occurred:
An error occurred while Lint/BlockAlignment cop was inspecting /build/lib/mysql2/console.rb:3:20.
An error occurred while Lint/BlockAlignment cop was inspecting /build/tasks/generate.rake:1:0.
Errors are usually caused by RuboCop bugs.
Please, report your problems to RuboCop's issue tracker.
Mention the following information in the issue report:
0.50.0 (using Parser 2.7.2.0, running on ruby 3.1.1 x86_64-linux)
RuboCop failed!
```

The latest RuboCop (v1.27.0) with Ruby 3.1 does not raise the error.
But this gem still supports Ruby 2.0, so we can't update RuboCop.
That's why I just excluded files that caused the error.

PR #1268, opened 2022-05-31
Upgrade RuboCop to v1.30

mishina2228:
RuboCop v1.30.0 can now parse Ruby 2.0 code.
This PR upgrades RuboCop to v1.30 and also fixes and suppresses some cops.
It was hard to fix all cops, so I fixed only the simple ones.

This PR supersedes #1256.


Gemfile
```diff
@@ -19,7 +19,7 @@ group :test do
   # Downgrade psych because old RuboCop can't use new Psych
   gem 'psych', '< 4.0.0'
```
sodabrew, line 20: The explicit `gem 'psych'` line can now be removed

Gemfile
```diff
@@ -2,11 +2,11 @@ source 'https://rubygems.org'
 
 gemspec
 
-gem 'rake', if Gem::Version.new(RUBY_VERSION) >= Gem::Version.new("2.2")
-              '~> 13.0.1'
-            else
-              '< 13'
-            end
+if Gem::Version.new(RUBY_VERSION) >= Gem::Version.new("2.2")
```
sodabrew, line 5: I think the next release could drop Ruby < 2.2

sodabrew (approved):
Looks great, thanks for doing this! The only specific change I'd like is removing the extra `psych` gem that was pinned lower for older versions of RuboCop.

The other comment about older versions of Rake can be dealt with in a Ruby 2.x or higher pass. There may be some other old syntax or old C extension methods that can be cleaned up as well.

Gemfile
mishina2228: So can we drop Ruby 2.1 from the CI matrix?
https://github.com/brianmario/mysql2/blob/80e0001f87ffd3974835478d2e9b068cc5d37f71/.github/workflows/build.yml#L18-L27

and also CentOS 7 with Ruby 2.0.0?
https://github.com/brianmario/mysql2/blob/80e0001f87ffd3974835478d2e9b068cc5d37f71/.github/workflows/container.yml#L14-L15

PR #1270, opened 2022-06-15
README: Fix streaming MySQL documentation link

64kramsystem:
The previously linked page doesn't exist anymore (and it was also HTTP).

Note that I've pointed to the "latest" URL (without the version in the path), which forwards to the URL of the current latest version of the documentation; in this case:

- "latest" form: https://dev.mysql.com/doc/c-api/en/mysql-use-result.html
- latest version: https://dev.mysql.com/doc/c-api/8.0/en/mysql-use-result.html


64kramsystem:
Closing for lack of communicaiton

PR #1271, opened 2022-06-15
README: Expand section on streaming, about early terminating a result set iteration

64kramsystem:
In the README, it was written that it was necessary to fully iterate a streaming resultset, before executing new queries, which constitutes a problem for error recovery.

I've done some diggin' in the source/documentation, and I believe this is not correct/exact, as it should be possible to invoke `mysql_free_result()` (exposed by the gem as `Mysql::Client#free()`); see [here](https://dev.mysql.com/doc/refman/8.0/en/commands-out-of-sync.html):

> This can happen, for example, if you are using mysql_use_result() and try to execute a new query before you have called mysql_free_result(). It can also happen if you try to execute two queries that return data without calling mysql_use_result() or mysql_store_result() in between.

Note that I've tested on my system, and it works as intended, but I don't know with certainty if there are edge cases (based on the documentation, it seems that there isn't).


64kramsystem:
Closing for lack of communication.

PR #1278, opened 2022-08-19
Add path to support homebrew on arm based macbook

stiak:
The default homebrew path on my Apple M1 Pro is `/opt/homebrew` (as [described here](https://github.com/Homebrew/brew/pull/9117)).  I've added an additional path that should mean I don't have to manually configure mysql2.


sodabrew:
Thank you! I think this could be sorted up above the /usr paths for clarity.

sodabrew:
FYI the test failures are not related to your change. I'll merge this as soon as you have a minute to update!

stiak:
> Thank you! I think this could be sorted up above the /usr paths for clarity.

Thanks for the feedback!  Will do.

stiak:
Let me know if the new location makes more sense.

PR #1284, opened 2022-11-21
Lock on the current Fiber rather than current Thread

casperisfine:
Applications using fiber are able to do concurrent queries from the same thread.

Ref: https://github.com/rails/rails/pull/46519#issuecomment-1321730478

~~NB: I'm not sure when `rb_fiber_current()` was introduced, so I'm waiting for CI to see.~~

### DIFF
--- a/ext/mysql2/client.h
+++ b/ext/mysql2/client.h
@@ -3,7 +3,7 @@
 
 typedef struct {
   VALUE encoding;
-  VALUE active_thread; /* rb_thread_current() or Qnil */
+  VALUE active_fiber; /* rb_fiber_current() or Qnil */
   long server_version;
   int reconnect_enabled;
   unsigned int connect_timeout;
@@ -15,7 +15,6 @@ typedef struct {
   MYSQL *client;
 } mysql_client_wrapper;
 
-void rb_mysql_client_set_active_thread(VALUE self);
 void rb_mysql_set_server_query_flags(MYSQL *client, VALUE result);
 
 #define GET_CLIENT(self) \
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -2,6 +2,8 @@ require 'rspec'
 require 'mysql2'
 require 'timeout'
 require 'yaml'
+require 'fiber'
+
 DatabaseCredentials = YAML.load_file('spec/configuration.yml')
 
 if GC.respond_to?(:verify_compaction_references)
--- a/ext/mysql2/client.c
+++ b/ext/mysql2/client.c
@@ -193,7 +193,7 @@ static void rb_mysql_client_mark(void * wrapper) {
   mysql_client_wrapper * w = wrapper;
   if (w) {
     rb_gc_mark(w->encoding);
-    rb_gc_mark(w->active_thread);
+    rb_gc_mark(w->active_fiber);
   }
 }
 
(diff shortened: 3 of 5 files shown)


casperisfine:
> NB: I'm not sure when `rb_fiber_current()` was introduced, so I'm waiting for CI to see.

Looks like 2.1 has it, no idea for 2.0, but it's not ran on CI.

casperisfine:
> no idea for 2.0

I checked, it has it, I'll remove the conditional.

ioquatix:
This looks reasonable to me but would it be better to just use a mutex?

sodabrew:
This is great! I think it may solve quite a few mystery reports of out of sequence commands despite not crossing threads.

Re: mutex, the gem assumes the client connection pool is handling this. Is that assumption worth revisiting?

spec/spec_helper.rb
```diff
@@ -2,6 +2,8 @@
 require 'mysql2'
 require 'timeout'
 require 'yaml'
+require 'fiber'
```
sodabrew, line 5: It sounds like there's no need to raise the minimum Ruby for this import, but I definitely would be comfortable raising the baseline for this.

ext/mysql2/client.h
```diff
@@ -3,7 +3,7 @@
 
 typedef struct {
   VALUE encoding;
-  VALUE active_thread; /* rb_thread_current() or Qnil */
+  VALUE active_fiber; /* rb_thread_current() or Qnil */
```
methodmissing, line 6: Minor nit: the comment should reflect `rb_fiber_current()`

ioquatix:
Well, it sounds like what you basically have is a poor version of a mutex, since you are tracking a critical section with an `active_whatever`. If you just used Mutex for this, you wouldn't have this PR, because it would already just work. You can bail on locking a mutex by checking `Mutex#locked?` or if you want reentrancy, `Mutex#owned?` (the current execution context has locked it).

casperisfine:
> but would it be better to just use a mutex?

This is essentially the same thing, Ruby's mutex is also just a struct with the current fiber store in. This is essentially an inlined mutex.

We could take it out and use a Mutex on the Ruby side, but that would reduce perf a bit, and expose us to bugs with interrupts and such (we'd need to use `Thead.handle_interrupt`).

So I'd say no.

> I think it may solve quite a few mystery reports of out of sequence commands despite not crossing threads.

Actually I doubt so. Short of people using a fiber scheduler and some weird enumerator usages this shouldn't change much.

However I do think the current protection has a few holes, in several cases we release the lock too early which I think open some small race-condidtion windows that can cause segfaults.

I started working on improving this a bit, but haven't finished yet, I'll try to get back on it: https://github.com/Shopify/mysql2/commit/13b38da3024084e23f78189fd237194a0b588b0d

> Re: mutex, the gem assumes the client connection pool is handling this. Is that assumption worth revisiting?

IMO no. You shouldn't use the connection concurrently, you should use a pool as you say. This check is just here so that if you mess it up we raise and error instead of segfaulting.

spec/spec_helper.rb
casperisfine: Yeah I don't think it's needed.

IMHO 2.0 is really really old and people who are still on it don't deserve time. I now require 2.5 across all my projects. But that is totally independent of this PR.

ext/mysql2/client.h
casperisfine: Thank you. Fixed.

casperisfine:
@sodabrew any changes required for this PR to be mergeable?

ioquatix:
Nice work team.

PR #1292, opened 2022-12-21
CI: Drop allow-failure option in some Ubuntu cases.

junaruga:
As we fixed some of the issues, I think that we can drop the allow-failure option in some cases.

I saw the issue #1194 happened on the latest master's MacOS mysql case too. So, I chnaged the issue number from 1165 to 1194.
https://github.com/brianmario/mysql2/issues/1194#issuecomment-1361760742

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -29,12 +29,10 @@ jobs:
         include:
           # Comment out due to ci/setup.sh stucking.
           # - {os: ubuntu-18.04, ruby: 2.4, db: mariadb10.1}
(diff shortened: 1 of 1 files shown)


Faq:
According this, there is no problem with ruby 3?, Just 2.6?

junaruga:
> According this, there is no problem with ruby 3?, Just 2.6?

Are you talking about the MacOS ruby 2.6 case? I am not sure. Maybe you can test it on your forked repo by changing the "2.6" to "3.0" if you like.

junaruga:
Thanks for checking!

PR #1294, opened 2022-12-22
Print the mysql client version info in a SSL support warning.

junaruga:
I am seeing interesting warnings in CI Fedora cases.

https://github.com/brianmario/mysql2/actions/runs/3758719509/jobs/6387395753#step:4:310

```
/build/lib/mysql2/client.rb:58: warning: Your mysql client library does not support setting ssl_mode; full support comes with 5.7.11.
```

There are 2 possibilities about where the warning comes from. Because the 2 lines below have the same warning message.

* https://github.com/brianmario/mysql2/blob/7f6f33a6e0bd652d5e7087edb6e5d00df248d35b/ext/mysql2/client.c#L125
* https://github.com/brianmario/mysql2/blob/7f6f33a6e0bd652d5e7087edb6e5d00df248d35b/ext/mysql2/client.c#L166

So, I printed the version in the warning. The result was "30207" was printed. 

https://github.com/junaruga/mysql2/actions/runs/3761039661/jobs/6392364841#step:4:401

```
/build/lib/mysql2/client.rb:58: warning: Your mysql client library version 30207 does not support setting ssl_mode; full support comes with 5.7.11.
```

Because the client library [mariadb-connector-c](https://github.com/mariadb-corporation/mariadb-connector-c) version 3.2.7 was used on the environment.

So, seeing the logic to print the warning. I think the condition `version < 50703` is only considered for MySQL and MariaDB (a client library bundled in MariaDB). But not for mariadb-connector-c. Maybe we can fix the condition in another PR.
https://github.com/brianmario/mysql2/blob/7f6f33a6e0bd652d5e7087edb6e5d00df248d35b/ext/mysql2/client.c#L122-L127

I think this PR is helpful to debug around this kind of issue.

## Coding style

Originally the `rb_warn` in line 125 and 166 has spaces inside the "()". However, checking other parts below, I removed the spaces to align other parts.

```
$ grep -r rb_warn ext/
ext/mysql2/client.c:    rb_warn( "Your mysql client library does not support setting ssl_mode; full support comes with 5.7.11." );
ext/mysql2/client.c:    rb_warn( "Your mysql client library does not support ssl_mode %d.", val );
ext/mysql2/client.c:    rb_warn( "Your mysql client library does not support ssl_mode as expected." );
ext/mysql2/client.c:  rb_warn( "Your mysql client library does not support setting ssl_mode; full support comes with 5.7.11." );
ext/mysql2/client.c:        rb_warn("mysql2 failed to invalidate FD safely, closing unsafely\n");
ext/mysql2/client.c:    rb_warn("%s\n", mysql_error(wrapper->client));
ext/mysql2/client.c:    rb_warn("Connections are always closed by garbage collector on Windows");
ext/mysql2/client.c:    rb_warn("%s\n", mysql_error(wrapper->client));
ext/mysql2/result.c:    rb_warn(":cache_rows is ignored if :stream is true");
ext/mysql2/result.c:    rb_warn(":cache_rows is forced for prepared statements (if not streaming)");
ext/mysql2/result.c:    rb_warn(":cast is forced for prepared statements");
ext/mysql2/result.c:      rb_warn(":database_timezone option must be :utc or :local - defaulting to :local");
```

## Commit message

The commit message is below.

It's to know the details of the warning, and to distinguish warnings between the client.c:125 and client.c:166, as these messages are the same. I wanted to know where the warning came from.

### DIFF
--- a/ext/mysql2/client.c
+++ b/ext/mysql2/client.c
@@ -122,7 +122,7 @@ static VALUE rb_set_ssl_mode_option(VALUE self, VALUE setting) {
   unsigned long version = mysql_get_client_version();
 
   if (version < 50703) {
-    rb_warn( "Your mysql client library does not support setting ssl_mode; full support comes with 5.7.11." );
+    rb_warn("Your mysql client library version %lu does not support setting ssl_mode; full support comes with 5.7.11.", version);
     return Qnil;
   }
 #if defined(HAVE_CONST_MYSQL_OPT_SSL_VERIFY_SERVER_CERT) || defined(HAVE_CONST_MYSQL_OPT_SSL_ENFORCE)
@@ -163,7 +163,7 @@ static VALUE rb_set_ssl_mode_option(VALUE self, VALUE setting) {
   return INT2NUM(result);
 #endif
 #ifdef NO_SSL_MODE_SUPPORT
-  rb_warn( "Your mysql client library does not support setting ssl_mode; full support comes with 5.7.11." );
+  rb_warn("Your mysql client library does not support setting ssl_mode; full support comes with 5.7.11.");
   return Qnil;
 #endif
 }
--- a/ext/mysql2/client.c
+++ b/ext/mysql2/client.c
@@ -121,7 +121,7 @@ struct nogvl_select_db_args {
 static VALUE rb_set_ssl_mode_option(VALUE self, VALUE setting) {
   unsigned long version = mysql_get_client_version();
 
-  if (version < 50703) {
+  if (version >= 50000 && version < 50703) {
     rb_warn("Your mysql client library version %lu does not support setting ssl_mode; full support comes with 5.7.11.", version);
     return Qnil;
   }


sodabrew:
I recall there is a similar-looking comparison with a different version family for the mysql-connector-c as well. I've noticed quite a few distributions are switching to MariaDB, so a common situation is MariaDB client libraries and a managed MySQL like AWS RDS MySQL, AWS RDS Aurora, Azure MySQL, etc.

junaruga:
Okay.

Checking the code of the `mysql_get_client_version` in mariadb-connector-c v3.2.7 below, the `MARIADB_PACKAGE_VERSION_ID` is returned.

https://github.com/mariadb-corporation/mariadb-connector-c/blob/497d463d9c647647751a4c609c4bee646ec595f4/libmariadb/mariadb_lib.c#L3989-L3992

https://github.com/mariadb-corporation/mariadb-connector-c/blob/497d463d9c647647751a4c609c4bee646ec595f4/CMakeLists.txt#L41-L43

And checking a history of how the `MARIADB_PACKAGE_VERSION_ID` is defined. Now the mysql-connector-c returns the own version ID. But I assume it returns a different one in the past.

https://github.com/mariadb-corporation/mariadb-connector-c/commit/1ffe387f3c3fb43559d0743b86d316f7e022c571

I think it may be difficult to print a warning by just checking the client library version now.

Can we find an equivalent way to set the condition without using the version?
Can we find a flag to show the type of the client library?

Or just removing the warning?

```
diff --git a/ext/mysql2/client.c b/ext/mysql2/client.c
index c200b36..c09f5fe 100644
--- a/ext/mysql2/client.c
+++ b/ext/mysql2/client.c
@@ -121,10 +121,6 @@ struct nogvl_select_db_args {
 static VALUE rb_set_ssl_mode_option(VALUE self, VALUE setting) {
   unsigned long version = mysql_get_client_version();
 
-  if (version < 50703) {
-    rb_warn( "Your mysql client library does not support setting ssl_mode; full support comes with 5.7.11." );
-    return Qnil;
-  }
 #if defined(HAVE_CONST_MYSQL_OPT_SSL_VERIFY_SERVER_CERT) || defined(HAVE_CONST_MYSQL_OPT_SSL_ENFORCE)
   GET_CLIENT(self);
   int val = NUM2INT( setting );
```

sodabrew:
Do we need a table of what version ranges are in use by which connector libraries?

junaruga:
> Do we need a table of what version ranges are in use by which connector libraries?

Yeah, I think the document of the version ranges is helpful for users.

> Trying to get clear on this - is the issue that MariaDB's connector is returning a value that is lower than 50703 but actually fully supports these SSL features?

Yes. Maybe. I don't see the official document about their SSL support in the mariadb-connector-c. But in the Fedora project, the mysql2 RPM is using the mariadb-connector-c 3.2.7 (30207) to run SSL tests. And I can confirm that it just works.

Right now the mariadb-connector-c latest version is 3.3.3. And I assume that the version number is 30303 that is lower than 50703.
https://github.com/mariadb-corporation/mariadb-connector-c

sodabrew:
So then I'd just tighten this warning to `if version >= 50000 && version < 50703`

junaruga:
OK. That makes sense. I just added the 2nd commit to this PR, to tighten the warning criteria on your way. It's up to you to merge this PR, or cherry-pick the commit or you push your own commit.

junaruga:
Wait, the Fedora case still prints another warning.

https://github.com/brianmario/mysql2/actions/runs/3877871203/jobs/6613392548#step:4:270

```
/build/lib/mysql2/client.rb:58: warning: Your mysql client library does not support ssl_mode as expected.
```

Possibly at https://github.com/brianmario/mysql2/blob/bab4969264250abca836e85a9c332c42be573dfc/ext/mysql2/client.c#L150 .

sodabrew:
I'm looking at the subtle workaround code for 5.7.0 - 5.7.3 and 5.7.3 - 5.7.11, and given that the current 5.7 release is 5.7.40, I am thinking it may be better to simply refuse to compile with client libraries that are so old.

I also found in the docs that MySQL 5.6.36 adds the same SSL support as 5.7.11 and above.

MariaDB appears to have a slightly different set of SSL options. It's been a while since I've dug through the differences between the two. The docs at https://mariadb.com/kb/en/mysql_optionsv/#tls-options look good. Also, I didn't know about `mysql_optionsv` that is specific to MariaDB! Both `mysql_options` and `mysql_optionsv` take the same arguments, except where a MariaDB option requires more than one argument, then the `v` variant is required.

So far I only saw https://mariadb.com/kb/en/mysql_optionsv/#plugin-options `MARIADB_OPT_USERDATA` that requires 4 arguments. BUT LOOK ABOVE IT -- `mysql_optionsv(mysql, MARIADB_OPT_CONNECTION_HANDLER, (void *)"aurora");` that is awesome --> #1025  https://mariadb.com/kb/en/mariadb-connectorc-plugins/

junaruga:
> I'm looking at the subtle workaround code for 5.7.0 - 5.7.3 and 5.7.3 - 5.7.11, and given that the current 5.7 release is 5.7.40, I am thinking it may be better to simply refuse to compile with client libraries that are so old.

OK. I can assume the subtle workaround codes are below.

https://github.com/brianmario/mysql2/blob/a89fd5b5948e03f38d2a15678cc2615b3840777d/ext/mysql2/client.c#L124-L127

https://github.com/brianmario/mysql2/blob/a89fd5b5948e03f38d2a15678cc2615b3840777d/ext/mysql2/client.c#L132-L152

In the terms of maintaining mysql2 RPM package only depending on the MariaDB Connector C relatively new versions in Fedora and the downstream Linux distros, I am okay for mysql2 to refuse to compile with the client libraries that are so old. 

> I also found in the docs that MySQL 5.6.36 adds the same SSL support as 5.7.11 and above.

OK.

> MariaDB appears to have a slightly different set of SSL options. It's been a while since I've dug through the differences between the two. The docs at https://mariadb.com/kb/en/mysql_optionsv/#tls-options look good. Also, I didn't know about mysql_optionsv that is specific to MariaDB! Both mysql_options and mysql_optionsv take the same arguments, except where a MariaDB option requires more than one argument, then the v variant is required.

Oh, good news? The `mysql_optionsv` that can take more than one argument makes the code simple in the MariaDB case. But we need to write MariaDB specific logic?

> So far I only saw https://mariadb.com/kb/en/mysql_optionsv/#plugin-options MARIADB_OPT_USERDATA that requires 4 arguments. BUT LOOK ABOVE IT -- mysql_optionsv(mysql, MARIADB_OPT_CONNECTION_HANDLER, (void *)"aurora"); that is awesome --> https://github.com/brianmario/mysql2/pull/1025 https://mariadb.com/kb/en/mariadb-connectorc-plugins/

Oh, that is a way to connect to a specific DB's client with the plugin? I assume that it makes us easy to implement Aurora specific feature. :)

sodabrew:
I totally forgot the > 50000 condition to isolate the warning to MySQL-range version!

Thanks for pinging this ticket again, figure out some sensible fix and document a version matrix

PR #1293, opened 2022-12-22
Add an option to set a custom SSL pem files directory in test.

junaruga:
In the Fedora project, we are running the mysql2 tests on the build environment with a user permission, without root permission and without `sudo`.

For [the change](https://github.com/brianmario/mysql2/pull/1290) in MariaDB 10.5.18 on Fedora, we need to set up the SSL pem files manually to run the SSL tests.

In this case, we couldn't set up the pem files required to run SSL tests in the `/etc/mysql`. This custom SSL directory option gives an option to run the SSL tests executed in the environment.

How to use:

```
$ TEST_RUBY_MYSQL2_SSL_DIR=/tmp/mysql2 \
  bundle exec rake spec
```

PR #1296, opened 2022-12-23
Add an option to use certifications generated from localhost.

junaruga:
In the Fedora project, we are running the mysql2 tests on the build environment with a user permission, without root permission and without `sudo`.

In this case, we couldn't set up the custom domain "mysql2gem.example.com" required to run SSL tests. This feature to create a set of the certifications from the localhost gives an option to run the SSL tests executed in the environment.

How to generate the certificaton files:

```
$ cd spec/ssl/
$ bash gen_certs.sh
```

The files are generated in the `spec/ssl/<host name>`. Added the new files generated from the domain localhost in the `spec/ssl/localhost`.

How to use:

```
$ TEST_RUBY_MYSQL2_SSL_CERT_HOST=localhost \
  bundle exec rake spec
```

---

The implementation is similar with https://github.com/brianmario/mysql2/pull/1293 .


junaruga:
Rebasing this PR on the latest master branch. However, the https://github.com/brianmario/mysql2/pull/1293 is a higher priority for me. :)

sodabrew:
Since #1293 landed, is this PR still needed to build on it or close it out?

junaruga:
> Since https://github.com/brianmario/mysql2/pull/1293 landed, is this PR still needed to build on it or close it out?

I think this PR is still needed. I rebased it on the latest master branch.

In Fedora's building environment, we cannot edit the /etc/hosts file to add the "mysql2gem.example.com". So, in the `spec/mysql2/client_spec.rb`, I want to set the host used to create the cert files by given environment variable, while currently the "mysql2gem.example.com" is hard-coded. What do you think?

sodabrew:
Closing as substantially superseded — #1293 landed the configurability this PR was built around: `spec_helper.rb`'s `ssl_cert_dir`/`ssl_cert_host` already read `TEST_RUBY_MYSQL2_SSL_CERT_DIR`/`TEST_RUBY_MYSQL2_SSL_CERT_HOST` on current master, and `ci/ssl.sh` already respects the custom cert dir. That's the mechanism this PR needed to build on, and it's already there.

What's still genuinely unique here is the pre-generated `spec/ssl/localhost/` cert fixtures, so restricted build environments (Fedora's sandboxed builders, as described) don't need to run `gen_certs.sh` themselves. If that convenience is still wanted, a fresh, much smaller PR against current master adding just those fixture files would be easy to review and merge — this branch is from December 2022 and would need a full rebase against three years of changes to get there.

Thanks for the original diagnosis and for pushing #1293 forward — appreciate the persistence on this.

Issue #1297, opened 2022-12-23
CI MacOS: failing by upgrading Python from 3.10.8 to 3.10.9


junaruga:
Right now the CI mariadb and mysql cases on MacOS are failing.
https://github.com/brianmario/mysql2/actions/runs/3767036141/jobs/6404171930#step:7:281

```
==> Upgrading python@3.10
  3.10.8 -> 3.10.9 

==> Pouring python@3.10--3.10.9.monterey.bottle.tar.gz
Error: The `brew link` step did not complete successfully
The formula built, but is not symlinked into /usr/local
Could not symlink bin/2to3
Target /usr/local/bin/2to3
already exists. You may want to remove it:
  rm '/usr/local/bin/2to3'

To force the link and overwrite all conflicting files:
  brew link --overwrite python@3.10

To list all files that would be deleted:
  brew link --overwrite --dry-run python@3.10
```

Maybe here is the issue ticket: https://github.com/actions/setup-python/issues/577 .

sodabrew:
Since resolved

PR #1299, opened 2022-12-28
CI: GitHub Actions MacOS images now include openssl

sodabrew:
Reduces CI time and sidesteps the failure at #1297


sodabrew:
Bah, fails for the MySQL on MacOS build with
```
ld: warning: directory not found for option '-L/usr/local/opt/openssl@3/lib'
```

junaruga:
I debugged this behavior. The CI log is [here](https://github.com/junaruga/mysql2/actions/runs/3795035451/jobs/6453728470#step:6:1).

Currently the `extconf.rb gets the openssl library path from `brew --prefix openssl`.

https://github.com/brianmario/mysql2/blob/bab4969264250abca836e85a9c332c42be573dfc/ext/mysql2/extconf.rb#L31

However the behavior of openssl un-versioned package name gets the openssl version 3 on the OSX image.

```
$ brew --prefix openssl
/usr/local/opt/openssl@3
```
But the actual directory exists for only openssl 1.1.

```
$ ls -ld /usr/local/opt/openssl*
lrwxr-xr-x  1 runner  admin  36 Dec 16 00:42 /usr/local/opt/openssl -> /usr/local/Cellar/openssl@1.1/1.1.1s
lrwxr-xr-x  1 runner  admin  28 Dec 16 00:25 /usr/local/opt/openssl@1.1 -> ../Cellar/openssl@1.1/1.1.1s
```

And this is intended behavior by the openssl brew package maintainer https://github.com/Homebrew/brew/issues/12879 , and he suggested using the `brew --prefix openssl@1.1` instead of that.

```
$ brew --prefix openssl@1.1
/usr/local/opt/openssl@1.1
```

sodabrew:
IIRC, the libmysqlclient can be built against openssl 1.1 or 3, so the question isn't as much "what's installed on this system?" as it is "what matches the client library we'll be linking with / loading at runtime?"

junaruga:
Yes, I understand your point. The mysql2 should work with both openssl 1.1 and 3 if we can link the runtime library. I meant we could access to the openssl 1.1 library to make mysql2 work when we don't install additional openssl on the GitHub Actions macos case.

junaruga:
How do you fix this issue to pass MacOS cases?

junaruga:
I opened another PR https://github.com/brianmario/mysql2/pull/1301 based on the first 2 commits on this PR.

junaruga:
Maybe shall we change this PR's title "CI: GitHub Actions MacOS images now include openssl" to another one that describes what the PR do in a better way? I suppose that the "CI: GitHub Actions MacOS images now include openssl" was fixed by https://github.com/brianmario/mysql2/pull/1303 .

junaruga:
Have you considered the unified YAML file for Ubuntu cases? The `ubuntu-focal.yml` and `ubuntu-jammy.yml` are similar. 

```
$ diff ubuntu-focal.yml ubuntu-jammy.yml
1c1
< name: Ubuntu Focal 20.04 with Ruby 2.x
---
> name: Ubuntu Jammy 22.04 with Ruby 3.x
12c12
<           - ubuntu-20.04 # focal
---
>           - ubuntu-22.04 # jammy
14,21c14,17
<           - '2.7'
<           - '2.6'
<           - '2.5'
<           - '2.4'
<           - '2.3'
<           - '2.2'
<           - '2.1'
<           - '2.0'
---
>           - 'head'
>           - '3.2'
>           - '3.1'
>           - '3.0'
23,25c19,21
<           - {os: ubuntu-20.04, ruby: '2.7', db: mariadb10.6}
<           - {os: ubuntu-20.04, ruby: '2.7', db: mysql57}
<           - {os: ubuntu-20.04, ruby: '2.7', db: mysql80}
---
>           - {os: ubuntu-22.04, ruby: '3.0', db: mariadb10.6}
>           - {os: ubuntu-22.04, ruby: '3.0', db: mysql57}
>           - {os: ubuntu-22.04, ruby: '3.0', db: mysql80}
```

I tired the unified one `ubuntu.yml` on my forked repository on this PR. Then some cases failed with the network connection error. The [CI result](https://github.com/junaruga/mysql2/actions/runs/3969720675/jobs/6804588280#step:6:88).

ubuntu.yml
```yaml
name: Ubuntu
on: [push, pull_request]
jobs:
  build:
    name: >-
      ${{ matrix.os }} ruby ${{ matrix.ruby }} ${{ matrix.db }}
    runs-on: ${{ matrix.os }}
    continue-on-error: ${{ matrix.allow-failure || false }}
    strategy:
      matrix:
        include:
          # jammy
          - {os: ubuntu-22.04, ruby: 'head'}
          - {os: ubuntu-22.04, ruby: '3.2'}
          - {os: ubuntu-22.04, ruby: '3.1'}
          - {os: ubuntu-22.04, ruby: '3.0'}
          - {os: ubuntu-22.04, ruby: '3.0', db: mariadb10.6}
          - {os: ubuntu-22.04, ruby: '3.0', db: mysql57}
          - {os: ubuntu-22.04, ruby: '3.0', db: mysql80}
          # focal
          - {os: ubuntu-20.04, ruby: '2.7'}
          - {os: ubuntu-20.04, ruby: '2.6'}
          - {os: ubuntu-20.04, ruby: '2.5'}
          - {os: ubuntu-20.04, ruby: '2.4'}
          - {os: ubuntu-20.04, ruby: '2.3'}
          - {os: ubuntu-20.04, ruby: '2.2'}
          - {os: ubuntu-20.04, ruby: '2.1'}
          - {os: ubuntu-20.04, ruby: '2.0'}
          - {os: ubuntu-20.04, ruby: '2.7', db: mariadb10.6}
          - {os: ubuntu-20.04, ruby: '2.7', db: mysql57}
          - {os: ubuntu-20.04, ruby: '2.7', db: mysql80}
      # On the fail-fast: true, it cancels all in-progress jobs
      # if any matrix job fails, which we don't want.
      fail-fast: false
    env:
      BUNDLE_WITHOUT: development
    steps:
      - uses: actions/checkout@v3
      # https://github.com/ruby/setup-ruby
      - uses: ruby/setup-ruby@v1
        with:
          ruby-version: ${{ matrix.ruby }}
          bundler-cache: true # runs 'bundle install' and caches installed gems automatically
      - if: matrix.db
        run: echo 'DB=${{ matrix.db }}' >> $GITHUB_ENV
      - run: sudo echo "127.0.0.1 mysql2gem.example.com" | sudo tee -a /etc/hosts
      - run: bash ci/setup.sh
      - run: bundle exec rake spec
```

.github/workflows/ubuntu-focal.yml
```diff
@@ -0,0 +1,42 @@
+name: Ubuntu Focal 20.04 with Ruby 2.x
```
junaruga, line 1: The value of the "name" element is also used as the badge image's text. Maybe shorter text is better. See <https://github.com/brianmario/mysql2/actions/workflows/ubuntu-focal.yml/badge.svg>.

junaruga:
Thanks for the work! Just note that when we see the network connection error, we can use a retry bash script that is like [this](https://github.com/ruby/ruby/blob/master/tool/travis_retry.sh) used in the ruby/ruby.

PR #1303, opened 2023-01-19
Add an OpenSSL directory configuration option

sodabrew:
CI: Use on MacOS to pick up the correct OpenSSL version. Updates #1301 and #1299

cc @junaruga

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -26,42 +26,41 @@ jobs:
           - '2.3'
           - '2.2'
           - '2.1'
-        db: ['']
         include:
           # Comment out due to ci/setup.sh stucking.
           # - {os: ubuntu-18.04, ruby: 2.4, db: mariadb10.1}
           - {os: ubuntu-20.04, ruby: '2.4', db: mariadb10.3}
           - {os: ubuntu-18.04, ruby: '2.4', db: mysql57}
           - {os: ubuntu-20.04, ruby: '2.4', db: mysql80}
-          - {os: ubuntu-18.04, ruby: 'head', db: ''}
+          - {os: ubuntu-18.04, ruby: 'head'}
           # db: A DB's brew package name in macOS case.
           #     Set a name "db: 'name@X.Y'" when using an old version.
           # MariaDB lastet version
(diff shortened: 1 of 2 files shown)


sodabrew:
Need one more edit here, which is that `dir_config('openssl')` should only activate if specified on the command line, otherwise assume the defaults are OK.

junaruga:
Thanks for the PR! I see this PR is better than mine. Just I have a small proposal in my quick review. With the modification below, I was able to pass the tests without MacOS specific condition: `which brew`.

```
diff --git a/.github/workflows/build.yml b/.github/workflows/build.yml
index b76f91a..b224242 100644
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -27,7 +27,6 @@ jobs:
           - '2.2'
           - '2.1'
         db: ['']
-        ssl: ['']
         include:
           # Comment out due to ci/setup.sh stucking.
           # - {os: ubuntu-18.04, ruby: 2.4, db: mariadb10.1}
@@ -63,4 +62,6 @@ jobs:
         run: echo 'DB=${{ matrix.db }}' >> $GITHUB_ENV
       - run: sudo echo "127.0.0.1 mysql2gem.example.com" | sudo tee -a /etc/hosts
       - run: bash ci/setup.sh
-      - run: bundle exec rake spec -- $([ -x "$(which brew)" ] && echo --with-openssl-dir=$(brew --prefix ${{ matrix.ssl }}))
+      - run: echo "rake_spec_opts=\"--with-openssl-dir=$(brew --prefix ${{ matrix.ssl }})\"" >> $GITHUB_ENV
+        if: matrix.ssl
+      - run: bundle exec rake spec -- $rake_spec_opts
```

sodabrew:
I made a change in the second to last commit, if you rebase your test branch you'll see `if with_config('openssl')` changes to `if with_config('openssl-dir')`

junaruga:
OK! After rebasng my test branch, I was able to pass the [CI](https://github.com/junaruga/mysql2/actions/runs/3962191688/jobs/6788538080#step:8:10) with the modification above. 😄

PR #1306, opened 2023-01-22
Update README and improve handling of ssl_mode settings

### LINKED ISSUE #1305: Suppressing warnings in MariaDB Connector C (Fedora) case
Seeing the Fedora cases CI log on the latest master branch (6528137e2f3a743c0afb38abb4fbe1de1cbc9695), there are some warnings printed in the process of running the test. I would like to suppress the warnings.

https://github.com/brianmario/mysql2/actions/runs/3972990412/jobs/6811349042#step:4:283
```
+ bundle exec rake spec
...
 /build/lib/mysql2/client.rb:58: warning: Your mysql client library does not support setting ssl_mode; full support comes with 5.6.36+, 5.7.11+, 8.0+
    should set ssl_mode option verify_identity
/build/lib/mysql2/client.rb:58: warning: Your mysql client library does not support setting ssl_mode; full support comes with 5.6.36+, 5.7.11+, 8.0+
    should set ssl_mode option required
/build/lib/mysql2/client.rb:58: warning: Your mysql client library does not support setting ssl_mode; full support comes with 5.6.36+, 5.7.11+, 8.0+
    should set ssl_mode option verify_ca
    should be able to connect via SSL options
/build/lib/mysql2/client.rb:58: warning: Your mysql client library does not support setting ssl_mode; full support comes with 5.6.36+, 5.7.11+, 8.0+
    should set ssl_mode option disabled
/build/lib/mysql2/client.rb:58: warning: Your mysql client library does not support setting ssl_mode; full support comes with 5.6.36+, 5.7.11+, 8.0+
...
   #each
/build/spec/mysql2/statement_spec.rb:311: warning: :cache_rows is ignored if :stream is true
/build/spec/mysql2/statement_spec.rb:311: warning: :cache_rows is ignored if :stream is true
     should yield different value for #first if streaming
...
```

The issue was discussed at https://github.com/brianmario/mysql2/pull/1294, however I want to open the issue ticket to manage the status.

sodabrew:
Print the client version number in SSL warning.
Add comments to explain the ssl mode setting function.
Add MariaDB Connector/C 3.x to the relevant SSL mode code path.

Supersedes #1294
Resolves #1305

@junaruga please review!

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,6 +65,11 @@ This may be needed if you deploy to a system where these libraries
 are located somewhere different than on your build system.
 This overrides any rpath calculated by default or by the options above.
 
+* `--with-openssl-dir[=/path/to/openssl]` - Specify the directory where OpenSSL
+is installed. In most cases, the Ruby runtime and MySQL client libraries will
+link against a system-installed OpenSSL library and this option is not needed.
+Use this option when non-default library paths are needed.
+
 * `--with-sanitize[=address,cfi,integer,memory,thread,undefined]` -
 Enable sanitizers for Clang / GCC. If no argument is given, try to enable
 all sanitizers or fail if none are available. If a command-separated list of
@@ -89,12 +94,47 @@ the library file `libmysqlclient.so` but is missing the header file `mysql.h`
 
 ### Mac OS X
 
-You may use MacPorts, Homebrew, or a native MySQL installer package. The most
+You may use Homebew, MacPorts, or a native MySQL installer package. The most
 common paths will be automatically searched. If you want to select a specific
 MySQL directory, use the `--with-mysql-dir` or `--with-mysql-config` options above.
 
 If you have not done so already, you will need to install the XCode select tools by running
 `xcode-select --install`.
+
+Later versions of MacOS no longer distribute a linkable OpenSSL library. It is
+common to use Homebrew or MacPorts to install OpenSSL. Make sure that both the
+Ruby runtime and MySQL client libraries are compiled with the same OpenSSL
+family, 1.0 or 1.1 or 3.0, since only one can be loaded at runtime.
+
+``` sh
+$ brew install openssl@1.1
+$ gem install mysql2 -- --with-openssl-dir=$(brew --prefix openssl@1.1)
+
+or
+
+$ sudo port install openssl11
+```
+
+Since most Ruby projects use Bundler, you can set build options in the Bundler
+config rather than manually installing a global mysql2 gem. This example shows
+how to set build arguments with [Bundler config](https://bundler.io/man/bundle-config.1.html):
+
+``` sh
+$ bundle config --local build.mysql2 -- --with-openssl-dir=$(brew --prefix openssl@1.1)
+```
+
+Another helpful trick is to use the same OpenSSL library that your Ruby was
+built with, if it was built with an alternate OpenSSL path. This example finds
+the argument `--with-openssl-dir=/some/path` from the Ruby build and adds that
+to the [Bundler config](https://bundler.io/man/bundle-config.1.html):
+
+``` sh
+$ bundle config --local build.mysql2 -- $(ruby -r rbconfig -e 'puts RbConfig::CONFIG["configure_args"]' | xargs -n1 | grep with-openssl-dir)
+```
+
+Note the additional double dashes (`--`) these separate command-line arguments
+that `gem` or `bundler` interpret from the addiitonal arguments that are passed
+to the mysql2 build process.
 
 ### Windows
 
@@ -205,7 +245,7 @@ result = statement.execute(1, "CA", :as => :array)
 
 Session Tracking information can be accessed with
 
-```ruby
+``` ruby
 c = Mysql2::Client.new(
   host: "127.0.0.1",
   username: "root",
@@ -261,19 +301,13 @@ type of connection to make, with special interpretation you should be aware of:
 * An IPv4 or IPv6 address will result in a TCP connection.
 * Any other value will be looked up as a hostname for a TCP connection.
 
-### SSL options
+### SSL/TLS options
 
-Setting any of the following options will enable an SSL connection, but only if
-your MySQL client library and server have been compiled with SSL support.
-MySQL client library defaults will be used for any parameters that are left out
-or set to nil. Relative paths are allowed, and may be required by managed
-hosting providers such as Heroku.
-
-For MySQL versions 5.7.11 and higher, use `:ssl_mode` to prefer or require an
-SSL connection and certificate validation. For earlier versions of MySQL, use
-the `:sslverify` boolean. For details on each of the `:ssl_mode` options, see
-[https://dev.mysql.com/doc/refman/8.0/en/connection-options.html](https://dev.mysql.com/doc/refman/8.0/en/connection-options.html#option_general_ssl-mode).
-
+Setting any of the following options will enable an SSL/TLS connection, but
+only if your MySQL client library and server have been compiled with SSL
+support. MySQL client library defaults will be used for any parameters that are
+left out or set to nil. Relative paths are allowed, and may be required by
+managed hosting providers such as Heroku.
 
 ``` ruby
 Mysql2::Client.new(
@@ -284,9 +318,27 @@ Mysql2::Client.new(
   :sslcapath => '/path/to/cacerts',
   :sslcipher => 'DHE-RSA-AES256-SHA',
   :sslverify => true, # Removed in MySQL 8.0
-  :ssl_mode = :disabled / :preferred / :required / :verify_ca / :verify_identity,
+  :ssl_mode => :disabled / :preferred / :required / :verify_ca / :verify_identity,
   )
 ```
+
+For MySQL versions 5.7.11 and higher, use `:ssl_mode` to prefer or require an
+SSL connection and certificate validation. For earlier versions of MySQL, use
+the `:sslverify` boolean. For details on each of the `:ssl_mode` options, see
+[https://dev.mysql.com/doc/refman/8.0/en/connection-options.html](https://dev.mysql.com/doc/refman/8.0/en/connection-options.html#option_general_ssl-mode).
+
+The `:ssl_mode` option will also set the appropriate MariaDB connection flags:
+
+| `:ssl_mode`        | MariaDB option value                 |
+| ---                | ---                                  |
+| `:disabled`        | MYSQL_OPT_SSL_ENFORCE = 0            |
+| `:required`        | MYSQL_OPT_SSL_ENFORCE = 1            |
+| `:verify_identity` | MYSQL_OPT_SSL_VERIFY_SERVER_CERT = 1 |
+
+MariaDB does not support the `:preferred` or `:verify_ca` options. For more
+information about SSL/TLS in MariaDB, see
+[https://mariadb.com/kb/en/securing-connections-for-client-and-server/](https://mariadb.com/kb/en/securing-connections-for-client-and-server/)
+and [https://mariadb.com/kb/en/mysql_optionsv/#tls-options](https://mariadb.com/kb/en/mysql_optionsv/#tls-options)
 
 ### Secure auth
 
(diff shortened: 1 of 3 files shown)


junaruga:
Thanks for the PR! I am happy to review!

ext/mysql2/client.c
```diff
@@ -121,52 +121,78 @@ struct nogvl_select_db_args {
...
-  if (version < 50630 || (version >= 50700 && version < 50703)) {
-    rb_warn( "Your mysql client library does not support setting ssl_mode; full support comes with 5.6.36+, 5.7.11+, 8.0+" );
+  /* Warn about versions that are known to be incomplete; these are pretty
+   * ancient, we want people to upgrade if they need SSL/TLS to work
+   *
+   * MySQL 5.x before 5.6.30 -- ssl_mode introduced but not fully working until 5.6.36)
+   * MySQL 5.7 before 5.7.3 -- ssl_mode introduced but not fully working until 5.7.11)
+   */
+  if ((version >= 50000 && version < 50630) || (version >= 50700 && version < 50703)) {
+    rb_warn("Your mysql client library version %lu does not support setting ssl_mode; full support comes with 5.6.36+, 5.7.11+, 8.0+", version );
```
junaruga, line 131: It's small issue. It's better to trim the space in the `version );`

ext/mysql2/client.c
```diff
@@ -121,52 +121,78 @@ struct nogvl_select_db_args {
...
+      int result = mysql_options(wrapper->client, MYSQL_OPT_SSL_ENFORCE, &b);
       return INT2NUM(result);
     }
 #endif
-    rb_warn( "Your mysql client library does not support ssl_mode %d.", val );
+    rb_warn("Your mysql client library does not support ssl_mode %d", val);
     return Qnil;
   } else {
-    rb_warn( "Your mysql client library does not support ssl_mode as expected." );
+    rb_warn("Your mysql client library does not support ssl_mode as expected");
```
junaruga, line 168: Can we add the MySQL client library version to these 2 warnings too like this?

```
@@ -162,10 +162,10 @@ static VALUE rb_set_ssl_mode_option(VALUE self, VALUE setting) {
       return INT2NUM(result);
     }
 #endif
-    rb_warn("Your mysql client library does not support ssl_mode %d", val);
+    rb_warn("Your mysql client library version %lu does not support ssl_mode %d", version, val);
     return Qnil;
   } else {
-    rb_warn("Your mysql client library does not support ssl_mode as expected");
+    rb_warn("Your mysql client library version %lu does not support ssl_mode as expected", version);
     return Qnil;
   }
 #endif
```

ext/mysql2/client.c
```diff
@@ -121,52 +121,78 @@ struct nogvl_select_db_args {
...
     if (val == SSL_MODE_DISABLED || val == SSL_MODE_REQUIRED) {
-      my_bool b = ( val == SSL_MODE_REQUIRED );
-      int result = mysql_options( wrapper->client, MYSQL_OPT_SSL_ENFORCE, &b );
+      my_bool b = (val == SSL_MODE_REQUIRED);
+      int result = mysql_options(wrapper->client, MYSQL_OPT_SSL_ENFORCE, &b);
       return INT2NUM(result);
     }
 #endif
-    rb_warn( "Your mysql client library does not support ssl_mode %d.", val );
+    rb_warn("Your mysql client library does not support ssl_mode %d", val);
```
junaruga, line 165: Seeing this PR's CI fedora case, I still see the warning below.

https://github.com/brianmario/mysql2/actions/runs/3977427296/jobs/6818560696#step:4:502
> /build/lib/mysql2/client.rb:58: warning: Your mysql client library does not support ssl_mode 0

This is because MariaDB Connector C doesn't support the ssl_mode: disabled (= 0), right? 

https://github.com/brianmario/mysql2/blob/6528137e2f3a743c0afb38abb4fbe1de1cbc9695/spec/mysql2/client_spec.rb#L169

Perhaps, we may be able to enhance the unit tests with each ssl mode and the MySQL client library versions.
In the cases of printing warnings, we can use RSpec's [output-matcher](https://relishapp.com/rspec/rspec-expectations/docs/built-in-matchers/output-matcher) to capture the warning message and suppress it in the result of the unit test.

README.md
```diff
@@ -284,10 +318,27 @@ Mysql2::Client.new(
...
+
+| `:ssl_mode`        | MariaDB option value                 |
+| ---                | ---                                  |
+| `:disabled`        | MYSQL_OPT_SSL_ENFORCE = 0            |
+| `:required`        | MYSQL_OPT_SSL_ENFORCE = 1            |
+| `:verify_identity` | MYSQL_OPT_SSL_VERIFY_SERVER_CERT = 1 |
+
+MariaDB does not support the `:preferred` or `:verify_ca` options. For more information SSL/TLS in MariaDB, see
+[https://mariadb.com/kb/en/securing-connections-for-client-and-server/](https://mariadb.com/kb/en/securing-connections-for-client-and-server/)
+and [https://mariadb.com/kb/en/mysql_optionsv/#tls-options](https://mariadb.com/kb/en/mysql_optionsv/#tls-options)
```
junaruga, line 340: We are printing the values of the ssl_mode in the warning message below. Maybe it's helpful to have a document about the mapping table of ssl_mode and the value or column for that on this table, such as `:disabled` | `0` | `MYSQL_OPT_SSL_ENFORCE = 0`

```
rb_warn("Your mysql client library does not support ssl_mode %d", val);
```

ext/mysql2/client.c
```diff
@@ -121,52 +121,78 @@ struct nogvl_select_db_args {
...
-  if ((version >= 50703 && version < 50711) || (version >= 60103 && version < 60200) || version >= 100000) {
+  int val = NUM2INT(setting);
+
+  /* Expected code path for MariaDB 10.x and MariaDB Connector/C 3.x
+   * Workaround code path for MySQL 5.7.3 - 5.7.10 and MySQL Connector/C 6.1.3 - 6.1.x
+   */
+  if (version >= 100000                         // MariaDB (all versions numbered 10.x)
+    || (version >= 30000 && version < 40000)    // MariaDB Connector/C (all versions numbered 3.x)
+    || (version >= 50703 && version < 50711)    // Workaround for MySQL 5.7.3 - 5.7.10
+    || (version >= 60103 && version < 60200)) { // Workaround for MySQL Connector/C 6.1.3 - 6.1.x
```
junaruga, line 150: We print the client library version in warning messages. So, what do you think about adding a MySQL/MariaDB client and the version number mapping table to the README like this?

```
| MySQL/MariaDB client | Version |
| ---                  | ---                                  |
| MariaDB (all versions numbered 10.x) | version >= 10000     |
| MariaDB Connector/C (all versions numbered 3.x) | version >= 30000 && version < 40000 |
...
```

README.md
```diff
@@ -89,13 +94,48 @@ the library file `libmysqlclient.so` but is missing the header file `mysql.h`
...
+
+``` sh
+$ bundle config --local build.mysql2 -- --with-openssl-dir=$(brew --prefix openssl@1.1)
+```
+
+Another helpful trick is to use the same OpenSSL library that your Ruby was
+built with. This example finds the argument `--with-openssl-dir=/some/path`
+from the Ruby build and adds that to the Bundler config:
+
+
```
junaruga, line 130: Small issue. There are 2 blank lines. You can trim one blank line.

README.md
```diff
@@ -89,13 +94,48 @@ the library file `libmysqlclient.so` but is missing the header file `mysql.h`
...
+config rather than manually installing a global mysql2 gem. This example shows
+how to set Bundler build arguments:
+
+``` sh
+$ bundle config --local build.mysql2 -- --with-openssl-dir=$(brew --prefix openssl@1.1)
+```
+
+Another helpful trick is to use the same OpenSSL library that your Ruby was
+built with. This example finds the argument `--with-openssl-dir=/some/path`
+from the Ruby build and adds that to the Bundler config:
```
junaruga, line 128: I think it's helpful to explain this example works only "if the Ruby is configured with the `--with-openssl-dir=/some/path`". I tried the example below for my built Ruby on local, and I couldn't find the `--with-openssl-dir=/some/path`, because maybe the Ruby was not configured with `--with-openssl-dir=/some/path`.

```
$ which ruby
/usr/local/ruby-3.2.0-preview2/bin/ruby

$ ruby -r rbconfig -e 'puts RbConfig::CONFIG["configure_args"]' | xargs -n1
--prefix=/usr/local/ruby-3.2.0-preview2
--enable-shared
```

README.md
```diff
@@ -89,13 +94,48 @@ the library file `libmysqlclient.so` but is missing the header file `mysql.h`
...
+
+$ sudo port install openssl11
+```
+
+Since most Ruby projects use Bundler, you can set build options in the Bundler
+config rather than manually installing a global mysql2 gem. This example shows
+how to set Bundler build arguments:
+
+``` sh
+$ bundle config --local build.mysql2 -- --with-openssl-dir=$(brew --prefix openssl@1.1)
```
junaruga, line 123: I didn't know the tip, and tried it on my local :)

How about adding the sentence like this?

"See `man 1 bundle-config` - BUILD OPTIONS for details."

Or the link <https://bundler.io/man/bundle-config.1.html> - BUILD OPTIONS.

junaruga:
I reviewed and commented!

ext/mysql2/client.c
```diff
@@ -121,52 +121,78 @@ struct nogvl_select_db_args {
...
-  int val = NUM2INT( setting );
-  // Either MySQL 5.7.3 - 5.7.10, or Connector/C 6.1.3 - 6.1.x, or MariaDB 10.x and later
-  if ((version >= 50703 && version < 50711) || (version >= 60103 && version < 60200) || version >= 100000) {
+  int val = NUM2INT(setting);
+
+  /* Expected code path for MariaDB 10.x and MariaDB Connector/C 3.x
+   * Workaround code path for MySQL 5.7.3 - 5.7.10 and MySQL Connector/C 6.1.3 - 6.1.x
+   */
+  if (version >= 100000                         // MariaDB (all versions numbered 10.x)
+    || (version >= 30000 && version < 40000)    // MariaDB Connector/C (all versions numbered 3.x)
```
junaruga, line 148: I just double-checked the following definitions are available in the MariaDB connector C case.

* https://mariadb.com/kb/en/configuring-mariadb-connectorc-with-option-files/#ssl-verify-server-cert - MYSQL_OPT_SSL_VERIFY_SERVER_CERT - Introduced: MariaDB Connector/C 3.0.0 
* https://mariadb.com/kb/en/configuring-mariadb-connectorc-with-option-files/#ssl-enforce - MYSQL_OPT_SSL_ENFORCE - Introduced: MariaDB Connector/C 3.1.1

README.md
sodabrew: I think the awkward message is really because the translation from Ruby atom to C constant is in Ruby code while the mode setting action is in C code. The C code has the error but doesn't know what the original Ruby atom was.

A deeper cleanup might be moving the whole method into Ruby space, but I think it would be better get in the way of good enough for now.

ext/mysql2/client.c
sodabrew: Yes, that could work:
```
$ irb -r mysql2
irb> Mysql2::Client.info
=> {:id=>80031, :version=>"8.0.31", :header_version=>"8.0.31"}
```

junaruga: OK. The enhancement of the unit tests could be perhaps in the future. And what is the context of what you showed the client info on the IRB?

README.md
junaruga: OK. I agree with it!

ext/mysql2/client.c
sodabrew: I think the ifdefs will handle this as is, at the risk of someone compiling against 3.0.x but running against 3.1.x or vice versa.

ext/mysql2/client.c
sodabrew: I'll switch to the printable version numbers instead!

ext/mysql2/client.c
junaruga: I think so. My comment above is not about the issue or not about asking to change the code. I just checked and noted here that the 2 definitions are available for the MariaDB Connector C version range: `version >= 30000 && version < 40000`.

PR #1323, opened 2023-08-02
Support utf8mb3 charset naming for MySQL 8 and MariaDB 10.6

### LINKED ISSUE #1315: Mysql2::Error: Unsupported charset: '"utf8mb3"'
Hello,

after upgrading to Mariadb 10.6, I've encountered this issue:

```
Mysql2::Error: Unsupported charset: '"utf8mb3"'
/var/www/example_s/shared/bundle/ruby/3.2.0/gems/mysql2-0.5.5/lib/mysql2/client.rb:49:in `charset_name='
/var/www/example_s/shared/bundle/ruby/3.2.0/gems/mysql2-0.5.5/lib/mysql2/client.rb:49:in `initialize'
```

I think the issue is due to the fact that `utf8mb3` is not in `mysql2_mysql_enc_name_to_rb_map` (https://github.com/brianmario/mysql2/blob/b439a895ef6b289e1bc5e07303fc3952713fb948/ext/mysql2/mysql_enc_name_to_ruby.h#L99).

MariaDB changed the default `utf8` to be named `utf8mb3` in 10.6 (https://mariadb.com/kb/en/unicode/) and MySQL will changed it in `MySQL 8.0.28` (https://dev.mysql.com/doc/refman/8.0/en/charset-unicode-utf8mb3.html).
To be future compatible, it would be nice to use `utf8mb3` as encoding.

Unfortunately I feel not be able to create a good PR for it. It would be nice if someone could add this in the future.

Thanks!

jeremy:
MySQL and MariaDB have long aliased the `utf8` charset to the underlying `utf8mb3`. They both [switched to report the underlying `utf8mb3` charset name](https://dev.mysql.com/doc/refman/8.0/en/charset-unicode-utf8mb3.html) instead of the `utf8` alias name, allowing `utf8` to target `utf8mb4` in the future.

That means we need to explicitly map `utf8mb3` for folks running newer MySQL/MariaDB with older `utf8` dbs.

Closes #1315


sodabrew:
Failures fixed by #1324, thanks for both!

PR #1325, opened 2023-08-03
CI: Update MariaDB 10.11 package name

sodabrew:


jeremy:
Jinx! Covered this and the defunct 10.6 repo URL in #1326

PR #1327, opened 2023-08-03
CI: GitHub dropped Ubuntu 18.04 runners; switch to 20.04 for MySQL 5.7 test

jeremy:
https://github.blog/changelog/2022-08-09-github-actions-the-ubuntu-18-04-actions-runner-image-is-being-deprecated-and-will-be-removed-by-12-1-22/

We'll need to switch to running MySQL in Docker to restore 5.7 test coverage. (Maybe be simpler than doing package installs, anyway?)


sodabrew:
The CentOS builds are dockerized already, that may be a decent template? Or use side-by-side containers, so MySQL is accessed over the network from the build/test container

sodabrew:
Superseded by #1371

PR #1339, opened 2024-01-02
Fix warning on ruby 3.3 related to bigdecimal

VitaliySerov:
This will fix this warning for `ruby-3.3` projects that require `mysql2`:

```
/opt/hostedtoolcache/Ruby/3.3.0/x64/lib/ruby/gems/3.3.0/gems/mysql2-0.5.5/lib/mysql2.rb:2: warning: bigdecimal was loaded from the standard library, but will no longer be part of the default gems since Ruby 3.4.0. Add bigdecimal to your Gemfile or gemspec. Also contact author of mysql2-0.5.5 to add bigdecimal into its gemspec.
```

### DIFF
--- a/ci/mariadb1011.sh
+++ b/ci/mariadb1011.sh
@@ -6,5 +6,5 @@ rm -fr /etc/mysql
 rm -fr /var/lib/mysql
 
 apt-key add support/C74CD1D8.asc
-add-apt-repository "deb https://downloads.mariadb.com/MariaDB/mariadb-10.11/repo/ubuntu $(lsb_release -cs) main"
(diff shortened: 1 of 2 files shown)


sodabrew:
Hmm, this is probably the right simple thing to do. Big decimal isn't necessarily required by all schemas, so I could imagine a user wanting to disable it.

Did Ruby proper add new native support for large decimals?

sodabrew:
Oh, could you add an explicit Ruby 3.3 to the CI matrix while you're here? It would confirm this works.

VitaliySerov:
@sodabrew Yeah ok, I will create a separate PR for 3.3 first 

But seems here we got some troubles with version resolving, need to investigate and understand why simple change broke full CI

PR #1341, opened 2024-01-05
Fix faling CI

VitaliySerov:
1. Ubuntu-based CI failing because of old keys
2. Fedora-based CI is failing because of missing dependency

### DIFF
--- a/ci/mysql57.sh
+++ b/ci/mysql57.sh
@@ -7,6 +7,7 @@ rm -fr /etc/mysql
 rm -fr /var/lib/mysql
 apt-key add support/5072E1F5.asc # old signing key
 apt-key add support/3A79BD29.asc # 5.7.37 and higher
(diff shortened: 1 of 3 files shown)


flavorjones (approved):

ci/mysql57.sh
```diff
@@ -7,6 +7,7 @@ rm -fr /etc/mysql
 rm -fr /var/lib/mysql
 apt-key add support/5072E1F5.asc # old signing key
 apt-key add support/3A79BD29.asc # 5.7.37 and higher
+apt-key adv --keyserver keyserver.ubuntu.com --recv-keys B7B3B788A8D3785C
```
sodabrew, line 10: For the other keys, I checked the public key into the repo - this may be archaic knowledge, but at least a few times I recall CI failures because the old keys didn't fetch.

Issue #1346, opened 2024-01-30
mysql2 does not build with mysql-client 8.3.0


xjunior:
Compiling mysql2 with `mysql-client` 8.3.0 fails:

```bash
❯ make
compiling client.c
In file included from client.c:15:
./mysql_enc_name_to_ruby.h:43:1: warning: a function definition without a prototype is deprecated in all versions of C and is not supported in C2x [-Wdeprecated-non-prototype]
mysql2_mysql_enc_name_to_rb_hash (str, len)
^
./mysql_enc_name_to_ruby.h:86:1: warning: a function definition without a prototype is deprecated in all versions of C and is not supported in C2x [-Wdeprecated-non-prototype]
mysql2_mysql_enc_name_to_rb (str, len)
^
client.c:1438:3: error: call to undeclared function 'mysql_ssl_set'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
  mysql_ssl_set(wrapper->client,
  ^
client.c:1438:3: note: did you mean 'mysql_close'?
/opt/homebrew/opt/mysql-client/include/mysql/mysql.h:797:14: note: 'mysql_close' declared here
void STDCALL mysql_close(MYSQL *sock);
             ^
2 warnings and 1 error generated.
make: *** [client.o] Error 1
```

`mysql_ssl_set` was [deprecated on 8.2 and removed on 8.3](https://dev.mysql.com/doc/relnotes/mysql/8.3/en/news-8-3-0.html).

xjunior:
If you're currently having issues on a mac and homebrew, this is what you should do to get your environment back:

Uninstall current installed version of mysql-client:

```bash
$ brew uninstall mysql-client
```

Download 8.2 formulae:

```bash
$ wget https://raw.githubusercontent.com/Homebrew/homebrew-core/2f35529519fb6a2cc361ce3d464a1bd181505a54/Formula/m/mysql-client.rb
```

Install 8.2:

```bash
$ brew install ./mysql-client.rb
```

yboulkaid:
For homebrew users, there's also a pinned formula available at 8.0:
```
brew install mysql-client@8.0
```

xjunior:
@yboulkaid 8.0 still didn't do it for me, I don't quite remember why, but if it works for anyone, it should be better than downloading that formulae :)

noahgregory-basis:
Ultimately, the fix shouldn't be to downgrade, the maintainers of this gem should fix their code.

[...]

PR #1347, opened 2024-01-30
Replace mysql_ssl_set by mysql_options

xjunior:
Fixes https://github.com/brianmario/mysql2/issues/1346


AlexandreHauber (approved):

mikelkew (approved):

fmichaut-diff (approved):

Faq:
To keep compatibility with older versions, perhaps an if to check sql version?

xjunior:
@Faq I don't think it would work, as the compiler would require the symbols to be available at compile/linking time, unless this is done in compile time. Since I'm not a C expert, I defer that to the maintainers to either take on this work, or provide a strong suggestion on how they'd like to see this implemented.

xjunior:
Closing in favor of https://github.com/brianmario/mysql2/pull/1352#pullrequestreview-1869180636

Issue #1263, opened 2022-05-17
Using AWS IAM database authentication


rohittak16:
I am using [IAM database authentication](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/UsingWithRDS.IAMDBAuth.html)

To connect with database using IAM role works by creating a token via AWS API which is valid for 15 mins, but rails database.yml gets cached and will not be evaluated again for a new connection(which may fail after 15 min). 
What would be the best way to get IAM database authentication working?

floor114:
@rohittak16 Have you found a working solution? I’m trying to add IAM database authentication as well.

floor114:
@showwin I made a gem for that some time ago. Check it here https://github.com/floor114/mysql2-aws_rds_iam

sodabrew:
This is achievable today, no mysql2 changes needed — the pattern @showwin posted above is the right shape: generate the token per-connection with `aws-sdk-rds`, pass it as `:password`, and use `ssl_mode: :verify_identity` + `:sslca` pointed at Amazon's [RDS CA bundle](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/UsingWithRDS.SSL.html) (IAM auth requires TLS). @floor114's [mysql2-aws_rds_iam](https://github.com/floor114/mysql2-aws_rds_iam) gem packages exactly this.

The friction the original report describes — Rails' `database.yml`-driven connection config being evaluated once and cached, so a 15-minute-lived token goes stale — is a Rails/ActiveRecord connection-pool concern, not something mysql2 itself can fix; mysql2 already regenerates the token correctly whenever `Mysql2::Client.new` actually runs, which is what the monkey-patch above hooks into.

The one gap that *would* belong in mysql2: right now getting a fresh token into a *reconnect* (`:reconnect => true`, or any retry after a dropped connection) still requires patching `Client.new` the way this thread does, since `:password` is a plain string evaluated once at connect time. #1531 proposes accepting a callable `:password` (a `Proc`/lambda re-evaluated on every connect, reconnect included) — that would make this pattern first-class without the monkeypatch. Tracking further work there; closing this one as answered.

PR #1370, opened 2024-06-04
Dynamic Password Provider

davidsiaw:
Provide the ability to provide a password using a specified class, like if you want to use RDS IAM Auth

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -34,7 +34,7 @@ jobs:
           - {os: ubuntu-22.04, ruby: '3.0', db: mariadb10.6}
           - {os: ubuntu-20.04, ruby: '2.7', db: mariadb10.6}
           - {os: ubuntu-20.04, ruby: '2.7', db: mysql80}
-          - {os: ubuntu-18.04, ruby: '2.7', db: mysql57}
+          # - {os: ubuntu-18.04, ruby: '2.7', db: mysql57} # no longer exists
 
           # TODO - Windows CI
           # - {os: windows-2022, ruby: '3.2', db: mysql80}
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -45,7 +45,7 @@ Metrics/BlockNesting:
 # Offense count: 1
 # Configuration parameters: CountComments, CountAsOne.
 Metrics/ClassLength:
-  Max: 135
+  Max: 141
 
 # Offense count: 3
 # Configuration parameters: IgnoredMethods.
--- a/README.md
+++ b/README.md
@@ -107,7 +107,7 @@ Ruby runtime and MySQL client libraries are compiled with the same OpenSSL
 family, 1.0 or 1.1 or 3.0, since only one can be loaded at runtime.
 
 ``` sh
-$ brew install openssl@1.1
+$ brew install openssl@1.1 zstd
 $ gem install mysql2 -- --with-openssl-dir=$(brew --prefix openssl@1.1)
 
 or
--- a/ci/setup.sh
+++ b/ci/setup.sh
@@ -70,7 +70,7 @@ if [[ x$OSTYPE =~ ^xdarwin ]]; then
   done
 
   brew info "$DB"
-  brew install "$DB"
+  brew install "$DB" zstd
   DB_PREFIX="$(brew --prefix "${DB}")"
   export PATH="${DB_PREFIX}/bin:${PATH}"
   export LDFLAGS="-L${DB_PREFIX}/lib"
--- a/ext/mysql2/extconf.rb
+++ b/ext/mysql2/extconf.rb
@@ -108,7 +108,7 @@ if inc && lib
   warn "-----\nUsing --with-mysql-dir=#{File.dirname inc}\n-----"
   rpath_dir = lib
   have_library('mysqlclient')
-elsif (mc = (with_config('mysql-config') || Dir[GLOB].first))
+elsif (mc = with_config('mysql-config') || Dir[GLOB].first)
   # If the user has provided a --with-mysql-config argument, we must respect it or fail.
   # If the user gave --with-mysql-config with no argument means we should try to find it.
   mc = Dir[GLOB].first if mc == true
--- a/lib/mysql2/client.rb
+++ b/lib/mysql2/client.rb
@@ -98,7 +98,8 @@ module Mysql2
     end
 
     def extract_password(opts)
-      return (opts[:password] || opts[:pass]) if !opts[:password_provider]
+      return opts[:password] || opts[:pass] unless opts[:password_provider]
+
       klass = Kernel.const_get(opts[:password_provider])
 
       obj = klass.new(opts)
--- a/lib/mysql2/client.rb
+++ b/lib/mysql2/client.rb
@@ -79,7 +79,7 @@ module Mysql2
       end
 
       user     = opts[:username] || opts[:user]
-      pass     = opts[:password] || opts[:pass]
+      pass     = extract_password(opts)
       host     = opts[:host] || opts[:hostname]
       port     = opts[:port]
       database = opts[:database] || opts[:dbname] || opts[:db]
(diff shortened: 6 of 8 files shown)


sodabrew:
Catching up here, I like the approach and would be glad to land this PR. Do you have time to update it?

sodabrew:
## Investigation summary: what this should actually become

Dug into the mechanism here, compared it against how `floor114/mysql2-aws_rds_iam` (the companion gem someone built for the exact motivating use case, AWS RDS IAM auth, referenced in #1263) solves the same problem, and looked at what mysql2 itself could sensibly own vs. what should stay in vendor-specific satellite gems. Writing this up in full so whoever picks up implementation doesn't have to re-derive it.

**Bottom line recommendation: don't land `:password_provider` as written. Replace it with "`:password` may be a callable, invoked fresh at every real connect, and mysql2 refuses to combine it with `reconnect: true`."** That's a smaller, safer surface that solves the part of this problem an application genuinely can't solve for itself.

### 1. This PR's branch is stale and its diff is noisy

5 of the 7 changed files (`.github/workflows/build.yml`, `.rubocop_todo.yml`, the README's `zstd` line, `extconf.rb`'s paren removal, `mysql2.gemspec`'s `bigdecimal` dependency) are just artifacts of `master` having moved on since June 2024 -- those changes already exist independently on `master` now. The only intentional change is in `lib/mysql2/client.rb`:

```ruby
def extract_password(opts)
  return opts[:password] || opts[:pass] unless opts[:password_provider]

  klass = Kernel.const_get(opts[:password_provider])
  obj = klass.new(opts)
  obj.password
end
```

Whatever lands should be a fresh branch off current `master`, not a rebase of this one.

### 2. `Kernel.const_get` on a hash value is a real security concern, not a style nit

`opts[:password_provider]` is resolved via `Kernel.const_get` and the resulting class is instantiated with the *entire* opts hash, return value used unchecked. If `Mysql2::Client.new(opts)` config ever flows through anything less trusted than static app code -- a per-tenant config table, a YAML file merged from multiple sources -- this is an arbitrary-class-instantiation primitive: the string picks *which class* gets `.new(opts)` called on it, with no constraint that it's actually a password provider. Same shape of risk as `YAML.unsafe_load`/`Marshal.load` on untrusted input. This alone is reason enough not to merge the mechanism as written, independent of the functional gaps below.

**Fix: accept a `Class`, or better, a `Proc`/anything responding to `#call`, never a `String`/`Symbol` looked up via `const_get`.**

### 3. What the companion gem actually does (full source, not just the README)

`floor114/mysql2-aws_rds_iam` ([source](https://github.com/floor114/mysql2-aws_rds_iam)) works via `Module#prepend`:

```ruby
# Mysql2::AwsRdsIam.apply_patch
const.prepend(ClientExtension)

# ClientExtension#initialize
def initialize(opts = {})
  opts = opts.dup
  if opts.delete(:aws_rds_iam_auth)
    raise ReconnectConfigEnabledError if opts[:reconnect]     # <-- load-bearing
    raise UsernameNotFoundError if opts[:username].nil?
    raise HostNotFoundError if opts[:host].nil?

    opts.delete(:password)
    opts[:password] = AuthToken::Factory.call(opts.delete(:aws_rds_iam_auth_token_generator), opts[:host], opts[:port], opts[:username])
    opts[:enable_cleartext_plugin] = true
  end
  super(opts)
end
```

Three load-bearing details:

1. **`raise ... if opts[:reconnect]`** -- explicitly forbids combining this with `reconnect: true`. This is not defensive-programming caution, it's structurally necessary (see #4 below).
2. **`enable_cleartext_plugin = true`, forced** -- AWS's IAM token is a long opaque bearer string, not a normal password; it has to go over the wire via MySQL's cleartext auth plugin (over TLS), not the default password-hashing exchange. This is AWS-specific plumbing.
3. A **named registry** of token-generator strategies (`AuthToken::Registry#fetch(:default)` or a custom one) -- pure userland Ruby, unrelated to the mysql2-integration question.

**`:password_provider` as written in this PR has none of #1 or #2, and does #3 differently (unsafe `const_get` vs. a registry lookup).** It's not just a smaller feature than the companion gem -- it's missing the one piece of correctness the companion gem's author clearly had to learn the hard way.

### 4. Why `reconnect: true` is fundamentally incompatible with any dynamic-password mechanism, and mysql2 is the only layer that can enforce it

Verified directly against current `ext/mysql2/client.c`: `reconnect: true` maps straight to libmysqlclient's native `MYSQL_OPT_RECONNECT` (`client.c` around `wrapper->reconnect_enabled` / `mysql_options(..., MYSQL_OPT_RECONNECT, ...)`). When that fires, it happens **transparently inside a blocking C call** (`mysql_ping()` or the next `mysql_real_query()`), reusing whatever credentials were captured at the *original* `mysql_real_connect()` -- entirely inside the C client library, with **no callback into Ruby**.

This means:
- A short-lived credential (AWS IAM tokens: 15 min; same shape for Azure AD tokens, GCP Cloud SQL IAM tokens, Vault dynamic DB credentials) can never be safely refreshed mid-reconnect if `MYSQL_OPT_RECONNECT` is what's doing the reconnecting. There is no hook to give it a new password once C has taken over.
- The only correct place to reconnect with a dynamic credential is a **new** `Mysql2::Client.new` call at the Ruby level (which re-runs `initialize` and can fetch a fresh credential) -- i.e. connection-pool-driven reconnection, not libmysqlclient's transparent one.
- Only mysql2 itself can see both sides of this (whether a password is dynamic, and whether reconnect is transparent-C vs. Ruby-level) and enforce the incompatibility centrally. An app-level wrapper has to reimplement this guard for every vendor/auth-scheme gem that wants to do this correctly -- which is exactly the check this PR is currently missing.

### 5. Proposed design

- **`opts[:password]` (and `:pass`) may be a callable** (`Proc`/lambda, or any object responding to `#call`) in addition to a `String`. No `const_get`, no class-name lookup, no injection surface.
- **Invoke it fresh at every point mysql2 is about to run a real `mysql_real_connect`** -- i.e. every `Client.new` / explicit reconnect at the Ruby level. This is "free" correctness: a connection pool that builds a new `Client` to replace a dead one automatically gets a fresh credential, no extra code required on the caller's part.
- **Raise (`ArgumentError`, at `Client.new` time) if a callable password is combined with `reconnect: true`.** Turns the companion gem's manually-authored, easy-to-forget guard into something mysql2 enforces for *any* dynamic-credential scheme, not just AWS's.
- **Redaction:** mysql2 already redacts `:password` from `Client#inspect` and from the query-options hash it exposes elsewhere (see the earlier "Redact password from query_options" fix). A callable password needs the same treatment -- never `#call`ed during inspection, never leaked if it's a closure capturing credentials.
- **Explicitly out of scope for mysql2 core:** AWS/Azure/GCP/Vault SDK calls, `enable_cleartext_plugin`-style vendor-specific connection-option requirements. Those stay in small satellite gems built on top of the generic callable hook -- mysql2 should not take on an `aws-sdk-rds`-shaped dependency or hardcode auth-plugin behavior that's wrong for other providers.

### 6. Concrete effect on the companion gem, for scale

With this design, `mysql2-aws_rds_iam` shrinks from a multi-file `prepend`-based wrapper (module-existence checks, `instance_method` checks, a prepended `initialize` override, custom error classes for the reconnect guard) down to roughly:

```ruby
module Mysql2::AwsRdsIam
  def self.password_for(host:, port:, username:, generator: :default)
    -> { AuthToken::Factory.call(generator, host, port, username) }
  end
end

# usage -- no monkeypatching, no custom flag, no prepend:
Mysql2::Client.new(
  host: host, username: username,
  password: Mysql2::AwsRdsIam.password_for(host: host, port: port, username: username),
  enable_cleartext_plugin: true,
  reconnect: false, # mysql2 now enforces this is required anyway
)
```

The `prepend`/flag-interception machinery disappears entirely; what's left is just the AWS SDK call itself, which is exactly the part that should stay outside mysql2.

### Suggested implementation shape

- `ext/mysql2/client.c` / `lib/mysql2/client.rb`: wherever `:password`/`:pass` is currently read as a string and handed toward `connect`, check `respond_to?(:call)` first and invoke it; validate the result is a `String` (clear `ArgumentError` if not, rather than an opaque native connect failure).
- Add the `reconnect: true` + callable-password incompatibility check at the same point options are validated (mirrors how other option conflicts are already validated in this file).
- Extend the existing password-redaction logic (`Client#inspect`, query-options exposure) to recognize a callable password and redact it the same way.
- New regression specs: callable password used for a normal connect; callable password + `reconnect: true` raises; redaction doesn't invoke or leak the callable.
- README: new subsection near "Connection options" documenting the callable-password mechanism, with a short recipe resembling the companion-gem example above.

This is a design/API decision more than a large diff -- happy to discuss the exact shape (e.g. `:password_provider` as a distinct option that must respond to `#call` vs. just widening what `:password` itself accepts) before implementation starts.

PR #1371, opened 2024-06-04
Fix CI

davidsiaw:
I would like to propose these updates to the CI to handle:
- deprecation of old ubuntu image 18.04 in github CI (github simply ignores requests for it now so they are always stuck yellow)
- we now need zstd it seems on MacOS
- add bigdecimal because it is removed from ruby 3.4


.github/workflows/build.yml
```diff
@@ -34,7 +34,7 @@ jobs:
           - {os: ubuntu-22.04, ruby: '3.0', db: mariadb10.6}
           - {os: ubuntu-20.04, ruby: '2.7', db: mariadb10.6}
           - {os: ubuntu-20.04, ruby: '2.7', db: mysql80}
-          - {os: ubuntu-18.04, ruby: '2.7', db: mysql57}
+          # - {os: ubuntu-18.04, ruby: '2.7', db: mysql57} # no longer exists
```
sodabrew, line 37:
```suggestion
          # - {os: ubuntu-20.04, ruby: '2.7', db: mysql57} # no longer exists
```
To continue testing MySQL 5.7 let's see if it'll run on 20.04

.github/workflows/build.yml
```diff
@@ -34,7 +34,7 @@ jobs:
           - {os: ubuntu-22.04, ruby: '3.0', db: mariadb10.6}
           - {os: ubuntu-20.04, ruby: '2.7', db: mariadb10.6}
           - {os: ubuntu-20.04, ruby: '2.7', db: mysql80}
-          - {os: ubuntu-18.04, ruby: '2.7', db: mysql57}
+          # - {os: ubuntu-20.04, ruby: '2.7', db: mysql57} # no longer exists
```
sodabrew, line 37:
```suggestion
          - {os: ubuntu-20.04, ruby: '2.7', db: mysql57}
```

PR #1375, opened 2024-07-18
Add vector type added in MySQL 9.0

dbussink:
This type was added in MySQL 9.0, so ensure we can parse it here. See also:

https://dev.mysql.com/doc/dev/mysql-server/latest/field__types_8h.html


simi:
centos7 CI seems broken, is it worth to keep around?

dbussink:
> centos7 CI seems broken, is it worth to keep around?

Ah yeah, looks like it's EOL.

PR #1377, opened 2024-08-31
Support get_server_public_key option 

samitani:
# What

This PR fixes below MySQL authentication error with SSL-disabled connection.
```
Mysql2::Error: Authentication plugin 'caching_sha2_password' reported error: 
Authentication requires secure connection.
```

# Background

This error only happen in case of first authentication under cleartext TCP connection. `caching_sha2_password` authentication requires a secure connection at first authentication for each users. The first authentication means that the cache is not ready yet at server-side. The reason why secure connection is required is that MySQL server need to know user password to create cache.  After the cache is created, MySQL server does not request secure connection anymore. Please refer to the MySQL manual to know details.

> For each user account, the first client connection for the user after any of the following operations must use a secure connection (made using TCP using TLS credentials, a Unix socket file, or shared memory) or RSA key pair-based password exchange: 
https://dev.mysql.com/doc/refman/8.0/en/caching-sha2-pluggable-authentication.html

`libmysqlclient` provides MYSQL_OPT_GET_SERVER_PUBLIC_KEY option which enables clients to create a requested secure connection automatically during authentication. This PR carries it to mysql2.

# Sample code

```
require 'mysql2'

port     = 3306
hostname = 'hostname'
username = 'user'
password = 'password'

# connect with SSL to issue FLUSH PRIVILEGES
client = Mysql2::Client.new(:host => hostname, :port => port, :username => username, :password => password, :ssl_mode => :required)

# force clear cache
client.query("FLUSH PRIVILEGES")

print "# without get_server_public_key\n"
# this returns Authentication requires secure connection.
begin
  client = Mysql2::Client.new(:host => hostname, :port => port, :username => username, :password => password, :ssl_mode => :disabled)
rescue => e
  printf("connect ng %s\n", e)
end

print "# with get_server_public_key\n"
begin
  client = Mysql2::Client.new(:host => hostname, :port => port, :username => username, :password => password, :ssl_mode => :disabled, :get_server_public_key => true)
  print "connect ok\n"
rescue => e
  printf("connect ng %s\n", e)
end
```

```
$ ruby ../test2.rb
# without get_server_public_key
connect ng Authentication plugin 'caching_sha2_password' reported error: Authentication requires secure connection.
# with get_server_public_key
connect ok
```

### DIFF
--- a/ext/mysql2/client.c
+++ b/ext/mysql2/client.c
@@ -996,6 +996,13 @@ static VALUE _mysql_client_options(VALUE self, int opt, VALUE value) {
       retval  = charval;
       break;
 
+#ifdef HAVE_CONST_MYSQL_OPT_GET_SERVER_PUBLIC_KEY
+    case MYSQL_OPT_GET_SERVER_PUBLIC_KEY:
+      boolval = (value == Qfalse ? 0 : 1);
+      retval = &boolval;
+      break;
+#endif
+
 #ifdef HAVE_MYSQL_DEFAULT_AUTH
     case MYSQL_DEFAULT_AUTH:
       charval = (const char *)StringValueCStr(value);
@@ -1485,6 +1492,14 @@ static VALUE set_init_command(VALUE self, VALUE value) {
   return _mysql_client_options(self, MYSQL_INIT_COMMAND, value);
 }
 
+static VALUE set_get_server_public_key(VALUE self, VALUE value) {
+#ifdef HAVE_CONST_MYSQL_OPT_GET_SERVER_PUBLIC_KEY
+  return _mysql_client_options(self, MYSQL_OPT_GET_SERVER_PUBLIC_KEY, value);
+#else
+  return Qfalse;
+#endif
+}
+
 static VALUE set_default_auth(VALUE self, VALUE value) {
 #ifdef HAVE_MYSQL_DEFAULT_AUTH
   return _mysql_client_options(self, MYSQL_DEFAULT_AUTH, value);
@@ -1596,6 +1611,7 @@ void init_mysql2_client() {
   rb_define_private_method(cMysql2Client, "default_file=", set_read_default_file, 1);
   rb_define_private_method(cMysql2Client, "default_group=", set_read_default_group, 1);
   rb_define_private_method(cMysql2Client, "init_command=", set_init_command, 1);
+  rb_define_private_method(cMysql2Client, "get_server_public_key=", set_get_server_public_key, 1);
   rb_define_private_method(cMysql2Client, "default_auth=", set_default_auth, 1);
   rb_define_private_method(cMysql2Client, "ssl_set", set_ssl_options, 5);
   rb_define_private_method(cMysql2Client, "ssl_mode=", rb_set_ssl_mode_option, 1);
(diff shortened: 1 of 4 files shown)


ext/mysql2/client.c
```diff
@@ -1596,6 +1611,7 @@ void init_mysql2_client() {
   rb_define_private_method(cMysql2Client, "default_file=", set_read_default_file, 1);
   rb_define_private_method(cMysql2Client, "default_group=", set_read_default_group, 1);
   rb_define_private_method(cMysql2Client, "init_command=", set_init_command, 1);
+  rb_define_private_method(cMysql2Client, "get_server_public_key=", set_get_server_public_key, 1);
```
sodabrew, line 1614: The method name `set_get_...` seems extraneous? Can it be plain `get_server_public_key`?

ext/mysql2/extconf.rb
```diff
@@ -159,6 +159,7 @@ def add_ssl_defines(header)
 have_const('SERVER_QUERY_WAS_SLOW', mysql_h)
 have_const('MYSQL_OPTION_MULTI_STATEMENTS_ON', mysql_h)
 have_const('MYSQL_OPTION_MULTI_STATEMENTS_OFF', mysql_h)
+have_const('MYSQL_OPT_GET_SERVER_PUBLIC_KEY', mysql_h)
```
sodabrew, line 162: It's not necessary to test for this constant into its own HAVE_ constant, because we can directly test for the constant at time of use (see suggested change)

ext/mysql2/client.c
```diff
@@ -1485,6 +1492,14 @@ static VALUE set_init_command(VALUE self, VALUE value) {
   return _mysql_client_options(self, MYSQL_INIT_COMMAND, value);
 }
 
+static VALUE set_get_server_public_key(VALUE self, VALUE value) {
+#ifdef HAVE_CONST_MYSQL_OPT_GET_SERVER_PUBLIC_KEY
```
sodabrew, line 1496:
```suggestion
#ifdef MYSQL_OPT_GET_SERVER_PUBLIC_KEY
```

ext/mysql2/client.c
```diff
@@ -996,6 +996,13 @@ static VALUE _mysql_client_options(VALUE self, int opt, VALUE value) {
       retval  = charval;
       break;
 
+#ifdef HAVE_CONST_MYSQL_OPT_GET_SERVER_PUBLIC_KEY
```
sodabrew, line 999:
```suggestion
#ifdef MYSQL_OPT_GET_SERVER_PUBLIC_KEY
```

sodabrew (dismissed):

ext/mysql2/extconf.rb
samitani: @sodabrew Hello, Thank you for reviewing my PR.
`MYSQL_OPT_GET_SERVER_PUBLIC_KEY` is defined as `enum`. If my understand correctly, `ifdef` does not work with `enum`.
https://github.com/mysql/mysql-server/blob/8.0/include/mysql.h.pp#L411-L448

sodabrew: You’re absolutely right!

sodabrew:
Thank you for the PR!

PR #1383, opened 2024-12-02
Save affected_rows on the wrapper when reading results

byroot:
Between fetching the result and accessing the affected_rows property, GC might have been triggered and might have freed some Mysql2::Statement objects. This calls mysql_stmt_close which resets the connection affected_rows to -1, which in turn is treated as an error when calling mysql_affected_rows.

```
client.query("SELECT 1")
client.affected_rows # raises Mysql2::Error
```

cc @matthewd @zzak (This is what is causing the flaky test on Rails).

@tenderlove You think you could merge and release this? This bugs make prepare statements fundamentally unsafe to use....

### DIFF
--- a/ext/mysql2/client.c
+++ b/ext/mysql2/client.c
@@ -411,6 +411,7 @@ static VALUE allocate(VALUE klass) {
   wrapper->initialized = 0; /* will be set true after calling mysql_init */
   wrapper->closed = 1; /* will be set false after calling mysql_real_connect */
   wrapper->refcount = 1;
+  wrapper->affected_rows = -1;
   wrapper->client = (MYSQL*)xmalloc(sizeof(MYSQL));
 
   return obj;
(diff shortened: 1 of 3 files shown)


sodabrew (approved):
This is great.

ext/mysql2/client.c
```diff
@@ -1155,12 +1157,12 @@ static VALUE rb_mysql_client_session_track(VALUE self, VALUE type) {
  * if it was an UPDATE, DELETE, or INSERT.
  */
 static VALUE rb_mysql_client_affected_rows(VALUE self) {
-  my_ulonglong retVal;
+  uint64_t retVal;
```
sodabrew, line 1160: Ok, I see the type changed between MySQL 5.7 and 8.0:
https://dev.mysql.com/doc/c-api/8.0/en/mysql-affected-rows.html
```
uint64_t
mysql_affected_rows(MYSQL *mysql)
```
https://dev.mysql.com/doc/c-api/5.7/en/mysql-affected-rows.html
```
my_ulonglong
mysql_affected_rows(MYSQL *mysql)
```

PR #1384, opened 2024-12-02
extconf.rb: locate zstd when using homebrew

byroot:
Fix compilation issues on my system.

cc @tenderlove

### DIFF
--- a/ext/mysql2/extconf.rb
+++ b/ext/mysql2/extconf.rb
@@ -63,7 +63,12 @@ if with_config('openssl-dir')
 # Homebrew OpenSSL on MacOS
 elsif RUBY_PLATFORM =~ /darwin/ && system('command -v brew')
   openssl_location = `brew --prefix openssl`.strip
-  $LDFLAGS << " -L#{openssl_location}/lib" if openssl_location
+  $LIBPATH << "#{openssl_location}/lib" unless openssl_location.empty?
+end
+
+if RUBY_PLATFORM =~ /darwin/ && system('command -v brew')
+  zstd_location = `brew --prefix zstd`.strip
+  $LIBPATH << "#{zstd_location}/lib" unless zstd_location.empty?
 end
 
 ### Find MySQL client library


ext/mysql2/extconf.rb
```diff
@@ -66,6 +66,11 @@ def add_ssl_defines(header)
   $LDFLAGS << " -L#{openssl_location}/lib" if openssl_location
 end
 
+if RUBY_PLATFORM =~ /darwin/ && system('command -v brew')
+  zstd_location = `brew --prefix zstd`.strip
+  $LIBPATH << "#{zstd_location}/lib"
```
sodabrew, line 71: If zstd is not installed (perhaps for an older version of mysqlclient) I think `$LIBPATH` should not be modified. How about this:
```suggestion
  $LIBPATH << "#{zstd_location}/lib" if zstd_location
```

ext/mysql2/extconf.rb
```diff
@@ -66,6 +66,12 @@ def add_ssl_defines(header)
   $LDFLAGS << " -L#{openssl_location}/lib" if openssl_location
 end
 
+if RUBY_PLATFORM =~ /darwin/ && system('command -v brew')
+  zstd_location = `brew --prefix zstd`.strip
+
+  $LIBPATH << "#{zstd_location}/lib" unless zstd_location.empty?
```
sodabrew, line 72: If you found `.empty?` necessary vs. falsiness of the empty string, then the `if openssl_location` clause above should change to match -- did you test to see if brew's output when the package is not installed is in fact "truthy"?

```suggestion
  $LIBPATH << "-L#{zstd_location}/lib" unless zstd_location.empty?
```
Also I notice the openssl pattern above adds `-L` -- I think these ought to match? But if I'm missing something let me know!

byroot: > vs. falsiness of the empty string

Empty strings aren't falsy in Ruby.

> Also I notice the openssl pattern above adds -L -- I think these ought to match? 

They're not adding to the same variables. `$LIBPATH` takes paths, `$LDFLAGS` takes flags. I guess I can rewrite the OpenSSL one to use LIBPATH and properly do its error checks.

byroot: Done.

sodabrew: 🤦 quite right. So what's that line above testing? Seems like a bogon, unless there's some long-lost mystery behavior of yore (old code, after all)

sodabrew (approved):
Thank you! Looking consistent and sensible now

ext/mysql2/extconf.rb
byroot: It wasn't testing anything. Was likely just an oversight.

ext/mysql2/extconf.rb
```diff
@@ -63,7 +63,13 @@ def add_ssl_defines(header)
 # Homebrew OpenSSL on MacOS
 elsif RUBY_PLATFORM =~ /darwin/ && system('command -v brew')
   openssl_location = `brew --prefix openssl`.strip
-  $LDFLAGS << " -L#{openssl_location}/lib" if openssl_location
+  $LIBPATH << "#{openssl_location}/lib" unless openssl_location.empty?
+end
+
+if RUBY_PLATFORM =~ /darwin/ && system('command -v brew')
+  zstd_location = `brew --prefix zstd`.strip
+
```
sodabrew, line 71:
```suggestion
```

PR #1385, opened 2024-12-03
Improve prepared statement ergonomics

byroot:
While working on Mysql2 prepared statement support in Rails, I found them very hard to use. Most notably we sometimes need to close them eagerly, but it's complicated by the fact that you can only call `close` once, any subsequent call raises an error instead of just nooping.

There's also no way to check if a statement was closed or not.

cc @sodabrew @tenderlove

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -17,7 +17,8 @@ group :test do
   gem 'rspec', '~> 3.2'
 
   # https://github.com/bbatsov/rubocop/pull/4789
-  gem 'rubocop', '~> 1.30', '>= 1.30.1' if Gem::Version.new(RUBY_VERSION) >= Gem::Version.new('2.6')
+  # 1.51 is the last version supporting Ruby 2.6
+  gem 'rubocop', '>= 1.30.1', '< 1.51' if Gem::Version.new(RUBY_VERSION) >= Gem::Version.new('2.6')
 end
 
 group :benchmarks, optional: true do
--- a/ext/mysql2/statement.c
+++ b/ext/mysql2/statement.c
@@ -10,9 +10,12 @@ static VALUE intern_sec_fraction, intern_usec, intern_sec, intern_min, intern_ho
 #define TypedData_Get_Struct(obj, type, ignore, sval) Data_Get_Struct(obj, type, sval)
 #endif
 
-#define GET_STATEMENT(self) \
+#define RAW_GET_STATEMENT(self) \
   mysql_stmt_wrapper *stmt_wrapper; \
   TypedData_Get_Struct(self, mysql_stmt_wrapper, &rb_mysql_statement_type, stmt_wrapper); \
+
+#define GET_STATEMENT(self) \
+  RAW_GET_STATEMENT(self) \
   if (!stmt_wrapper->stmt) { rb_raise(cMysql2Error, "Invalid statement handle"); } \
   if (stmt_wrapper->closed) { rb_raise(cMysql2Error, "Statement handle already closed"); }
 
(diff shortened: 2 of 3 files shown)


ext/mysql2/extconf.rb
```diff
@@ -113,7 +113,7 @@ def add_ssl_defines(header)
   warn "-----\nUsing --with-mysql-dir=#{File.dirname inc}\n-----"
   rpath_dir = lib
   have_library('mysqlclient')
-elsif (mc = (with_config('mysql-config') || Dir[GLOB].first))
+elsif (mc = with_config('mysql-config') || Dir[GLOB].first)
```
byroot, line 116: I ran `rubocop -a` and it did this, figured I might as well leave it.

spec/mysql2/statement_spec.rb
```diff
@@ -320,7 +320,7 @@ def stmt_count
       result = @client.prepare("SELECT 1 UNION SELECT 2").execute(stream: true, cache_rows: false)
       expect do
         result.each {}
-        result.each {} # rubocop:disable Style/CombinableLoops
```
sodabrew, line 323: This is still needed

ext/mysql2/statement.c
```diff
@@ -603,12 +606,27 @@ static VALUE rb_mysql_stmt_affected_rows(VALUE self) {
...
-  GET_STATEMENT(self);
-  stmt_wrapper->closed = 1;
-  rb_thread_call_without_gvl(nogvl_stmt_close, stmt_wrapper, RUBY_UBF_IO, 0);
+  RAW_GET_STATEMENT(self);
+
+  if (!stmt_wrapper->closed) {
+      stmt_wrapper->closed = 1;
+      rb_thread_call_without_gvl(nogvl_stmt_close, stmt_wrapper, RUBY_UBF_IO, 0);
+  }
+
```
sodabrew, line 615: Makes sense!

spec/mysql2/statement_spec.rb
byroot: Indeed. Not sure why rubocop -a removed it. Maybe different version on my local.

ext/mysql2/extconf.rb
sodabrew: Very old line of code!

spec/mysql2/statement_spec.rb
byroot: Ah, I see:

```
        ruby-version: 2.6
        bundler-cache: true # runs 'bundle install' and caches installed gems automatically
```

It's using an old rubocop because rubocop dropped ruby 2.6 support a while ago.

sodabrew: Hah, so we're cross-rubocop-versions too now 😉 even recently reducing the compat window, it's still quite wide

Gemfile
```diff
@@ -17,7 +17,8 @@ group :test do
   gem 'rspec', '~> 3.2'
 
   # https://github.com/bbatsov/rubocop/pull/4789
-  gem 'rubocop', '~> 1.30', '>= 1.30.1' if Gem::Version.new(RUBY_VERSION) >= Gem::Version.new('2.6')
+  # 1.51 is the last version supporting Ruby 2.6
+  gem 'rubocop', '>= 1.30.1', '< 1.51' if Gem::Version.new(RUBY_VERSION) >= Gem::Version.new('2.6')
```
byroot, line 21: This should ensure a consistent rubocop result in local and CI.

spec/mysql2/statement_spec.rb
```diff
@@ -1,6 +1,6 @@
 require './spec/spec_helper'
 
-RSpec.describe Mysql2::Statement do
+RSpec.describe Mysql2::Statement do # rubocop:disable Metrics/BlockLength
```
sodabrew, line 3: I would guess that if not for the tests, this length could be reduced dramatically:
https://github.com/brianmario/mysql2/blob/03ac203ff077591171e54f65190071a3ef84525a/.rubocop_todo.yml#L37-L38

byroot: IMO that cop is just plain bad.

byroot: All `Metrics` are.

PR #1394, opened 2025-02-25
Add precision to time, datetime, and timestamp field types

ankane:
This updates `Result#field_types` to return the precision for `time`, `datetime`, and `timestamp` fields (like other types).

https://dev.mysql.com/doc/refman/8.4/en/fractional-seconds.html


sodabrew:
I like this, but would need to put it into a major-ish release

ankane:
Makes sense. I think it should probably always include the precision in that case.

sodabrew:
> Makes sense. I think it should probably always include the precision in that case.

Off the cuff, yes, ish? I think so.

PR #1409, opened 2025-09-23
doc: update macOS OpenSSL instructions to use OpenSSL 3.x

tnir:
OpenSSL 1.1 (1.1.1) had reached the EOL two years ago (2023-09). The relevant instructions should be removed and they should be with OpenSSL 3 (3.5 LTS).

#### Ref.

- https://openssl-corporation.org/post/2023-09-11-eol-111/

#### cf.

- https://formulae.brew.sh/formula/openssl@3
- https://ports.macports.org/port/openssl3/


sodabrew (approved):
Thank you!

PR #1412, opened 2025-10-01
Update `affected_rows` after calling `mysql_next_result`

### LINKED ISSUE #1411: affected_rows incorrect with multi-statements
When running multi-statements, the behavior of affected_rows appears to have changed in v0.5.7. The function now always returns the first statement’s affected_rows, instead of reflecting the result of the current statement.

This seems to be caused by [#1383](https://github.com/brianmario/mysql2/pull/1383/files)

```ruby
client.query('INSERT INTO roles (store_id, type) VALUES (1, 1), (2, 3); UPDATE roles SET type = 2 WHERE id = 1', as: :array)
client.affected_rows # => 2
client.next_result # => true
client.affected_rows # before was correctly 1, now returns 2
```

byroot:
Fix: https://github.com/brianmario/mysql2/issues/1411
Followup: https://github.com/brianmario/mysql2/pull/1383

When using multi statement, we need to update the affected rows after each call to `mysql_next_result`.

### DIFF
--- a/ext/mysql2/client.c
+++ b/ext/mysql2/client.c
@@ -1282,6 +1282,7 @@ static VALUE rb_mysql_client_next_result(VALUE self)
     int ret;
     GET_CLIENT(self);
     ret = mysql_next_result(wrapper->client);
+    wrapper->affected_rows = mysql_affected_rows(wrapper->client);
     if (ret > 0) {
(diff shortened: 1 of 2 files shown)


sodabrew:
lgtm!

PR #1417, opened 2025-10-21
Fix Client#affected_rows for queries with no affected rows

byroot:
Fix: https://github.com/brianmario/mysql2/issues/1414

e.g. `SELECT sleep(0.1)` or `SELECT 1`.

PR #1410, opened 2025-09-26
Fix prepared statements + 0 rows + fields segfault

skipkayhil:
Ref https://github.com/rails/rails/issues/55745

```
require "mysql2"

c = ::Mysql2::Client.new(database: "activerecord_unittest", username: "rails")
ps = c.prepare("SELECT * FROM products WHERE 1=0")
r = ps.execute
puts r.fields
```

This script (connection configured for the Rails repo, but unimportant) segfaults consistently on 0.5.7 but runs fine on 0.5.6. 

Reverting #1399 with no other changes fixes the segfault. I'm not sure if this is fully correct, but it doesn't crash.

PR #1418, opened 2025-10-22
Add test from PR 1410

sodabrew:
Add the test from PR #1410, which tests a bug that should also be fixed by #1417

### DIFF
--- a/ci/mysql57.sh
+++ b/ci/mysql57.sh
@@ -11,5 +11,5 @@ apt-key add support/B7B3B788A8D3785C.asc # 8.1 and higher
 # Verify the repository as add-apt-repository does not.
 wget -q --spider http://repo.mysql.com/apt/ubuntu/dists/$(lsb_release -cs)/mysql-5.7
 add-apt-repository 'http://repo.mysql.com/apt/ubuntu mysql-5.7'
-apt-get update -qq
-apt-get install -qq mysql-server libmysqlclient-dev
+apt-get update -o Acquire::Check-Valid-Until=false -qq
+apt-get install -o Acquire::Check-Valid-Until=false -qq mysql-server libmysqlclient-dev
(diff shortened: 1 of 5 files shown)


skipkayhil:
I tried testing with #1417 and I can still reliably segfault with this script:

```
# frozen_string_literal: true

require "mysql2"

c = ::Mysql2::Client.new(database: "activerecord_unittest", username: "rails")
ps = c.prepare("SELECT * FROM products WHERE 1=0")
r = ps.execute
puts r.fields
```

sodabrew:
Closing this — its goal has already been achieved by unrelated upstream work, and merging now would just add a duplicate test.

## What this PR was for

Add a regression test for #1410: a prepared statement's `Result#fields` segfaults when the query returns 0 rows, because mysql2 frees the underlying C result set as soon as iteration exhausts it, and `#fields` was reading from freed memory afterward. The PR description guessed #1417 might have fixed it as a side effect.

## What actually happened

#1417 didn't fix it — that was an unrelated `affected_rows` fix. The real fix landed later via #1427 and #1450, which cache `fields`/`field_types` right before freeing the result set, in every free path.

I verified this branch's test against the commit just before that fix (`b57e04e`) — it fails there (`expected 1, got 0`), confirming it was a real, meaningful regression test for a genuine bug, not a no-op.

## Why closing instead of merging

While preparing to strengthen the test, I found #1450 *also* independently added equivalent coverage directly to `spec/mysql2/statement_spec.rb` (already on `master`):

```ruby
it "should keep fields and field_types accessible for exhausted empty results" do
  statement = @client.prepare 'SELECT 1 AS only_col WHERE 1 = 0'
  result = statement.execute
  expect(result.to_a).to eql([])
  expect(result.fields).to eql(["only_col"])
  expect(result.field_types.length).to eql(1)
end
```

This covers the exact prepared-statement scenario this PR set out to test, and is more thorough than what's proposed here (it also checks `field_types`). Merging this PR now would add a duplicate test for an already-fixed, already-covered bug.

PR #1422, opened 2025-12-10
[Tracking] Tuning shop: result-fetching perf parcels (#1482-#1490)

jeremy:
**This is a tracking/integration PR — please review and merge the individual parcels below, not this branch.** It exists so the nine-way composition gets real CI, and so there's one place to watch the campaign converge: as parcels merge, this diff shrinks toward empty.

This branch is rebuilt from current master as the exact sum of the still-open standalone PRs, one commit per parcel, in dependency order. The original omnibus history is superseded — every fix from the per-parcel review cycles (UAF fix, UBSan gating, audit fixes) is included here via the parcels themselves.

## The parcels

| PR | What it does | Headline | Stack |
|---|---|---|---|
| #1482 | Share one frozen default for `Client#query`'s options argument | skips per-call Hash merge | ✅ **merged** (7bf6e9d) |
| #1483 | Build `Result#server_flags` lazily from status captured at creation | wrapper-slot memo; fixes `server_flags` nil on `store_result` (disclosed) | ✅ **merged** (e925cca) |
| #1484 | Fast-path integer parsing for integer result columns | Trilogy-style, hardened vs UB, UBSan-gated | ✅ **merged** (b46818d) |
| #1485 | Cache encoding lookups instead of redoing them per row | −13.7% / −10.6% string-heavy (Linux x86_64 / macOS arm64) | ✅ **merged** (7bb1d4d) |
| #1486 | `:force_encoding` query option that retags string results | skips charset→encoding handling when encodings are known | ✅ **merged** (2c85e4b) |
| #1487 | Array-mode results: materialize field names once, not per cell | −2.3% array-mode; closes an array-mode use-after-free on master | ✅ **merged** (fc8a284) |
| #1488 | Cache the parsed query options on the result wrapper | −19.8% argument-less re-`#each` | ✅ **merged** (03c0959) |
| #1489 | `cast: false` rows: skip the per-cell type dispatch entirely | small standalone win; substrate for #1490 | ✅ **merged** (466a9ae) |
| #1490 | `cast: :fast` rows: cast cheap types, defer expensive ones as strings | −64% / −41% vs full casting | ✅ **merged** (ceb90ed) |

Merge in any order; the stacks above are the only ordering constraints. Whatever lands first, I'll rebase the remaining parcels and this branch promptly.

## Integration verification

- Linear rebuild is tree-identical to the locally proven nine-way merge (the cross-parcel touch points — `mysql2_set_field_string_encoding` signature threading into #1489/#1490's call sites, #1488's parse-once absorbing #1486's per-`each` rejection and #1490's cast-mode enum — are folded into the parcel commits here).
- Full suite: 500 examples, 0 failures, 10 pending; rubocop clean (36 files, no offenses).
- CI on this PR exercises the composed stack across the full matrix.

## Aggregate benchmark: master vs full stack

Same protocol as the per-parcel numbers: quiet Linux x86_64 box (load < 0.5), 3 serial alternating base/stack pairs, each invocation reporting per-arm median of 9 samples (each sample the min of 9 timed runs, arms interleaved); numbers below are the median across the 3 pairs. 1000-row mixed table (`INT`×3, `BIGINT`, `DOUBLE`, `DECIMAL(12,4)`, `DATETIME`, `VARCHAR`×2), 20 queries per run. Client lib is libmariadb 12.3.2.

| Workload | master 96252dc | full stack | Δ time | Δ allocs/run |
|---|---|---|---|---|
| defaults (hash, `cast: true`) | 29.22 ms | 27.92 ms | **−4.4%** | −40 |
| `as: :array` | 27.58 ms | 25.97 ms | **−5.8%** | −20 |
| `cast: false` | 18.31 ms | 16.18 ms | **−11.6%** | −20 |
| `cast: :fast` (vs full cast) | 29.29 ms | 16.12 ms | **−45.0%** | −40,021 (−25%) |
| `:force_encoding` (string-only) | 7.00 ms | 6.60 ms | **−5.7%** | −20 |
| int-only columns | 6.68 ms | 6.20 ms | **−7.2%** | −40 |
| string-only columns | 7.04 ms | 6.55 ms | **−7.0%** | −40 |
| argument-less re-`#each` ×5 | 31.18 ms | 29.87 ms | **−4.2%** | −40 |

Composition: no workload regresses, and the per-arm deltas are directionally consistent with the standalone parcel numbers on this shape (`cast: :fast` lands at −45% here because only 2 of 8 columns are DECIMAL/DATETIME — the −64% headline shape is denser in expensive types; string arms land below #1485's −13.7% string-heavy shape for the same reason, shorter strings and fewer of them). The `cast: :fast` allocation drop is the deferred BigDecimal/Time construction.

### DIFF
--- a/.github/workflows/build-ubuntu.yml
+++ b/.github/workflows/build-ubuntu.yml
@@ -4,14 +4,21 @@ on:
 jobs:
   build:
     name: >-
-      ${{ matrix.os }} ruby ${{ matrix.ruby }} ${{ matrix.db }}
+      ${{ matrix.os }} ruby ${{ matrix.ruby }} ${{ matrix.db }}${{ matrix.allow_failure && ' (allow failure)' || '' }}
     runs-on: ${{ matrix.os }}
     timeout-minutes: 30
+    # allow_failure entries track something that moves outside this repository,
+    # so their failures are informational and must not block unrelated PRs.
+    continue-on-error: ${{ matrix.allow_failure || false }}
     strategy:
       matrix:
         include:
+          # Ruby 4.x on Ubuntu 26.04 LTS
+          # db: mysql-innovation tracks the newest MySQL innovation release, not an LTS.
+          - {os: ubuntu-26.04, ruby: '4.0', db: mysql-innovation, allow_failure: true}
+          - {os: ubuntu-26.04, ruby: 'head', db: mysql97, allow_failure: true}
+
           # Ruby 3.x / 4.x on Ubuntu 24.04 LTS
-          - {os: ubuntu-24.04, ruby: 'head', db: mysql97}
           - {os: ubuntu-24.04, ruby: '4.0', db: mysql97}
           - {os: ubuntu-24.04, ruby: '4.0', db: mysql84}
           - {os: ubuntu-24.04, ruby: '3.4', db: mysql84}
--- a/ci/mysql-innovation.sh
+++ b/ci/mysql-innovation.sh
@@ -1,0 +1,29 @@
+#!/usr/bin/env bash
+
+set -eux
+
+apt-get purge -qq '^mysql*' '^libmysql*'
+rm -fr /etc/mysql
+rm -fr /var/lib/mysql
+
+RELEASE=$(lsb_release -cs)
+# Innovation releases get no version-specific component of their own.
+COMPONENT=mysql-innovation
+
+# Verify the component exists, as apt-get update only warns when it is missing.
+wget -q --spider "https://repo.mysql.com/apt/ubuntu/dists/$RELEASE/$COMPONENT"
+
+install -d -m 0755 /etc/apt/keyrings
+install -m 0644 support/B7B3B788A8D3785C.asc /etc/apt/keyrings/mysql-keyring.asc
+
+tee /etc/apt/sources.list.d/mysql.sources <<- EOF
+	X-Repolib-Name: MySQL
+	Types: deb
+	URIs: https://repo.mysql.com/apt/ubuntu
+	Suites: $RELEASE
+	Components: $COMPONENT
+	Signed-By: /etc/apt/keyrings/mysql-keyring.asc
+EOF
+
+apt-get update -qq
+apt-get install -qq mysql-server libmysqlclient-dev
--- a/ci/setup.sh
+++ b/ci/setup.sh
@@ -27,6 +27,12 @@ if [[ -n ${DB-} && x$DB =~ ^xmysql97 ]]; then
   CHANGED_PASSWORD_SHA2=true
 fi
 
+# Install the newest MySQL innovation release if DB=mysql-innovation
+if [[ -n ${DB-} && x$DB =~ ^xmysql-innovation ]]; then
+  sudo bash ci/mysql-innovation.sh
+  CHANGED_PASSWORD_SHA2=true
+fi
+
 # Install MariaDB 10.6 if DB=mariadb10.6
 if [[ -n ${GITHUB_ACTIONS-} && -n ${DB-} && x$DB =~ ^xmariadb10.6 ]]; then
   sudo bash ci/mariadb106.sh
--- a/ext/mysql2/client.c
+++ b/ext/mysql2/client.c
@@ -698,8 +698,6 @@ static VALUE rb_mysql_client_async_result(VALUE self) {
   Check_Type(current, T_HASH);
   resultObj = rb_mysql_result_to_obj(self, wrapper->encoding, current, result, Qnil);
 
-  rb_mysql_set_server_query_flags(wrapper->client, resultObj);
-
   return resultObj;
 }
 
--- a/ext/mysql2/result.c
+++ b/ext/mysql2/result.c
@@ -5,6 +5,12 @@
 
 static rb_encoding *binaryEncoding;
 
+/* Trilogy-style encoding index cache: charset number → encoding index
+ * 0 = not yet cached, -1 = cached "not found", > 0 = valid encoding index
+ * Amortizes the expensive rb_enc_find_index() lookup across all queries
+ */
+static int encoding_index_cache[256] = {0};
+
 /* on 64bit platforms we can handle dates way outside 2038-01-19T03:14:07
  *
  * (9999*31557600) + (12*2592000) + (31*86400) + (11*3600) + (59*60) + 59
@@ -413,18 +419,29 @@ static VALUE mysql2_set_field_string_encoding(VALUE val, MYSQL_FIELD field, rb_e
     /* MySQL 4.x may not provide an encoding, binary will get the bytes through */
     rb_enc_associate(val, binaryEncoding);
   } else {
-    /* lookup the encoding configured on this field */
-    const char *enc_name;
-    int enc_index;
+    /* Trilogy-style cached encoding lookup: check cache first */
+    int enc_index = encoding_index_cache[field.charsetnr];
 
-    enc_name = (field.charsetnr-1 < MYSQL2_CHARSETNR_SIZE) ? mysql2_mysql_enc_to_rb[field.charsetnr-1] : NULL;
+    if (enc_index == 0) {
+      /* Cache miss: do full lookup and cache result */
+      const char *enc_name;
+      enc_name = (field.charsetnr-1 < MYSQL2_CHARSETNR_SIZE) ? mysql2_mysql_enc_to_rb[field.charsetnr-1] : NULL;
 
-    if (enc_name != NULL) {
-      /* use the field encoding we were able to match */
-      enc_index = rb_enc_find_index(enc_name);
+      if (enc_name != NULL) {
+        enc_index = rb_enc_find_index(enc_name);
+      } else {
+        enc_index = -1;  /* Mark as "not found" */
+      }
+
+      /* Cache the result (even -1 for "not found") */
+      encoding_index_cache[field.charsetnr] = enc_index;
+    }
+
+    /* Use the encoding index (from cache or fresh lookup) */
+    if (enc_index > 0) {
       rb_enc_set_index(val, enc_index);
     } else {
-      /* otherwise fall-back to the connection's encoding */
+      /* Encoding not found, fall back to connection encoding */
       rb_enc_associate(val, conn_enc);
     }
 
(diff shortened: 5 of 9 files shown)


jeremy:
This is an omnibus branch will all perf tunings. I'll split these out into distinct PRs for individual consideration.

sodabrew:
Looking good!!

nuclearspike:
While benchmarking a similar set of optimizations against master I hit one behavioral edge that applies to the "skip field name allocation in as: :array mode" change here, in case it saves a review cycle: the per-cell field fetch that array mode currently wastes is also the only thing that materializes `#fields` for *streaming* results. If array-mode iteration skips it entirely, then after a streaming `each(as: :array)` completes (result freed), `result.fields` goes from returning the names to raising `Result set has already been freed`:

```ruby
result = client.query("SELECT 1 AS a, 2 AS b", stream: true, cache_rows: false, as: :array)
result.each { |row| }
result.fields  # master: ["a", "b"] — with names skipped in array mode: raises
```

A cheap one-time materialization at `each` entry (rather than per cell) keeps both the win and the `#fields` contract; that's the shape I ended up with after the regression bit me.

FWIW, on Apple Silicon / Ruby 3.3.11 / MySQL 9.6 I also measured a few optimizations that look complementary to this set, with interleaved-A/B methodology on 50k-row tables: a `:utc` datetime fast path (days-from-civil + `rb_time_timespec_new`, guarded by extconf with the funcall fallback) at −20.6% text / −28.8% binary on DATETIME(6)-heavy results; `rb_hash_new_capa` row pre-sizing (Ruby 3.2+, guarded); binding stmt result buffers once per result instead of per row; and skipping the rows-array capacity reservation under `cache_rows: false` (8 bytes/row of never-used capacity, 16.4MB on a 2M-row iteration). Happy to share branches or rebase any of them on top of this PR if useful.

sodabrew:
@jeremy summary of where this landed, so you know what's ready to act on.

Reviewed this against current `master`, assuming #1457 and #1461 land. Headline finding: none of the items below actually overlap with #1452/#1457/#1458/#1459/#1460/#1461 — those are a separate, independently-implemented optimization effort (different author, different mechanisms: statement bind-buffer caching, DATETIME parsing, GVL release timing). I checked `master`'s `result.c`/`result.h`/`client.c` directly for every marker this PR introduces (`encoding_index_cache`, `forced_encoding`, `CAST_NONE`/`CAST_FAST`, `ull_from_buf`/`ll_from_buf`, the cached-options wrapper fields) and found zero matches. So this PR's wins are still fully unclaimed — nothing here is redundant with what's already landing.

| Optimization | Status |
|---|---|
| Skip field-name allocation in `as: :array` mode | ready to cherry-pick as standalone PR |
| Skip Hash allocation in `Client#query` when no options passed | ready to cherry-pick as standalone PR |
| Cache query options on the wrapper (avoid per-row Hash lookups in `#each`) | ready to cherry-pick as standalone PR |
| Integer fast-path parsing (`ull_from_buf`/`ll_from_buf`, Trilogy-style) | ready to cherry-pick as standalone PR |
| Encoding cache (charset→index cache + cached `conn_enc`/`default_internal_enc`) | ready to cherry-pick as standalone PR |
| `force_encoding: "utf-8"` option | ready to cherry-pick as standalone PR (bundle with the encoding cache — shares the same `mysql2_set_field_string_encoding` signature change) |
| Lazy-load server flags (`Result#server_flags`) | to be investigated further |
| `cast: false` fast path | to be investigated further |
| Partial casting (`cast: :fast`) | to be investigated further |

## Notes on the "ready to cherry-pick" group

All six only need a mechanical rebase, not a redesign — `result.h`'s `mysql2_result_wrapper` struct has moved on (my #1463 added `result_buffers_bound`/`mysql2_result_force_free`; presumably #1459 too), so the new fields this PR adds (`symbolize_keys`, `as_array`, `cast_bool`, `cache_rows`, `cast`, `db_timezone`, `app_timezone`, `default_internal_enc`, `conn_enc`, `forced_encoding`) need re-adding against the current struct shape, not the one this PR's base predates. No conceptual conflict with anything merged since, just line drift.

## Notes on the "to be investigated further" group

- **Lazy server flags**: the diff only removes the eager `rb_mysql_set_server_query_flags()` call from `client.c`'s async-result path. `statement.c` has its own call to the same function (line 614 on current `master`) that this PR doesn't touch — so as submitted, prepared-statement results would still eagerly compute what non-prepared results now compute lazily. Also changes `Result#server_flags` from `attr_reader` to a computed method with a different key set (`no_good_index_used`/`no_index_used`/`query_was_slow` vs. whatever it returns today) — worth confirming that's not an observable behavior change for anyone relying on the old shape.
- **`cast: false` fast path**: only implemented in `rb_mysql_result_fetch_row` (the plain-query path). `rb_mysql_result_fetch_row_stmt` (prepared statements) doesn't get the same fast path in this diff — same parity question as above. Worth deciding whether that's an intentional scope limit or a gap to close before merging.
- **`cast: :fast` partial casting**: the largest surface here — a new public `cast:` value, new selective-casting semantics, and the same prepared-statement parity gap (only wired into `rb_mysql_result_fetch_row`, not `_stmt`). This is also the one most worth a second pair of eyes on the boolean-cast fallthrough (`MYSQL_TYPE_TINY` falls through into the shared integer path when `castBool` is false or `length != 1`) and on what "leave DECIMAL/DATETIME/TIME as strings" means for callers who aren't ActiveRecord and don't expect that contract to depend on a `cast:` value. I'd want this one benchmarked and reviewed on its own, separate from the other two buckets.

Happy to actually do the cherry-pick rebasing for the first group if that's useful — they're small, mechanical, and don't need design input.

Ready for next steps whenever you are.

jeremy:
Ahh thanks for fleshing those PRs out @nuclearspike - love it.

I'll see what remains to be mined here.

jeremy:
Extraction status: this branch is being parceled into standalone PRs against current master (base 82f9e18), redesigned where master has moved underneath the original commits. Wave 1 is up:

| Original commit(s) | Standalone PR |
|---|---|
| ed97ddb (skip options merge) | #1482 |
| 0f876e6 (lazy `server_flags`) | #1483 |
| 96bb584 (integer fast path) | #1484 |
| 73b7e26 + 9b85a9c (encoding caches) | #1485 |

Each PR body carries its own benchmarks and disclosed deltas. The four branches merge cleanly together locally (full suite green on the combined stack), but they're independent — any merge order works, and I'll rebase the remainder as each lands.

Still to come from this branch: `force_encoding:` (1940ce1, stacks on #1485), array-mode field-name skip (702c635 + 33a028e, re-measuring first), `cast: false` fast path (7000a99), and cached query options (946a52b) + `cast: :fast` (1100052) — the last two get benchmarked against current master first and will be parked here with numbers if they no longer pay their way.

Keeping this PR open as the tracking umbrella until every parcel is merged or parked.

jeremy:
Extraction complete: all nine parcels from this branch are now standalone PRs against current master (96252dc), each with its own benchmarks (quiet Linux x86_64 box + macOS arm64), specs verified to fail without the change, and full CI. Updated map:

| Original commit(s) | Standalone PR | Notes |
|---|---|---|
| ed97ddb (skip options merge) | #1482 | redesigned: shared frozen default arg, exact master semantics |
| 0f876e6 (lazy `server_flags`) | #1483 | wrapper-slot memo; fixes `server_flags` nil on `store_result` (disclosed behavior change) |
| 96bb584 (integer fast path) | #1484 | hardened vs the original's UB; UBSan-gated with linked runtime |
| 73b7e26 + 9b85a9c (encoding caches) | #1485 | −13.7% / −10.6% string-heavy on the two platforms |
| 1940ce1 (`force_encoding:`) | #1486 | stacked on #1485 (shared `mysql2_set_field_string_encoding` change) |
| 702c635 + 33a028e (field-name skip) | #1487 | −2.3% array-mode, plus closes an array-mode use-after-free on master |
| 7000a99 (`cast: false` fast path) | #1489 | stacked on #1485+#1487; small standalone win, substrate for #1490 |
| 946a52b (cached query options) | #1488 | benchmark-gated: −19.8% argument-less re-`#each` earned the ship |
| 1100052 (`cast: :fast`) | #1490 | stacked on #1489; −64%/−41% vs full casting, ties the `cast: false` floor |

Nothing parked: every parcel cleared its benchmark gate (the two gated ones, #1487 and #1488, carried their numbers in the body; #1490's ceiling prescreen and results are in its body too).

For merge ordering: the nine branches are independent against master except the disclosed stacks (#1485→#1486, #1485+#1487→#1489→#1490). I keep a local integration branch with all nine merged — full suite green (500 examples, 0 failures) — so the cross-parcel touch points are known and mechanical: #1486's `mysql2_set_field_string_encoding` signature threads into #1489/#1490's new call sites, and #1488's parse-once structure absorbs #1486's per-`each` rejection and #1490's cast-mode enum. Whatever merges first, I'll rebase the rest promptly and re-run suites.

Keeping this PR open as the tracking umbrella until the parcels land; happy to close it sooner if you'd rather track via the individual PRs.

sodabrew:
Nice! Merged about half, the other need to be rebased due to (minor) merge conflicts.

jeremy:
Thanks for the quick merges! The remaining four are now rebased onto current master (fc8a284) and all show clean-applying again — full suite green on each:

- #1485 (encoding caches) — the base of the rest; the conflict was field adjacency with #1483's struct change
- #1486 (`:force_encoding`) — on #1485
- #1489 (`cast: false` fast path) — on #1485 (its field-name-skip layer now comes via master)
- #1490 (`cast: :fast`) — on #1489

So the remaining order constraint is just: #1485 first, then #1486/#1489 in either order, #1490 last. I'll keep rebasing promptly as these land.

jeremy:
All nine parcels are merged — the tracking branch has collapsed to master (its tree is byte-identical to `2c85e4b`), so this PR's diff is now empty and the campaign is complete. Closing.

Huge thanks @sodabrew for the fast, careful review train. 🚂

| PR | Merged as |
|---|---|
| #1482 | `7bf6e9d` |
| #1483 | `e925cca` |
| #1484 | `b46818d` |
| #1485 | `7bb1d4d` |
| #1486 | `2c85e4b` |
| #1487 | `fc8a284` |
| #1488 | `03c0959` |
| #1489 | `466a9ae` |
| #1490 | `ceb90ed` |

Aggregate benchmark numbers for the full stack are in the body above.

PR #1423, opened 2026-02-20
MariaDB determine json from Extended Metadata

grooverdan:
In MariaDB the JSON of a result value is from the extended metadata of the protocol.

This is returned when PROTOCOL_41 is in the connection flags.

ref: https://mariadb.com/docs/server/reference/clientserver-protocol/4-server-response-packets/result-set-packets#extended-metadata

`rb_mariadb_json_type` compiles to 0 the MySQL connector as its using `mariadb_field_attr` from MariaDB Connector/C.

JSON is supported on all MariaDB supported version so I've just dropped the version check.

### DIFF
--- a/ext/mysql2/result.c
+++ b/ext/mysql2/result.c
@@ -240,6 +240,18 @@ static VALUE rb_mysql_result_fetch_field(VALUE self, unsigned int idx, int symbo
   return rb_field;
 }
 
+static int rb_mariadb_json_type(const MYSQL_FIELD *field) {
+#if defined(MARIADB_PACKAGE_VERSION)
+    MARIADB_CONST_STRING field_attr;
+
+    if (!mariadb_field_attr(&field_attr, field,
+                            MARIADB_FIELD_ATTR_FORMAT_NAME)) {
+      return field_attr.length == 4 && !memcmp(field_attr.str, "json", 4);
+    }
+#endif
+    return 0;
+}
+
 static VALUE rb_mysql_result_fetch_field_type(VALUE self, unsigned int idx) {
   VALUE rb_field_type;
   GET_RESULT(self);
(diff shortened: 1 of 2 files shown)


ext/mysql2/result.c
```diff
@@ -240,6 +240,18 @@ static VALUE rb_mysql_result_fetch_field(VALUE self, unsigned int idx, int symbo
   return rb_field;
 }
 
+static int rb_mariadb_json_type(const MYSQL_FIELD *field) {
+#if defined(MARIADB_PACKAGE_VERSION)
+    MARIADB_CONST_STRING field_attr;
+
+    if (!mariadb_field_attr(&field_attr, field,
+                            MARIADB_FIELD_ATTR_FORMAT_NAME)) {
+      return field_attr.length == 4 && strncmp(field_attr.str, "json", 4) == 0;
```
dr-m: I think that we can rely on the short-circuit evaluation of the length, and just compare 4 bytes using `memcmp`, which the compiler can replace with equivalent instructions. https://godbolt.org/z/K8Wz7sn7P demonstrates that the `memcmp` variant would be much shorter and avoid a call to a library function.

ext/mysql2/result.c
```diff
@@ -309,7 +321,9 @@ static VALUE rb_mysql_result_fetch_field_type(VALUE self, unsigned int idx) {
         rb_field_type = rb_sprintf("decimal(%d,%d)", precision, field->decimals);
         break;
       case MYSQL_TYPE_STRING:       // char[]
-        if (field->flags & ENUM_FLAG) {
+	if (rb_mariadb_json_type(field)) {
```
sodabrew: Looks like a whitespace issue?

sodabrew:
I merged your updates for the package signing keys, could you rebase the PR so we see CI run through?

grooverdan:
Thanks @sodabrew

PR #1427, opened 2026-03-25
Fix segfault when accessing fields/field_types on freed result

### LINKED ISSUE #1426: Segfault in rb_mysql_result_fetch_fields when accessing fields after result freed (0-row results)
## Summary

A segmentation fault occurs when calling `.fields` on a mysql2 Result object after the internal result has been freed. This happens specifically with 0-row query results where the internal caching mechanism frees the result before `.fields` is accessed.

## Environment

- **Ruby version**: 3.4.3 (also reproducible on 3.2.x, 3.3.x)
- **mysql2 version**: 0.5.7
- **MySQL server**: 5.7.x / 8.0.x
- **OS**: Linux (AWS EC2), macOS
- **Rails**: 8.0.4 (but not Rails-specific)

## Crash Details

```
/path/to/gems/activerecord-8.0.4/lib/active_record/connection_adapters/mysql2/database_statements.rb:114:in 'fields'
/path/to/gems/activerecord-8.0.4/lib/active_record/connection_adapters/mysql2/database_statements.rb:114:in 'cast_result'
```

Crash address: `0x000000000001a007` (consistent across multiple crashes)

## Root Cause Analysis

The bug is in `ext/mysql2/result.c` in the `rb_mysql_result_fetch_fields` function (around line 941-944):

```c
if (wrapper->fields == Qnil) {
  wrapper->numberOfFields = mysql_num_fields(wrapper->result);  // CRASH: wrapper->result is freed
  wrapper->fields = rb_ary_new2(wrapper->numberOfFields);
}
```

**The problem**: There is no check for `wrapper->resultFreed` before accessing `wrapper->result`. 

This check exists in other functions (e.g., line 1132-1134 in `rb_mysql_result_each_`):
```c
if (wrapper->resultFreed) {
  rb_raise(cMysql2Error, "Result set has already been freed");
}
```

But it's missing in `rb_mysql_result_fetch_fields` and `rb_mysql_result_fetch_field_types`.

## Reproduction Scenario

The crash occurs when:

1. A query returns 0 rows
2. mysql2's internal row caching iterates over results (0 iterations for 0 rows)
3. The loop body never executes, so `wrapper->fields` is never populated (stays `Qnil`)
4. After iteration completes, mysql2 sees "0 rows processed == 0 rows total" and frees the result
5. Subsequently, `.fields` is called (e.g., by ActiveRecord's `cast_result`)
6. `rb_mysql_result_fetch_fields` checks `wrapper->fields == Qnil` (true)
7. It tries to call `mysql_num_fields(wrapper->result)` on the freed result
8. **SEGFAULT**

This is particularly common with queries like:
```sql
SELECT * FROM orders_extras WHERE booking_id IN (...) AND price > 0
-- Returns 0 rows when no extras exist
```

## Proposed Fix

Add a `resultFreed` check to `rb_mysql_result_fetch_fields` (and `rb_mysql_result_fetch_field_types`):

```c
if (wrapper->fields == Qnil) {
  if (wrapper->resultFreed) {
    rb_raise(cMysql2Error, "Result set has already been freed");
  }
  wrapper->numberOfFields = mysql_num_fields(wrapper->result);
  wrapper->fields = rb_ary_new2(wrapper->numberOfFields);
}
```

## Impact

We've observed 9+ production crashes in 30 days from this issue, all at the same memory address, all during queries that return 0 rows.

## Workaround

Currently using a workaround that bypasses ActiveRecord's `cast_result` by using raw mysql2 queries with `as: :hash` option, which doesn't require calling `.fields`.

## Related

This may be related to or a variant of previous segfault issues reported in this repository.

jonathan-biskycreative:
## Summary

Fixes segfault when accessing `.fields` or `.field_types` on a Result object after the internal result has been freed. This is particularly common with 0-row query results.

## Problem

A segfault occurs in the following sequence:
1. Query returns 0 rows
2. Internal row caching iterates (0 iterations), never populating `wrapper->fields`
3. Result is freed after iteration completes (0 processed == 0 total)
4. `.fields` is called, `wrapper->fields` is `Qnil`
5. `mysql_num_fields(wrapper->result)` accesses freed memory
6. **SEGFAULT**

## Fix

This PR includes two improvements:

### 1. Eager field population (primary fix)
Fields are now populated immediately when a result is created, before any iteration can free the result:

```c
// In rb_mysql_result_to_obj
if (r != NULL && !wrapper->is_streaming) {
  rb_mysql_result_fetch_fields(obj);
}
```

This prevents the problematic scenario from occurring at all.

### 2. Safety check (fallback)
The `resultFreed` check (which already exists in other functions like `rb_mysql_result_each_`) is now added to the fields accessor functions:

```c
if (wrapper->resultFreed) {
  rb_raise(cMysql2Error, "Result set has already been freed");
}
```

If an edge case still occurs, this raises a catchable Ruby exception instead of crashing the process.

## Testing

This fix has been tested in a production Rails 8.0 application where the segfault was occurring ~9 times per month during queries that return 0 rows. With the eager field population, the issue no longer occurs.

Fixes #1426

### DIFF
--- a/ext/mysql2/result.c
+++ b/ext/mysql2/result.c
@@ -939,6 +939,9 @@ static VALUE rb_mysql_result_fetch_fields(VALUE self) {
   }
 
   if (wrapper->fields == Qnil) {
+    if (wrapper->resultFreed) {
+      rb_raise(cMysql2Error, "Result set has already been freed");
+    }
     wrapper->numberOfFields = mysql_num_fields(wrapper->result);
     wrapper->fields = rb_ary_new2(wrapper->numberOfFields);
   }
@@ -958,6 +961,9 @@ static VALUE rb_mysql_result_fetch_field_types(VALUE self) {
   GET_RESULT(self);
 
   if (wrapper->fieldTypes == Qnil) {
+    if (wrapper->resultFreed) {
+      rb_raise(cMysql2Error, "Result set has already been freed");
+    }
     wrapper->numberOfFields = mysql_num_fields(wrapper->result);
     wrapper->fieldTypes = rb_ary_new2(wrapper->numberOfFields);
   }
@@ -1192,6 +1198,7 @@ VALUE rb_mysql_result_to_obj(VALUE client, VALUE encoding, VALUE options, MYSQL_
   wrapper->lastRowProcessed = 0;
   wrapper->resultFreed = 0;
   wrapper->result = r;
+  wrapper->numberOfFields = 0;
   wrapper->fields = Qnil;
   wrapper->fieldTypes = Qnil;
   wrapper->rows = Qnil;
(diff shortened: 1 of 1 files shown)


sodabrew:
@copilot Add the test from PR #1449 to this branch. Put it in a sensible location in the existing results spec file.

PR #1432, opened 2026-04-17
Fix test assertions for TLS-enabled mariadb-connector-c connections

FaramosCZ:
mariadb-connector-c 3.4.0+ enables MYSQL_OPT_SSL_VERIFY_SERVER_CERT by default (MDEV-31857), so all TCP connections use TLS. This changes the error messages surfaced when connections are disrupted:

- fork() corrupts shared TLS state, raising "TLS/SSL error" in both child and parent instead of silent success
- Connection kills and timeouts surface as TLS errors rather than "Lost connection to MySQL server"

Adapt test expectations to accept both TLS and non-TLS error messages, so the test suite passes regardless of whether the connector library enforces TLS.

### DIFF
--- a/spec/mysql2/client_spec.rb
+++ b/spec/mysql2/client_spec.rb
@@ -340,13 +340,24 @@ RSpec.describe Mysql2::Client do # rubocop:disable Metrics/BlockLength
           client.query('SELECT 1')
           client = nil
           run_gc
+        rescue Mysql2::Error
+          # With TLS, forking corrupts shared state and child queries
+          # may raise TLS errors. That is expected.
+          nil
         end
 
         Process.wait(child)
 
         # this will throw an error if the underlying socket was shutdown by the
-        # child's GC
-        expect { client.query('SELECT 1') }.to_not raise_exception
+        # child's GC.
+        # With TLS, the fork corrupts shared TLS state, so TLS errors
+        # are also acceptable here — the important thing is the socket
+        # was not closed by the child's GC.
+        begin
+          client.query('SELECT 1')
+        rescue Mysql2::Error => e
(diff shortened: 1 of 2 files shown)


sodabrew:
@copilot update from latest main, make sure we can run tests

sodabrew:
@copilot update from latest main

sodabrew:
Please rebase on latest master - significant CI improvements will help verify this PR!

sodabrew:
Closing as superseded — checked all four changes here directly against current master and each is already covered, via #1468 and its follow-ups earlier this cycle:

- Fork test (child/parent TLS tolerance): master pins `ssl_mode: 'disabled'` explicitly instead (`client_spec.rb:379`), sidestepping the TLS-fork-corruption issue rather than tolerating its symptoms.
- "should detect closed connection on query read error": master already broadens to the same `/Lost connection|TLS\/SSL error/` pattern (`client_spec.rb:703`).
- "should be impervious to connection-corrupting timeouts in #execute": master splits this into two tests — a strict `ssl_mode: 'disabled'` version and a separate TLS-tolerant version (`client_spec.rb:798` and `:810`) — rather than one test with inline rescue/tolerance.
- Streaming-timeout test: master already broadens to `/Lost connection|TLS\/SSL error/` (`result_spec.rb:409`), alongside a separate deterministic plaintext-pinned variant (`:379`).

Thanks for flagging the underlying mariadb-connector-c 3.4+ TLS-by-default issue — appreciate the diagnosis, even though this particular patch is now redundant with what landed. Closing rather than asking for a rebase that would just be a no-op (or reintroduce the inline-tolerance pattern we specifically moved away from in the two tests above).

Issue #1282, opened 2022-09-15
Segmentation faults when running GraphQL tests (Rails, Minitest)


jcat4:
## Overview
We have some GraphQL mutation tests in a rails application that have been inconsistently failing with `malloc`/`segfault` issues. The more tests we run, the more likely the issues are to arise.

The hangs and freezes have been occurring locally on an M1 Pro MacBook Pro. These issues were also reproducible in our CI pipeline (docker containers, unsure which distro) and in a remote development container (Ubuntu 21.10, Kernel version 5.10.68)

We ran the tests through Valgrind on the Ubuntu container mentioned above and were able to capture a log of what happened.

Here is a notable snippet from the log:
```
==716395== Invalid read of size 4
==716395==    at 0xCD504A6: BIO_clear_flags (in /usr/lib/x86_64-linux-gnu/libcrypto.so.1.1)
==716395==    by 0xCD6056B: ??? (in /usr/lib/x86_64-linux-gnu/libcrypto.so.1.1)
==716395==    by 0xCD4E409: ??? (in /usr/lib/x86_64-linux-gnu/libcrypto.so.1.1)
==716395==    by 0xCD534B3: ??? (in /usr/lib/x86_64-linux-gnu/libcrypto.so.1.1)
==716395==    by 0xCD53596: BIO_write (in /usr/lib/x86_64-linux-gnu/libcrypto.so.1.1)
==716395==    by 0xCC46E80: ??? (in /usr/lib/x86_64-linux-gnu/libssl.so.1.1)
==716395==    by 0xCC4D688: ??? (in /usr/lib/x86_64-linux-gnu/libssl.so.1.1)
==716395==    by 0xCC4D954: ??? (in /usr/lib/x86_64-linux-gnu/libssl.so.1.1)
==716395==    by 0xCC5C4A6: SSL_write (in /usr/lib/x86_64-linux-gnu/libssl.so.1.1)
==716395==    by 0x11E560E5: ??? (in /usr/lib/x86_64-linux-gnu/libperconaserverclient.so.20.3.26)
==716395==    by 0x11E332D9: ??? (in /usr/lib/x86_64-linux-gnu/libperconaserverclient.so.20.3.26)
==716395==    by 0x11E334A6: ??? (in /usr/lib/x86_64-linux-gnu/libperconaserverclient.so.20.3.26)
[...]

Issue #1019, opened 2019-01-11
mysql2 gem crashes randomly


herbertia:
I am in the process to upgrade our application to Rails 5. 
I have upgraded to ruby 2.4.5 with all the gems in our gemfile with Rails 5.2.2. The mysql2 gem is 0.5.2. When I am running our tests and I have randoms errors:
- Mysql2::Error
`# /usr/local/rvm/gems/ruby-2.4.5@clm/gems/activerecord-5.2.2/lib/active_record/connection_adapters/mysql/database_statements.rb:45:in 'affected_rows'`
` # /usr/local/rvm/gems/ruby-2.4.5@clm/gems/activerecord-5.2.2/lib/active_record/connection_adapters/mysql/database_statements.rb:45:in 'block in exec_delete'`
` # /usr/local/rvm/gems/ruby-2.4.5@clm/gems/activerecord-5.2.2/lib/active_record/connection_adapters/abstract_mysql_adapter.rb:196:in 'execute_and_free'`
` # /usr/local/rvm/gems/ruby-2.4.5@clm/gems/activerecord-5.2.2/lib/active_record/connection_adapters/mysql/database_statements.rb:45:in 'exec_delete'`
` # /usr/local/rvm/gems/ruby-2.4.5@clm/gems/activerecord-5.2.2/lib/active_record/connection_adapters/abstract/database_statements.rb:168:in 'delete'`
` # /usr/local/rvm/gems/ruby-2.4.5@clm/gems/activerecord-5.2.2/lib/active_record/connection_adapters/abstract/query_cache.rb:21:in 'delete'`
` # /usr/local/rvm/gems/ruby-2.4.5@clm/gems/activerecord-5.2.2/lib/active_record/relation.rb:409:in 'delete_all'`
` # /usr/local/rvm/gems/ruby-2.4.5@clm/gems/activerecord-5.2.2/lib/active_record/querying.rb:10:in 'delete_all'`

- or this kind of mysql errors:
`Failure/Error: _query(sql, @query_options.merge(options))`
` ActiveRecord::StatementInvalid:`
`   Mysql2::Error::ConnectionError: MySQL server has gone away: BEGIN`
` # /usr/local/rvm/gems/ruby-2.4.5@clm/gems/mysql2-0.5.2/lib/mysql2/client.rb:131:in '_query'`
[...]

Issue #777, opened 2016-07-14
mysql_ping causing Thread interrupts in Phusion passenger


technopoetic:
We've been having this issue in production for quite a while now, most of this year, actually.  It typically occurs several times a day, but it seems random.  So far I haven't been able to track down what's causing it, but my hypothesis is that when mysql_ping initiates a reconnect, it interrupts the current thread, causing Phusion passenger to raise a FATAL exception.  I've also had virtually no luck reproducing it locally.  Do you guys have any insight?

The fatal exception is:
PhusionPassenger::RequestHandler::ThreadHandler::Interrupted

Stacktrace:

    [GEM_ROOT]/gems/activerecord-4.2.5.2/lib/active_record/connection_adapters/mysql2_adapter.rb:91 :in `ping`
    [GEM_ROOT]/gems/activerecord-4.2.5.2/lib/active_record/connection_adapters/mysql2_adapter.rb:91 :in `active?`
    [GEM_ROOT]/gems/activerecord-4.2.5.2/lib/active_record/connection_adapters/abstract_adapter.rb:328 :in `verify!`
    [GEM_ROOT]/gems/activerecord-4.2.5.2/lib/active_record/connection_adapters/abstract/connection_pool.rb:456 :in `block in checkout_and_verify`
    [GEM_ROOT]/gems/activesupport-4.2.5.2/lib/active_support/callbacks.rb:88 :in `__run_callbacks__`
    [GEM_ROOT]/gems/activesupport-4.2.5.2/lib/active_support/callbacks.rb:778 :in `_run_checkout_callbacks`
    [GEM_ROOT]/gems/activerecord-4.2.5.2/lib/active_record/connection_adapters/abstract/connection_pool.rb:455 :in `checkout_and_verify`
    [GEM_ROOT]/gems/activerecord-4.2.5.2/lib/active_record/connection_adapters/abstract/connection_pool.rb:351 :in `block in checkout`
    /opt/rbenv/versions/2.2.3/lib/ruby/2.2.0/monitor.rb:211 :in `mon_synchronize`
    [GEM_ROOT]/gems/activerecord-4.2.5.2/lib/active_record/connection_adapters/abstract/connection_pool.rb:348 :in `checkout`
    [GEM_ROOT]/gems/activerecord-4.2.5.2/lib/active_record/connection_adapters/abstract/connection_pool.rb:263 :in `block in connection`
[...]

PR #1433, opened 2026-04-21
Fix use-after-free crash when close races with ping or query

bquorning:
## Summary

Fix a use-after-free crash when `close` is called from one thread while another thread is executing `ping`, `query`, or another operation without the GVL. Instead of calling `mysql_close` immediately (which frees the `MYSQL` struct while another thread is still using it), `close` now detects whether an operation is in-flight and takes one of two paths:

- **Idle** (no in-flight operation): call `mysql_close` directly while holding the GVL for a clean `COM_QUIT` disconnect.
- **Busy** (operation in-flight): set a `closed` flag and call `shutdown(SHUT_RDWR)` on the socket to interrupt the blocked I/O, then defer `mysql_close` to `decr_mysql2_client` (refcount == 0) where no concurrent access is possible.

**Reproducer:**
```ruby
client = Mysql2::Client.new(host: "127.0.0.1", port: 3306, username: "root")
Thread.new { client.ping }
client.close
```

This crashes with a segfault because both `nogvl_close` and `nogvl_ping` release Ruby's GVL via `rb_thread_call_without_gvl` and then call libmysql functions concurrently on the same `MYSQL *` connection. `mysql_close` frees internal state while `mysql_ping` is still accessing it.

## Why not a native mutex

A native mutex was considered but intentionally avoided: if a thread holds such a mutex at `fork()` time, the child inherits a permanently-locked mutex and any subsequent `close` or GC deadlocks. The flag + `shutdown()` approach avoids this problem entirely.

A Ruby-level `Mutex` was also [discussed and dismissed in #1284](https://github.com/brianmario/mysql2/pull/1284#issuecomment-1285657443) by @casperisfine, with valid arguments about method dispatch overhead, `Thread.handle_interrupt` handling, and behavioral changes.

## Why the connection pool cannot prevent this

The pool serializes *starting* operations at the Ruby level, but once two threads have each passed the Ruby-level checks (one doing `ping`, the other doing `close`), they are inside `rb_thread_call_without_gvl` concurrently. The pool has no visibility into that window.

## Changes

- Split `close` into idle/busy paths as described above
- Set `active_fiber` around `ping`'s `nogvl` call so `close` can detect in-flight pings
- Check `wrapper->closed` in `REQUIRE_CONNECTED`, `closed?`, and `ping` since `mysql_close` is now deferred and `CONNECTED()` remains true after `close`
- Add a doc comment to `client.c` explaining the three thread safety mechanisms (`active_fiber`, closed flag + shutdown, refcount)
- Add tests for concurrent close+ping, close+query, and fork safety

## Related issues

- #461 — "mysql2 doesn't seem to be threadsafe" — shows the exact crash pattern.
- #777 — `mysql_ping` crash with thread contention in Passenger.
- #983, #1019, #1282 — segfault reports likely caused by the same family of races.

## Test plan

- [x] Added specs for concurrent `ping`+`close`, `query`+`close`, and fork safety
- [ ] Existing test suite passes
- [ ] `rake spec:valgrind` passes

### DIFF
--- a/ext/mysql2/client.c
+++ b/ext/mysql2/client.c
@@ -1,34 +1,3 @@
-/*
- * Thread safety in mysql2
- *
- * Three mechanisms protect against concurrent access to a MYSQL connection.
- * They guard against different failure modes at different layers and are
- * not interchangeable:
- *
- * 1. active_fiber (Ruby level, GVL held)
- *    Prevents two fibers or threads from starting operations on the same
- *    connection concurrently.  Checked before rb_thread_call_without_gvl.
- *    Raises a clear Ruby error ("This connection is in use by: ...") so
- *    users can fix connection pool misuse.  ping and close intentionally
- *    skip this check — ping is a lightweight health check and close must
- *    work regardless of connection state.
- *
- * 2. mutex (C level, no GVL — rb_nativethread_lock_t)
- *    Serializes nogvl_close with other nogvl_* functions that touch the
- *    MYSQL struct.  Prevents use-after-free when close races with an
- *    in-flight operation that has already released the GVL.  The
- *    active_fiber check cannot help here because it runs before the GVL
- *    is released, leaving a window between the check and the actual
- *    libmysql call.
- *
- * 3. refcount (GC lifecycle)
- *    Tracks how many Ruby objects (Client + its Result objects) reference
- *    the mysql_client_wrapper.  A Result increments the count; GC of the
- *    Result decrements it.  The wrapper (and MYSQL struct) is only freed
- *    when the count reaches zero, preventing a GC ordering issue where
- *    Client is collected before an outstanding Result.
- */
-
 #include <mysql2_ext.h>
 
 #include <time.h>
@@ -150,7 +119,6 @@ struct nogvl_send_query_args {
 struct nogvl_select_db_args {
   MYSQL *mysql;
   char *db;
-  mysql_client_wrapper *wrapper;
 };
 
 static VALUE rb_set_ssl_mode_option(VALUE self, VALUE setting) {
@@ -391,14 +359,12 @@ static VALUE invalidate_fd(int clientfd)
 static void *nogvl_close(void *ptr) {
   mysql_client_wrapper *wrapper = ptr;
 
-  rb_native_mutex_lock(&wrapper->mutex);
   if (wrapper->initialized && !wrapper->closed) {
     mysql_close(wrapper->client);
     wrapper->closed = 1;
     wrapper->reconnect_enabled = 0;
     wrapper->active_fiber = Qnil;
   }
-  rb_native_mutex_unlock(&wrapper->mutex);
 
   return NULL;
 }
@@ -426,7 +392,6 @@ void decr_mysql2_client(mysql_client_wrapper *wrapper)
 #endif
 
     nogvl_close(wrapper);
-    rb_native_mutex_destroy(&wrapper->mutex);
     xfree(wrapper->client);
     xfree(wrapper);
   }
@@ -451,7 +416,6 @@ static VALUE allocate(VALUE klass) {
   wrapper->refcount = 1;
   wrapper->affected_rows = -1;
   wrapper->client = (MYSQL*)xmalloc(sizeof(MYSQL));
-  rb_native_mutex_initialize(&wrapper->mutex);
 
   return obj;
 }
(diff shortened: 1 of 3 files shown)


spec/mysql2/client_spec.rb
```diff
@@ -385,6 +385,24 @@ def run_gc
     end
   end
 
+  it "should not crash when ping and close are called concurrently" do
+    100.times do
+      client = new_client
+      thread = Thread.new { client.ping }
```
HeyNonster, line 391: I think if we do `client.query("SELECT SLEEP(1)")` then we can get it to consistently fail without worrying about the flakiness of `ping` executing before `client.close` does. That way we can probably remove the `100.times do`.

HeyNonster: That said, the block runs 100 times slightly faster than 1 second so `SELECT SLEEP(1)` would be moderately slower.

sodabrew:
This is great, thank you! Kicking off the CI run

byroot:
Currently at a conference so can't properly review the PR before a few days, but just based on the description I'd advise against using a native mutex, as it if the thread owning it dies (typically on fork) it could result in a deadlock of the VM.

I suspect there is a simpler fix for this race.

byroot:
So I got a few minutes to look, and yes, I believe the proper fix is to extend the use of `rb_mysql_client_set_active_fiber` to `ping` and `close` (and perhaps others?).

bquorning:
I’ll try that @byroot. Thanks for the review.

bquorning:
FYI @sodabrew CI seems to be failing because of expired signatures. I am not sure how to update them, but I suspect CI will continue to fail until that has been fixed.

spec/mysql2/client_spec.rb
```diff
@@ -385,6 +385,77 @@ def run_gc
...
+  it "should not crash when query and close are called concurrently" do
+    100.times do
+      client = new_client
+      thread = Thread.new { client.query("SELECT 1") rescue nil }
+      client.close
+      thread.join
+    end
+  end
+
+  it "should not deadlock when forking while a query is in-flight" do
```
bquorning, line 406: This spec may be too heavy-handed to keep around. Let me know if I should delete it before merge.

bquorning:
> I believe the proper fix is to extend the use of `rb_mysql_client_set_active_fiber` to `ping` and `close` (and perhaps others?).

I am not very comfortable in C and relying a lot on Claude Code (and you) to make the right decisions here 😅 What I’m getting back is that

- in `ping` we want to return true or false, but calling `rb_mysql_client_set_active_fiber` will raise an exception if the connection is already in use.
- And in the `close` method we don’t want to claim the connection. We want to check if anybody else is using it.

Instead we are adding a `NIL_P(wrapper->active_fiber)` check, which is identical to other checks already in the code.

byroot:
> * n `ping` we want to return true or false, but calling `rb_mysql_client_set_active_fiber` will raise an exception if the connection is already in use.

I believe it's fine, it's already the design of `mysql2` that concurrent usage raises. This should never happen, if it does the caller is doing something wrong.

> And in the close method we don’t want to claim the connection. We want to check if anybody else is using it.

Yes, but we also don't want anyone starting using it while we're closing it, which is achieved by setting the active fiber.

The `active_fiber` is effectively a cheaper mutex, that raise when contented.

After glancing at your PR, one problem remaining is that you must either ensure that the APIs you are calling between setting and unsetting the current fiber will never raise, or use `rb_protect` (or other API to handle exceptions)

ext/mysql2/client.c
```diff
@@ -1,3 +1,48 @@
...
+ *
+ * 1. active_fiber (Ruby level, GVL held)
+ *    Prevents two fibers or threads from starting operations on the same
+ *    connection concurrently.  Checked before rb_thread_call_without_gvl.
+ *    Raises a clear Ruby error ("This connection is in use by: ...") so
+ *    users can fix connection pool misuse.  ping returns false instead of
+ *    raising.  close intentionally skips this check — it must work
+ *    regardless of connection state.
+ *
+ * 2. closed flag + socket shutdown (C level)
```
byroot, line 16: I don't think this is necessary. We can rely on `initialized = 0`, `Mysql2::Client` can reconnect.

byroot:
Simpler fix in https://github.com/brianmario/mysql2/pull/1436

bquorning:
> Simpler fix in #1436

That looks so much better than my approach. I’ll close this PR.

bquorning:
> FYI @sodabrew CI seems to be failing because of expired signatures. I am not sure how to update them, but I suspect CI will continue to fail until that has been fixed.

Oh, a fix for the failing signatures on CI was suggested in #1424 back in February.

PR #1436, opened 2026-04-25
Prevent concurrent uses of `#ping` and `#close`

byroot:
Fix: https://github.com/brianmario/mysql2/pull/1433
    
Apply the same locking mechanism `#query` uses to these two methods.

### DIFF
--- a/ext/mysql2/client.c
+++ b/ext/mysql2/client.c
@@ -1,5 +1,6 @@
 #include <mysql2_ext.h>
 
+#include <stdbool.h>
 #include <time.h>
 #include <errno.h>
 #ifndef _WIN32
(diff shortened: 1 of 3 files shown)


bquorning:
There has been a lot of CI changes to the repo recently, so maybe a rebase on master would cause CI to pass?

byroot:
Looks like it's worse after a rebase.

bquorning:
But aren't all the test failures related to .close behavior now? I thought they were related to your changes.

byroot:
Right, the test suite was actually riddled with concurrent closing of connections held by still live threads. I cleaned all that up. It's now green.

PR #1438, opened 2026-05-03
CI: drop ubuntu-20.04 and centos:7

fkmy:
## Background
The `ubuntu-20.04` GitHub Actions runner was fully retired on 2025-04-15. Jobs that request it now stay queued indefinitely instead of failing fast, which has effectively broken every workflow entry that pinned to `ubuntu-20.04` for over a year.

Reference: [GitHub Actions: Ubuntu 20 runner image brownout dates and other breaking changes](https://github.blog/changelog/2025-01-15-github-actions-ubuntu-20-runner-image-brownout-dates-and-other-breaking-changes/)

### `.github/workflows/build.yml`
- Removed all `ubuntu-20.04` matrix entries:
  - Ruby 2.0–2.7 × `mysql80` (8 entries)
  - Ruby 2.7 × `mariadb10.6` (1 entry)
- Re-added `ubuntu-22.04, ruby: '2.7', db: mysql80` per review feedback as a smoke test for the last Ruby 2.x version. Older Ruby 2.x versions remain removed (all are upstream EOL).

### `.github/workflows/container.yml`
- Bumped the runner from `ubuntu-20.04` to `ubuntu-24.04`.
- Removed the `centos:7` matrix entry per review feedback. CentOS 7 reached upstream EOL on 2024-06-30.

### `ci/container.sh`
- Replaced `bundle install --path` with `bundle config set` syntax. The `--path` flag was removed in Bundler 4.0([rubygems/rubygems#8958](https://github.com/rubygems/rubygems/pull/8958)); Fedora latest
now ships Bundler 4.x and rejects the old syntax. This failure was masked previously because the workflow was queued indefinitely on the retired `ubuntu-20.04` runner.

## Why not migrate Ruby 2.x to ubuntu-22.04?

PR #1371 handled the previous `ubuntu-18.04` retirement by swapping the affected entry to `ubuntu-20.04`.
This PR removes the entries instead because Ruby 2.0–2.7 are all past upstream end-of-life (2.7 EOL'd 2023-03-31, others earlier). Whether mysql2 should keep CI coverage for EOL Rubies is a broader question worth discussing separately—this PR is scoped strictly to handling the retired runner.

If you'd prefer the PR #1371 approach (migrate Ruby 2.7 entries to `ubuntu-22.04`), I'm happy to update this PR—just let me know.


.github/workflows/build.yml
```diff
@@ -20,16 +20,6 @@ jobs:
           - {os: ubuntu-22.04, ruby: '3.1', db: mysql80}
           - {os: ubuntu-22.04, ruby: '3.0', db: mysql80}
 
-          # Ruby 2.x on Ubuntu 20.04 LTS
-          - {os: ubuntu-20.04, ruby: '2.7', db: mysql80}
```
sodabrew, line 24: While long last its sell-by date, I'd like to see if a last-version-of Ruby 2.x still runs:
```
          - {os: ubuntu-22.04, ruby: '2.7', db: mysql80}
```

sodabrew:
Thanks for posting this! Good cleanup. Centos 7 EOL was June 2024. Let's drop that CI job, too.

.github/workflows/build.yml
fkmy: Thanks! Added the entry back in 0a08190 under a new `# Ruby 2.x on Ubuntu 22.04 LTS` section to keep the smoke-test intent visible.

Confirmed passing in CI:
https://github.com/brianmario/mysql2/actions/runs/25270214954/job/74091220069

fkmy:
Thanks! Dropped `centos:7` and `ci/Dockerfile_centos` in 2e40267.

While we were at it, both Fedora jobs were also failing because `bundle install --path` is removed in Bundler 4.0 ([rubygems/rubygems#8958](https://github.com/rubygems/rubygems/pull/8958)).
Fixed `ci/container.sh` in 44198d3 — both Fedora jobs now pass: https://github.com/brianmario/mysql2/actions/runs/25270214954

sodabrew:
Actual green CI, it's been ages! Thank you!

image

PR #1443, opened 2026-06-13
Remove rcov configuration and rake task

fkmy:
Remove `spec/rcov.opts` and the `spec:rcov` rake task.

`RSpec::Core::RakeTask#rcov=` was deprecated in RSpec 2.99 and removed in RSpec 3.
RCov itself is not listed in the Gemfile or gemspec, so these files have been unused.

### DIFF
--- a/spec/rcov.opts
+++ b/spec/rcov.opts
@@ -1,3 +1,0 @@
---exclude spec,gem
---text-summary
---sort coverage --sort-reverse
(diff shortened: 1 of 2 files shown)


fkmy:
The ruby head job failure is not caused by this change. 
It appears to be due to eventmachine failing to build on the latest development version of Ruby.
https://github.com/brianmario/mysql2/actions/runs/27481258334/job/81229339159?pr=1443

PR #1445, opened 2026-06-22
Fix outdated information and URLs in README [ci skip]

fkmy:
- Update compatibility section to match current CI matrix
- Use HTTPS for all URLs
- Close unclosed <a> tag and unify "Mac OS X" → "macOS"


sodabrew:
Thank you!

PR #1447, opened 2026-07-06
Use numeric server version in invalid DATETIME specs

yahonda:
Fix #1446

The "should raise an error given an invalid DATETIME" specs branched on `@client.info[:version] < "8.0"`, which is wrong in two ways:

1. **MySQL version schema change**: MySQL now uses calendar versioning (YY.M), so `"26.7.0"` compares lexicographically less than `"8.0"` and the specs take the pre-8.0 branch, expecting `Mysql2::Error` to be raised — but servers with 8.0+ behavior convert the invalid CAST to `NULL` instead, so nothing is raised. Comparing the numeric `server_info[:id]` (`8.0.0` => `80000`, `26.7.0` => `260700`) keeps the comparison correct for calendar versions.

2. **Server version, not client version**: whether the invalid CAST returns the invalid value (which makes mysql2 raise) or `NULL` (8.0+ behavior) is decided by the server, but `@client.info` is `mysql_get_client_info()` — the linked client library's version. It only worked because CI installs matching client and server versions. `server_info` reflects the actually connected server.

MariaDB (verified through 12.3) still returns the invalid value, so it stays in the raise branch regardless of its numeric version.

### References

- [A More Predictable MySQL Release Model: Calendar Versions, LTS, and Innovation](https://blogs.oracle.com/mysql/a-more-predictable-mysql-release-model-calendar-versions-lts-and-innovation)
- [MySQL Community Server 26.7 Early Access Release](https://blogs.oracle.com/mysql/mysql-community-server-26-7-early-access-release)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1450, opened 2026-07-19
Cache result metadata just before freeing the result set

### LINKED ISSUE #1426: Segfault in rb_mysql_result_fetch_fields when accessing fields after result freed (0-row results)

fkmy:
## Summary

Fixes the two spec failures currently breaking CI on master. Replaces the eager field population from #1427 with caching both `fields` and `field_types` right before the C result set is freed.

Fixes #1426 for all free paths, including streaming and explicit `#free`.

## Background

The eager approach from #1427 populates only `fields`, only for non-streaming results, and does so at Result creation time. This left two specs failing on master:

- `field_types` on an exhausted 0-row result raises "Result set has already been freed" (the #1449 regression spec).
- `each(symbolize_keys: true)` returns string keys, because fields are cached at creation with the query-time options before `each` can supply its own (existing spec).

## Fix

Cache `fields` and `field_types` immediately before freeing, in every free path reachable from Ruby: full iteration, streaming completion, and explicit `Result#free`. By free time `each` has already applied its per-call options, so the eager-fetch regression disappears; the query options are only the fallback for 0-row results.

```
Before (#1427) — eager fetch at Result creation:

  query ──→ cache fields ──────→ each(opts) ──→ rows exhausted ──→ free
            (query-time options;  ↑ per-call options                ↑ field_types never
             non-streaming only)    already ignored                   cached: raises
                                                                      after free

After (this PR) — cache right before the free:

  query ──→ each(opts) ──→ rows exhausted ──→ cache fields ──→ free
                                              + field_types
                                              (per-call options already applied;
                                               same path for streaming and #free)
```

The GC free path still calls `rb_mysql_result_free_result` directly: it cannot call back into Ruby.

The first commit fixes a separate latent GC bug this change depends on: `wrapper->fieldTypes` was never registered with the GC mark/compact callbacks, so the cached array could be collected or moved while the wrapper still pointed at it.

## Trade-offs

`field_types` is no longer lazy: every fully-consumed or freed result builds the type-name strings for all columns — O(columns) small allocations per result, negligible next to per-row object creation. In the rare case metadata caching raises, the free falls back to the GC dtor, so the result set is never leaked.

## Testing

- Added specs: streaming 0-row results, explicit `#free`, and the prepared-statement path (auto-freed through the same code).
- Verified differentially: the new specs and the #1449 spec fail on master and pass with this change; same for the `symbolize_keys` spec.

The remaining red Build job on master (eventmachine vs ruby-head) is unrelated and not addressed here.


fkmy:
Note on the 3 failing jobs — none are related to this change:

- **ubuntu-24.04 ruby head**: eventmachine 1.2.7 fails to compile against ruby-head (also failing on master).
- **macos-latest ruby 4.0 mariadb@11.4**: single flaky failure in `em_spec.rb` (EventMachine close timing).
- **macos-latest ruby 4.0 mysql@8.4**: single flaky failure in `client_spec.rb:713` (`Timeout.timeout` race). The same job was already failing on master before the specs broke (the #1447 run, with a different EM spec).

nuclearspike:
Thanks for this — removing the eager population and caching at free time looks like the right shape for #1426, and it also un-breaks `each(symbolize_keys: true)` on results whose query options didn't set it (the other red spec on master).

One thing worth noting for whichever fix lands: while auditing this area I found `wrapper->fieldTypes` has never been GC-marked or compact-updated (`rb_mysql_result_mark`/`rb_mysql_result_compact` cover `fields`/`rows`/`encoding`/`client`/`statement` only). Any approach that stores the types array in the wrapper — including caching it right before free, as here — can have that array collected or moved out from under the C struct (`GC.compact`, or any GC between `#field_types` calls). I've split that fix into #1453 with a compaction regression spec; it composes with this PR rather than competing with it.

PR #1451, opened 2026-07-26
Remove Appveyor CI badge from README [ci skip]

fkmy:
- appveyor.yml was deleted in 7be15fe ("No longer using this CI service.", Sep 2024), but the badge was left behind.
- The "GitHub Actions" label is dropped as well, since there is no longer a second CI service to distinguish it from.


sodabrew:
Thank you!

PR #1454, opened 2026-08-05
Mark Result's fieldTypes so the GC can see it

jeremy:
`mysql2_result_wrapper` is xmalloc'd, so the collector neither scans it nor updates it — every `VALUE` in it has to be marked by hand. `fieldTypes` is declared in `result.h:9` but appears in neither `rb_mysql_result_mark` nor `rb_mysql_result_compact`, both of which handle exactly `fields`, `rows`, `encoding`, `client` and `statement`. Nothing else marks it, so nothing keeps the Array alive.

No compaction is involved — ordinary GC is enough.

## The window is inside the allocating call

This isn't "the Array is freed sometime later and a subsequent read is stale". It's freed during the call that creates it:

1. `rb_mysql_result_fetch_field_types` allocates the Array and stores it in `wrapper->fieldTypes`
2. the fill loop calls `rb_mysql_result_fetch_field_type` per column, which allocates Strings (`rb_sprintf` / `rb_str_new_cstr` / `rb_str_export_to_enc`) and can trigger a GC
3. `rb_ary_store(wrapper->fieldTypes, idx, ...)` then writes through a freed object slot

So the failure isn't only a bad read — step 3 is a write into whatever now occupies that slot.

## Reproducing

Public API only, no `Fiddle` and no raw memory access, so a crash can only come from the extension:

```ruby
require "mysql2"
client = Mysql2::Client.new(host: "127.0.0.1", username: "root")
res = client.query("select 1 as a, 'x' as b, 2.5 as c, now() as d")
GC.stress = true
types = res.field_types      # first-ever call on this Result
GC.stress = false
raise "not an Array: #{types.class}" unless types.is_a?(Array)
```

Built both trees inside the same step that ran the test, so the artifact can't drift from the source:

```
                     master     with this patch
GC.stress, 1 call    3/3 abort  3/3 pass
GC off (control)     3/3 pass   3/3 pass
```

`GC.stress` only makes it deterministic — it isn't required. Looping ordinary `field_types` calls with allocation churn between them and no stress, no `GC.start` and no compaction, master segfaults **within the first 25 calls, 5/5**, while the patched build completes 60,000 calls 3/3 (180,000 total). The exact rate depends on the surrounding allocation profile; a busy app supplies plenty.

## Affected

Present since `c3d1abc1e` (2020-03-16, #1068), which added `field_types` without extending the mark function. So 0.5.4, 0.5.5, 0.5.6, 0.5.7 and master. Tested against master `d4f8c13`, Ruby 4.0.6, MySQL 9.6.

## The patch

Two lines, matching the existing style in each function — `rb_gc_mark_movable` in the mark function and `rb_mysql2_gc_location` in the compact function, so `fieldTypes` stays movable like its five neighbours rather than being pinned.

Found while sweeping native gems for dangling `VALUE`s. Happy to add a regression test if you'd like one, though it needs `GC.stress` to be deterministic.

### DIFF
--- a/ext/mysql2/result.c
+++ b/ext/mysql2/result.c
@@ -70,6 +70,7 @@ static void rb_mysql_result_mark(void * wrapper) {
   mysql2_result_wrapper * w = wrapper;
   if (w) {
     rb_gc_mark_movable(w->fields);
+    rb_gc_mark_movable(w->fieldTypes);
     rb_gc_mark_movable(w->rows);
     rb_gc_mark_movable(w->encoding);
     rb_gc_mark_movable(w->client);
@@ -152,6 +153,7 @@ static void rb_mysql_result_compact(void * wrapper) {
   mysql2_result_wrapper * w = wrapper;
   if (w) {
     rb_mysql2_gc_location(w->fields);
+    rb_mysql2_gc_location(w->fieldTypes);
     rb_mysql2_gc_location(w->rows);
     rb_mysql2_gc_location(w->encoding);
     rb_mysql2_gc_location(w->client);


jeremy:
### CI is red, and all 20 failures are pre-existing on master

This branch changes two lines in `ext/mysql2/result.c` and nothing else (`git diff --stat d4f8c13 3ee2053` → `ext/mysql2/result.c | 2 ++`). Here is why the checks are red anyway, in a form you can check without taking my word for it.

**The 20 failing checks on `3ee20534d9112441e1bfed337744a6120a807a62`:**

| Workflow | Failing job(s) |
|---|---|
| RuboCop | `build` |
| Container | `fedora fedora:latest`, `fedora fedora:rawhide` |
| Build | `macos-latest ruby 2.6 mysql@8.0`, `macos-latest ruby 4.0 mariadb@11.4`, `macos-latest ruby 4.0 mysql@8.4`, `ubuntu-22.04 ruby 2.7 mariadb10.11`, `ubuntu-22.04 ruby 2.7 mysql80`, `ubuntu-22.04 ruby 3.0 mariadb10.11`, `ubuntu-22.04 ruby 3.0 mariadb10.6`, `ubuntu-22.04 ruby 3.0 mysql80`, `ubuntu-22.04 ruby 3.1 mysql80`, `ubuntu-22.04 ruby 3.2 mysql80`, `ubuntu-22.04 ruby 3.3 mysql80`, `ubuntu-22.04 ruby 3.4 mysql80`, `ubuntu-24.04 ruby 3.4 mariadb11.4`, `ubuntu-24.04 ruby 3.4 mysql84`, `ubuntu-24.04 ruby 4.0 mariadb11.4`, `ubuntu-24.04 ruby 4.0 mysql84`, `ubuntu-24.04 ruby head mysql84` |

**They fail on the base too.** The PR's base commit is `d4f8c130bd34a5026f95edadcfaf6801db1dceef` ("Remove Appveyor CI badge from README [ci skip] (#1451)"). Because of the `[ci skip]`, that commit has no Build/RuboCop/Container runs of its own — the only check attached to it is `copilot-pull-request-reviewer`. Its parent, `1aa1e6673813d4a6dd7cb1d582f84bf03c9f8091`, does have all three, and `git diff --stat 1aa1e667 d4f8c13` is `README.md | 3 ---` — README only, no code, no specs, no workflows. So master's last CI-bearing commit before this branch is the honest baseline:

- RuboCop — https://github.com/brianmario/mysql2/actions/runs/28830773462 — job `build` failed
- Build — https://github.com/brianmario/mysql2/actions/runs/28830773514 — the same 17 matrix jobs failed
- Container — https://github.com/brianmario/mysql2/actions/runs/28830773450 — `fedora fedora:latest` and `fedora fedora:rawhide` failed

Same 20 job names, red on master on 2026-07-06, before this branch existed.

**Two distinct causes, both in `spec/`, neither touched here.**

RuboCop, at the base:

```
spec/mysql2/result_spec.rb:3:1: C: Metrics/BlockLength: Block has too many lines. [483/477]
36 files inspected, 1 offense detected
```

Every Build and Container job, at the base:

```
353 examples, 2 failures, 3 pending
rspec ./spec/mysql2/result_spec.rb:60  # Mysql2::Result#each should yield rows as hash's with symbol keys if :symbolize_keys was set to true
rspec ./spec/mysql2/result_spec.rb:125 # Mysql2::Result#fields should keep fields and field_types accessible for exhausted empty results
```

And on this branch's head — identical, same count, same two examples:

```
353 examples, 2 failures, 3 pending
rspec ./spec/mysql2/result_spec.rb:60  # Mysql2::Result#each should yield rows as hash's with symbol keys if :symbolize_keys was set to true
rspec ./spec/mysql2/result_spec.rb:125 # Mysql2::Result#fields should keep fields and field_types accessible for exhausted empty results
```

So the patch neither fixes nor breaks those two. Worth saying plainly: `result_spec.rb:125` asserts on `field_types`, so it is fair to wonder whether it is related. It is not — it fails the same way with and without the patch, and it was failing on master a month before this branch.

### Separately: the patch itself, verified 3/3 red → green

This is a different run from the CI above, and it was done from the exact commits this PR points at, not from a working copy. Both trees were exported with `git archive` from the pushed objects, and their tree SHAs match what GitHub reports for those commits:

| Arm | Commit | Tree SHA |
|---|---|---|
| red (base, unpatched) | `d4f8c130bd34a5026f95edadcfaf6801db1dceef` | `4f0f521040c0a24de2356db2bcad993704cbf502` |
| green (this PR's head) | `3ee20534d9112441e1bfed337744a6120a807a62` | `16b253073681c627a26949b72503188b60c26040` |

Both were compiled and exercised inside a single container step, so the `.so` under test cannot drift from the source it claims to be. Ruby 4.0.6 (2026-07-14 revision 03b6d3f889) +PRISM [aarch64-linux], MySQL 9.6.0. The reproducer uses only public API — `client.query(...)`, then one first-ever `field_types` call with `GC.stress` on across just that call — no Fiddle, no raw memory reads, so a crash can only come from the extension.

```
BUILD OK   red    so-sha256=c49100ef318a2eb5  result.c-sha256=4764267ae16f7607
BUILD OK   green  so-sha256=99bdc4e9be94f293  result.c-sha256=3ab42507a9b6572e

red    stress run1 exit=134  Aborted
red    stress run2 exit=134  Aborted
red    stress run3 exit=134  Aborted
red    CONTROL(GC=off) exit=0   SURVIVED
green  stress run1 exit=0    SURVIVED
green  stress run2 exit=0    SURVIVED
green  stress run3 exit=0    SURVIVED
green  CONTROL(GC=off) exit=0   SURVIVED
```

The `GC=off` control is the part that matters for interpreting the red arm: with the collector disabled the unpatched build completes normally, so the crash is the GC reclaiming the Array, not a broken build or a bad query.

The red arm aborts at the `field_types` call itself:

```
mysql2_repro.rb:46: [BUG] rb_gc_mark(): unknown data type 0x18(0x0000ffffa0c0f968) corrupted object
ruby 4.0.6 (2026-07-14 revision 03b6d3f889) +PRISM [aarch64-linux]
...
mysql2_repro.rb:46:in '<main>'
```

where line 46 is `types = res.field_types`.

jeremy:
Closing in favour of #1453, which beat me to it by eight days and is a strict superset — the `rb_gc_mark_movable`/`rb_mysql2_gc_location` hunks there are character-identical to this diff, and it additionally carries the `resultFreed` guards, the replayability check in `_each`, and specs including a `field_types`-across-compaction regression test.

I've moved my analysis and reproduction data over to #1453 rather than leave it stranded here.

PR #1477, opened 2026-08-09
CI: Cancel superseded runs, cap job runtime, and drop a redundant brew search

fkmy:
## Cancel superseded pull request runs

Neither `Build` nor `RuboCop` declared a `concurrency` group, so a fixup push left the previous run going to completion. Of the last 60 `Build` runs on pull requests, 11 were still in flight when a newer push obsoleted them -- the longest overlap ran over five hours of the ~29-job matrix.

Grouping by workflow and ref cancels those. Master is exempt, since every commit that lands there should get its own result. `build.yml` is the caller for the four reusable workflows, so one group covers the whole fan-out.

## Add a 30 minute job timeout

Jobs finish in 2 to 3 minutes, but with no `timeout-minutes` they inherit GitHub's 360 minute default. On `ci-always-tls-tcp`, `macos-latest ruby 4.0 mariadb@11.4` hung and held a macOS runner for the full 360 minutes, on three consecutive runs ([one](https://github.com/brianmario/mysql2/actions/runs/31247515906)).

`build.yml`'s `uses:` jobs don't accept `timeout-minutes`, so it goes on the job inside each called workflow.

## Drop the brew search survey from the macOS setup

`ci/setup.sh` listed every mysql\* and mariadb\* formula before installing -- a question about Homebrew, not about this build, asked identically on every macOS job of every run. `brew info "$DB"` stays for the resolved version.

## Test plan

- [x] `actionlint` -- no findings on any changed line (the remaining notes are pre-existing, on untouched lines)
- [x] All six workflow files parse as YAML at each of the three commits
- [x] `bash -n ci/setup.sh` -- clean

### DIFF
--- a/.github/workflows/build-cross-version.yml
+++ b/.github/workflows/build-cross-version.yml
@@ -14,6 +14,7 @@ jobs:
       ${{ matrix.os }} ruby ${{ matrix.ruby }} ${{ matrix.db }}
     runs-on: ${{ matrix.os }}
     services: ${{ matrix.services }}
+    timeout-minutes: 30
     strategy:
       matrix:
         include:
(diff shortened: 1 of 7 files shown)


sodabrew:
Nice! Will land this in later today

PR #1478, opened 2026-08-10
CI: Swap apt-key for Signed-By in MySQL, align MariaDB scripts

fkmy:
`ci/mysql80.sh`, `ci/mysql84.sh` and `ci/mysql97.sh` install the MySQL signing key with `apt-key add`. That command is deprecated, and Ubuntu 26.04 does not ship it at all -- `command -v apt-key` finds nothing there, where it resolves to `/usr/bin/apt-key` on 24.04. The 22.04 and 24.04 runners the matrix uses today still have it, so nothing is broken right now; this is done ahead of the runner images moving on rather than in response to a failure.

Each script now writes a deb822 `.sources` file that points at a key under `/etc/apt/keyrings` via `Signed-By`, the shape `ci/mariadb*.sh` has used since #1401. `apt-key` put the key in the trusted set for every configured repository, while `Signed-By` scopes it to this one.

Naming the suite and the component explicitly also lets `add-apt-repository` go away, and with it a second full package-list refresh: without `--no-update` it runs an `apt-get update` of its own, on top of the one already in the script. The repository URI moves to https while here.

## Dropping two of the three keys

`mysql80.sh` installed three keys, but only `B7B3B788A8D3785C` signs `repo.mysql.com`'s `InRelease` today on either jammy or noble -- `gpgv` accepts that key and rejects the other two. `5072E1F5` and `3A79BD29` had expired as well, in 2022 and 2023. The second commit deletes both `.asc` files now that nothing installs them.

`B7B3B788A8D3785C` expires on 2027-10-23, so this does not remove the need to refresh the key -- it just reduces the refresh to swapping one file.

## Aligning the MariaDB scripts

Per review, the five `ci/mariadb*.sh` scripts get the same treatment: `install -d -m 0755 /etc/apt/keyrings` before the key is written, and `install -m 0644` in place of a plain `cp`. The key also moves ahead of the `.sources` file that references it, so both families read the same way.

A separate commit drops the redundant `>` from the heredoc in all eight scripts -- `tee` was given no filename, so the trailing redirect was doing the write. One visible effect: `tee` now duplicates the heredoc to stdout as well, so the generated `.sources` shows up in the job log.

## Test plan

Each script was run in the Ubuntu image its matrix entry uses, with the `purge` line removed (there is nothing to purge in a bare image) and the install switched to `--dry-run`:

| script | image | resolves |
| --- | --- | --- |
| `ci/mysql80.sh` | `ubuntu:22.04` | `8.0.46-1ubuntu22.04` from `repo.mysql.com` |
| `ci/mysql84.sh` | `ubuntu:24.04` | `8.4.11-1ubuntu24.04` from `repo.mysql.com` |
| `ci/mysql97.sh` | `ubuntu:24.04` | `9.7.2-1ubuntu24.04` from `repo.mysql.com` |
| `ci/mariadb106.sh` | `ubuntu:22.04` | `1:10.6.27+maria~ubu2204` from `mirror.rackspace.com` |
| `ci/mariadb1011.sh` | `ubuntu:22.04` | `1:10.11.18+maria~ubu2204` from `mirror.rackspace.com` |
| `ci/mariadb114.sh` | `ubuntu:24.04` | `1:11.4.12+maria~ubu2404` from `mirror.rackspace.com` |
| `ci/mariadb11.8.sh` | `ubuntu:24.04` | `1:11.8.8+maria~ubu2404` from `mirror.rackspace.com` |
| `ci/mariadb12.3.sh` | `ubuntu:24.04` | `1:12.3.2+maria~ubu2404` from `mirror.rackspace.com` |

- [x] All eight exit 0, with no `W:`, `NO_PUBKEY` or `EXPKEYSIG` out of the update step
- [x] `bash -n` and `shellcheck` clean on all eight

Ubuntu 26.04 is not in the matrix, so the claim about it rests on the `apt-key` check below rather than on a full run:

```console
$ docker run --rm ubuntu:24.04 bash -c 'command -v apt-key'
/usr/bin/apt-key
$ docker run --rm ubuntu:26.04 bash -c 'command -v apt-key'; echo "exit=$?"
exit=1
```

Ref: https://manpages.debian.org/bookworm/apt/apt-key.8.en.html


sodabrew:
Nice work! The `Signed-By` switch and the key cleanup are both the right call, and I like that this scopes trust down to the one repo instead of just papering over the deprecation warning.

Two small follow-ups:

- Could you carry a couple of these improvements back into the mariadb scripts too (`ci/mariadb1011.sh`, `ci/mariadb106.sh`, `ci/mariadb114.sh`, `ci/mariadb11.8.sh`, `ci/mariadb12.3.sh`), since they predate this pattern and are now the odd ones out?
  - `install -d -m 0755 /etc/apt/keyrings` before writing the key — the mariadb scripts currently assume that directory already exists. More portable/self-contained.
  - `install -m 0644 support/<key>.asc /etc/apt/keyrings/<name>-keyring.asc` instead of a plain `cp` — explicit permissions and an explicit rename, clearer than what's there today.

- Separately, in both the scripts you just touched and the existing mariadb ones: `tee <<- EOF > file` is functionally just `cat <<-EOF > file` here — `tee` is given no filename argument, so it only duplicates stdin to stdout, and the actual write happens via the trailing `>` redirect. Not a bug, just leftover style. `tee /etc/apt/sources.list.d/<name>.sources <<-EOF` (drop the redundant `>`) is the more idiomatic form.

Thanks for the clear write-up and the dry-run test plan, by the way, it made this easy to verify!

fkmy:
Thanks for the review, both follow-ups are in.
Glad to have all eight consistent now, it should make them easier to keep in sync.

PR #1405, opened 2025-07-14
adds tls sni support to mysql2

mikemackintosh:
This pull request introduces support for specifying a TLS SNI (Server Name Indication) name during MySQL client connections. This enhancement allows users to set a custom SNI hostname for TLS connections, improving flexibility and compatibility with certain server configurations.

### Enhancements to TLS Configuration:

* Updated the `rb_mysql_connect` function in `ext/mysql2/client.c` to accept an additional parameter, `tls_sni_name`, and use it to set the TLS SNI server name via the `MYSQL_OPT_TLS_SNI_SERVERNAME` option in the MySQL client library. [[1]](diffhunk://#diff-8cc85b931218589827ff4097193b14289b4d0fd58f28e95a15c6bd63cdaf5ed8L507-R510) [[2]](diffhunk://#diff-8cc85b931218589827ff4097193b14289b4d0fd58f28e95a15c6bd63cdaf5ed8R523-R534)
* Modified the `init_mysql2_client` function in `ext/mysql2/client.c` to reflect the updated `rb_mysql_connect` method signature, increasing the number of expected arguments from 8 to 9.

### Updates to Ruby Client Initialization:

* Added support for the `:tls_sni_name` option in the `initialize` method of `lib/mysql2/client.rb`, allowing users to specify the SNI name when creating a MySQL client instance.
* Ensured the `tls_sni_name` value is converted to a string if provided, and passed it as an argument to the updated `connect` method.

PR #1479, opened 2026-08-10
Fix HAVE_CONST_MYSQL_DEFAULT_AUTH and SSL-mode version specs

sodabrew:
Both bugs found here are the same shape: a silent logic error that always took the wrong branch, with no crash and no visible signal.

Ruby's mkmf `have_const(const, header)` always generates a macro named `HAVE_CONST_<NAME>`, never `HAVE_<NAME>` -- that naming is what `have_func` generates instead. `extconf.rb` calls `have_const('MYSQL_DEFAULT_AUTH', mysql_h)`, but `client.c` guarded on `HAVE_MYSQL_DEFAULT_AUTH` (missing the `_CONST_` infix) in two places, so the guard was always false and `set_default_auth` always raised "pluggable authentication is not available" regardless of what the linked client library actually supported.

The SSL-mode spec's version-branch had a similar problem: it passed `Mysql2::Client.info` (a Hash: `{id:, version:, header_version:}`) directly into a `case/when` block matching against numeric Ranges. `Range#===` against a Hash silently returns false rather than raising, so the branch meant to enable `preferred`/`verify_ca` on MySQL 5.6.36+/5.7.11+/8.0+ never actually ran on any CI leg -- every run silently fell through to the narrower `else` branch. Fixed by using `Mysql2::Client.info[:id]`, the actual integer. This also uncovered a `verifa_ca` typo (should be `verify_ca`) sitting right next to it in the array literal.

Both found incidentally while working on unrelated PR #1405 (TLS SNI support) in this session.

Bugs:
- `HAVE_MYSQL_DEFAULT_AUTH` in client.c never matched the `HAVE_CONST_*` macro extconf.rb actually generates, so `set_default_auth()` was permanently disabled regardless of client library support.
- The SSL-mode spec compared a Range against a Hash (`Client.info`) instead of the Hash's `:id` value, so `preferred`/`verify_ca` were never exercised by CI on any MySQL version, and a `verifa_ca` typo went unnoticed alongside it.

Fixes:
- `ext/mysql2/client.c`: `HAVE_MYSQL_DEFAULT_AUTH` -> `HAVE_CONST_MYSQL_DEFAULT_AUTH` at both occurrences, matching the working sibling guards right next to them (`HAVE_CONST_MYSQL_OPT_GET_SERVER_PUBLIC_KEY`, `HAVE_CONST_MYSQL_ENABLE_CLEARTEXT_PLUGIN`).
- `spec/mysql2/client_spec.rb`: use `Mysql2::Client.info[:id]` for the version comparison, and fix `verifa_ca` -> `verify_ca`.

Notes:
- Audited every other `have_const` call in extconf.rb against every `HAVE_*` guard in the C extension; found no other instances of this mistake.
- Verified end-to-end, not just that it compiles: built a dedicated SSL-enabled MySQL instance and confirmed `preferred`/`verify_ca` connect successfully now that the branch is actually reachable.

### DIFF
--- a/ext/mysql2/client.c
+++ b/ext/mysql2/client.c
@@ -1245,7 +1245,7 @@ static VALUE _mysql_client_options(VALUE self, int opt, VALUE value) {
       break;
 #endif
 
-#ifdef HAVE_MYSQL_DEFAULT_AUTH
+#ifdef HAVE_CONST_MYSQL_DEFAULT_AUTH
     case MYSQL_DEFAULT_AUTH:
       charval = (const char *)StringValueCStr(value);
       retval  = charval;
@@ -1782,7 +1782,7 @@ static VALUE set_get_server_public_key(VALUE self, VALUE value) {
 }
 
 static VALUE set_default_auth(VALUE self, VALUE value) {
-#ifdef HAVE_MYSQL_DEFAULT_AUTH
+#ifdef HAVE_CONST_MYSQL_DEFAULT_AUTH
   return _mysql_client_options(self, MYSQL_DEFAULT_AUTH, value);
 #else
   rb_raise(cMysql2Error, "pluggable authentication is not available, you may need a newer MySQL client library");
--- a/spec/mysql2/client_spec.rb
+++ b/spec/mysql2/client_spec.rb
@@ -177,10 +177,10 @@ RSpec.describe Mysql2::Client do # rubocop:disable Metrics/BlockLength
     end
 
     # 'preferred' or 'verify_ca' are only in MySQL 5.6.36+, 5.7.11+, 8.0+
-    version = Mysql2::Client.info
+    version = Mysql2::Client.info[:id]
     ssl_modes = case version
     when 50636...50700, 50711...50800, 80000...90000
-      %i[disabled preferred required verifa_ca verify_identity]
+      %i[disabled preferred required verify_ca verify_identity]
     else
       %i[disabled required verify_identity]
     end
--- a/spec/mysql2/client_spec.rb
+++ b/spec/mysql2/client_spec.rb
@@ -176,10 +176,15 @@ RSpec.describe Mysql2::Client do # rubocop:disable Metrics/BlockLength
       new_client(option_overrides)
     end
 
-    # 'preferred' or 'verify_ca' are only in MySQL 5.6.36+, 5.7.11+, 8.0+
+    # 'preferred' or 'verify_ca' are only in MySQL 5.6.36+, 5.7.11+, 8.0+.
+    # The upper bound stops at 100000 (not left unbounded) because MariaDB's
+    # client version numbering starts there (10.x = 100000+, 11.x = 110000+,
+    # 12.x = 120000+) and MariaDB has never implemented the 5-value ssl_mode
+    # API this branch is testing -- see FULL_SSL_MODE_SUPPORT in extconf.rb
+    # and rb_set_ssl_mode_option's own version >= 100000 MariaDB check.
     version = Mysql2::Client.info[:id]
     ssl_modes = case version
-    when 50636...50700, 50711...50800, 80000...90000
+    when 50636...50700, 50711...50800, 80000...100000
       %i[disabled preferred required verify_ca verify_identity]
     else
       %i[disabled required verify_identity]


sodabrew:
## CI verification: both `ssl_mode` fixes confirmed across the matrix

Compared this PR's latest CI run ([`b33f4b5`](https://github.com/brianmario/mysql2/actions/runs/31445723496)) against a baseline at the current tip of `master` ([`b997925`](https://github.com/brianmario/mysql2/commit/b9979257cc19275209ec5cb003e20b6fdca66c8f), same commit this PR is based on), to confirm both fixes in `client_spec.rb` actually change which `ssl_mode` tests run, not just that the suite stays green. Pulled directly from each job's raw log (`should set ssl_mode option ...` lines + the final `rspec` summary line).

Two fixes landed here:
1. `Mysql2::Client.info` returns `{id:, version:, header_version:}`, but the spec compared the whole Hash against numeric `Range`s. `Range#===` against a Hash silently returns `false`, so every leg fell through to the 3-mode `else` branch regardless of version — fixed by using `.info[:id]`.
2. The range's upper bound (`80000...90000`) was written when 8.0 was the newest MySQL and was never extended — MySQL 9.7's `info[:id]` (90702) fell just outside it. Extended to `80000...100000`, confirmed safe against MariaDB by checking this PR's own CI mkmf output: every MariaDB leg (10.6 through 12.3) fails the very first `have_const` check (`SSL_MODE_DISABLED`) before probing the rest, so MariaDB never gets `FULL_SSL_MODE_SUPPORT` regardless of where the upper bound sits — no risk of the wider range accidentally capturing it.

| Client library | `info[:id]` | Before (master tip) | After (this PR) |
|---|---|---|---|
| MySQL 8.0.46 | 80046 | 3 modes: `disabled`, `required`, `verify_identity` — 398 examples | **5 modes**: + `preferred`, `verify_ca` — 400 examples |
| MySQL 8.4.11 | 80411 | 3 modes — 398 examples | **5 modes** — 400 examples |
| MySQL 9.7.2 | 90702 | 3 modes — 398 examples | **5 modes** — 400 examples |
| MariaDB 10.6.27 | ~100627 | 3 modes — 398 examples | 3 modes — 398 examples *(unchanged, correct — no `SSL_MODE_*` support at all)* |
| MariaDB 10.11.18 | ~101118 | 3 modes — 398 examples | 3 modes — 398 examples *(unchanged, correct)* |
| MariaDB 11.4.12 | ~110412 | 3 modes — 398 examples | 3 modes — 398 examples *(unchanged, correct)* |
| MariaDB 11.8.8 | ~110808 | 3 modes — 398 examples | 3 modes — 398 examples *(unchanged, correct)* |
| MariaDB 12.3.2 | ~120302 | 3 modes — 398 examples | 3 modes — 398 examples *(unchanged, correct)* |

### Raw evidence — MySQL 8.0 (bug 1 taking effect)

Before, [master tip / `ubuntu-22.04 ruby 3.3 mysql80`](https://github.com/brianmario/mysql2/actions/runs/31443414172/job/93632601866):
```
should set ssl_mode option required
should set ssl_mode option disabled
should set ssl_mode option verify_identity
398 examples, 0 failures, 8 pending
```

After, [this PR / `ubuntu-22.04 ruby 3.3 mysql80`](https://github.com/brianmario/mysql2/actions/runs/31445723496/job/93639402628):
```
should set ssl_mode option verify_identity
should set ssl_mode option required
should set ssl_mode option preferred
should set ssl_mode option disabled
should set ssl_mode option verify_ca
400 examples, 0 failures, 8 pending
```

### Raw evidence — MySQL 9.7 (bug 2 taking effect)

Before, [master tip / `ubuntu-24.04 ruby 4.0 mysql97`](https://github.com/brianmario/mysql2/actions/runs/31443414172/job/93632601835):
```
should set ssl_mode option verify_identity
should set ssl_mode option disabled
should set ssl_mode option required
398 examples, 0 failures, 8 pending
```

After, [this PR / `ubuntu-24.04 ruby 4.0 mysql97`](https://github.com/brianmario/mysql2/actions/runs/31445723496/job/93639402682):
```
should set ssl_mode option preferred
should set ssl_mode option disabled
should set ssl_mode option verify_identity
should set ssl_mode option verify_ca
should set ssl_mode option required
400 examples, 0 failures, 8 pending
```

### Raw evidence — MariaDB 11.8 (no regression from widening the range)

Before, [master tip / `ubuntu-24.04 ruby 3.4 mariadb11.8`](https://github.com/brianmario/mysql2/actions/runs/31443414172/job/93632601850):
```
should set ssl_mode option verify_identity
should set ssl_mode option required
should set ssl_mode option disabled
398 examples, 0 failures, 7 pending
```

After, [this PR / `ubuntu-24.04 ruby 3.4 mariadb11.8`](https://github.com/brianmario/mysql2/actions/runs/31445723496/job/93639402688):
```
should set ssl_mode option required
should set ssl_mode option disabled
should set ssl_mode option verify_identity
398 examples, 0 failures, 7 pending
```

Same pattern held for MariaDB 10.6, 10.11, 11.4, and 12.3 — spot-checked 12.3 directly, all consistent: 3 modes, 398 examples, unchanged before/after.

All 33 CI legs pass on the current commit.

PR #1483, opened 2026-08-11
Build Result#server_flags lazily from status captured at creation

jeremy:
Every Result construction on master eagerly builds a three-key hash -- rb_hash_new plus three rb_hash_aset and an ivar write -- to populate `#server_flags`, an accessor most applications never read. That is pure per-Result overhead on the hot query path.

This defers the hash. `rb_mysql_result_to_obj` now copies the connection's `server_status` into an `unsigned int` on the result wrapper -- a plain struct field copy that cannot raise, so it is safe inside the no-raise window between the network command and `active_streaming_result` registration (#1463). `Result#server_flags` becomes a C method that builds the same hash on first call from that captured status -- same keys, same true/false values, same nil where the client library lacks a flag constant at compile time -- then memoizes it in a `VALUE server_flags` slot on the result wrapper, marked from `rb_mysql_result_mark` with `rb_gc_mark_movable` and updated in the compact callback, exactly like the wrapper's other VALUEs. A struct slot rather than an ivar because an ivar write raises FrozenError on a frozen Result, while master's frozen Results answered `#server_flags` fine (the ivar predated any freeze); specs pin both freeze-before-first-access and freeze-after-first-access. Returning the same object on later calls preserves master's hash-identity and mutation-visibility semantics, and the hash itself stays mutable, as on master.

Capturing at creation rather than reading live status at call time is load-bearing: by the time flags are read, the connection may have run other statements. A spec pins this with a MULTI_STATEMENTS batch whose two statements differ in SERVER_QUERY_NO_INDEX_USED (no-table SELECT, then a full scan), reading the first result's flags only after the second has been stored. Under a call-time read of live status -- demonstrated by temporarily pointing the flag macro at `wrapper->client_wrapper->client->server_status` and recompiling -- that spec fails exactly as predicted: the first result reports the second statement's full-scan flag. The prepared-statement spec fails the same way in the opposite direction (full scan first, flags read after a later no-table execute).

Behavior change

Results from `Client#store_result` (the multi-result-set path) previously returned nil from `#server_flags`: master set the ivar only in the async-query and prepared-statement paths, and store_result had no spec coverage. Capture now lives in `rb_mysql_result_to_obj`, the single choke point all three creation paths share (client.c async_result, client.c store_result, statement.c execute), so store_result results now report real flags; a spec pins that. This is a deliberate change for maintainers to accept, not an accident of the refactor.

Removed: `rb_mysql_set_server_query_flags` (client.c/client.h), its two call sites, and the Ruby-side attr_reader.

New specs: hash identity across calls, frozen-Result access (freeze before first read; same object after freeze), availability after full iteration and after #free, the multi-result-set capture pin, and the prepared-statement pin.

Benchmarks: per-Result-creation overhead -- small result sets issued at high rate, where the eager hash is proportionally largest. The `query + server_flags` arm is the deliberate worst case for laziness: every result's flags actually read, paying construction at access time instead of skipping it. Min of 9 runs per sample, 9 samples per arm, arms interleaved; each build compiled and measured separately.

thelio -- AMD Ryzen 9 7950X (32 threads), Arch Linux, Ruby 4.0.6, MySQL 9.6.0 in Docker (TCP loopback, port 3306), client libmariadb 12.3.2, load < 1.0 for every run. Three alternating base/branch pairs, medians of medians (ms per 200 queries):

    arm                     base (82f9e18)   this branch
    query SELECT 1          5.60             5.61
    query 10x3 rows         7.60             7.52
    stmt execute            5.42             5.40
    query + server_flags    5.57             5.62

All arms within +/-1%. Allocations: 8.001 -> 7.001 objects/query with flags unread; 8.002 -> 8.002 with flags read.

macOS arm64, Ruby 4.0.6, MySQL 9.6 (Homebrew, libmysqlclient.24) over a Unix socket. Two clean base runs and two clean branch runs, alternating; a contaminated run was discarded and its samples excluded (median / median, min-max in ms per 200 queries):

    arm                     base (82f9e18)              this branch
    query SELECT 1          3.73 / 3.69  (3.61-5.22)    3.70 / 3.67  (3.61-4.08)
    query 10x3 rows         6.02 / 5.85  (4.72-6.38)    5.88 / 5.86  (5.77-6.40)
    stmt execute            3.11 / 3.10  (2.45-3.50)    3.12 / 3.16  (3.01-3.56)
    query + server_flags    3.69 / 3.71  (3.55-3.92)    3.73 / 3.76  (3.62-4.15)

Both platforms show wall time flat within noise on every arm, worst case included: the eager hash was sub-microsecond work against the round trip to the server, so no end-to-end query benchmark can resolve it. The deterministic win is allocation count, measured via GC.stat(:total_allocated_objects) over 1000 `SELECT 1` queries per shape (macOS: 8.0 -> 7.0 objects/query flags-unread; 8.0 -> 8.0 flags-read).

Flags go unread in virtually every real workload (no widely-used caller reads server_flags), so the common case saves the hash and its three entries; readers pay exactly what they paid before, just at access time.

Implemented and benchmarked at base 82f9e18e65eabf3755b482feaf80c72d5f6f0ad8, then rebased onto 96252dcc575e8626026c1e823e06a3d25739d915 -- a CI-only change (apt-key to Signed-By swap, MariaDB script alignment) touching no library code, so the benchmarks remain valid. Full suite re-run green on the rebased head, exit status checked: 412 examples, 0 failures, 10 pending (SSL). Not tested locally: MariaDB server, Windows, other Rubies; the missing-constant #else branches compile away on these platforms and are verified by inspection only.

### DIFF
--- a/ext/mysql2/result.c
+++ b/ext/mysql2/result.c
@@ -64,7 +64,8 @@ static ID intern_new, intern_utc, intern_local, intern_localtime, intern_local_o
   intern_query_options;
 static VALUE sym_symbolize_keys, sym_as, sym_array, sym_database_timezone,
   sym_application_timezone, sym_local, sym_utc, sym_cast_booleans,
-  sym_cache_rows, sym_cast, sym_stream, sym_name, sym_rows_per_gvl_yield;
+  sym_cache_rows, sym_cast, sym_stream, sym_name, sym_rows_per_gvl_yield,
+  sym_no_good_index_used, sym_no_index_used, sym_query_was_slow;
 
 /* Mark any VALUEs that are only referenced in C, so the GC won't get them. */
 static void rb_mysql_result_mark(void * wrapper) {
@@ -76,6 +77,7 @@ static void rb_mysql_result_mark(void * wrapper) {
     rb_gc_mark_movable(w->encoding);
     rb_gc_mark_movable(w->client);
     rb_gc_mark_movable(w->statement);
+    rb_gc_mark_movable(w->server_flags);
   }
 }
 
@@ -221,6 +223,7 @@ static void rb_mysql_result_compact(void * wrapper) {
     rb_mysql2_gc_location(w->encoding);
     rb_mysql2_gc_location(w->client);
     rb_mysql2_gc_location(w->statement);
+    rb_mysql2_gc_location(w->server_flags);
   }
 }
 #endif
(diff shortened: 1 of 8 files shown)


ext/mysql2/result.c
```diff
@@ -1604,6 +1607,53 @@ static VALUE rb_mysql_result_each(int argc, VALUE * argv, VALUE self) {
...
+    /* Memoize in the wrapper struct, marked from rb_mysql_result_mark: a
+     * plain C field write, so it works even on a frozen Result (an ivar set
+     * would raise FrozenError), and later calls return this same Hash object
+     * (mutations included), as the eager version did. */
+    wrapper->server_flags = server_flags;
+  }
+
+  return wrapper->server_flags;
+}
+#undef flag_to_bool
```
sodabrew, line 1655: Nit: No need to undef this

PR #1485, opened 2026-08-11
Cache encoding lookups instead of redoing them per row

jeremy:
Materializing a string-heavy result runs three encoding lookups per row (two of them per string value) that always produce the same answer:

* `rb_to_encoding(wrapper->encoding)` -- unwraps the connection's Encoding object. Fixed at connect time (`Client#initialize` always runs `charset_name=`), so it can never change over a Result's life.
* `rb_default_internal_encoding()` -- re-read on every row.
* `rb_enc_find_index(enc_name)` -- a name-based hash lookup mapping the field's charsetnr to a Ruby encoding index, re-done for every string value of every row, though both sides of the mapping are fixed for the process.

Cache each at the widest scope that preserves its semantics:

* **Connection encoding, per Result.** Unwrapped once in `rb_mysql_result_to_obj` and stored on the wrapper as `rb_encoding *`. `rb_encoding` structs are immortal and unmoved by compaction, so the pointer cannot go stale. `to_obj` must not raise between taking ownership of the `MYSQL_RES` and returning (the streaming lifecycle depends on it); `rb_to_encoding` on a genuine Encoding instance is a plain data unwrap with no coercion, allocation, or exception path, and both call sites (client.c, statement.c) pass the client's Encoding instance.
* **`Encoding.default_internal`, per `#each` call.** Captured once at `rb_mysql_result_each` entry into `result_each_args` instead of once per row. `#fields`, `#field_types`, and field-name interning keep their per-call read.
* **charsetnr -> encoding index, per process.** A static array mirroring `mysql2_mysql_enc_to_rb` (323 entries), indexed by `charsetnr - 1` and bounds-checked to `1 <= charsetnr <= MYSQL2_CHARSETNR_SIZE`; out-of-range charsetnrs fall back to the connection encoding, uncached, exactly as before. Slot values are stored offset by +1 so 0 stays the "not yet cached" sentinel without conflating Ruby encoding index 0 (ASCII-8BIT); -1 caches "no table mapping". Only connection-independent facts are stored: the fallback for unmapped charsetnrs is resolved per Result at use time, so clients with different connection encodings share the cache soundly. Writers all hold the GVL (nothing in `mysql2_set_field_string_encoding` releases it); even without that, racing writers could only store the same deterministic int.

One disclosed behavior change: changing `Encoding.default_internal` from inside an iteration block no longer affects later rows of that same `#each` call -- the next `#each` call observes the new value. A spec pins the new behavior deliberately (it fails against the per-row capture). Nothing else observable moves: rows, fields, field_types, and every encoding/transcode result are byte-identical.

One unreachable corner hardened rather than preserved: if `rb_enc_find_index` ever returned a negative index, master passed it straight to `rb_enc_set_index`; now the value falls back to the connection encoding (and is not cached). Unreachable in practice -- every name in the table resolves under CRuby.

On the ASCII-8BIT (encoding index 0) cache slot: its only charsetnr mapping is 63 (binary), and every charsetnr-63 field MySQL 9.6 produces carries BINARY_FLAG -- probed with `mysql --column-type-info` across `BINARY 'a'`, `CAST(... AS BINARY)`, `UNHEX`, `RANDOM_BYTES`, `COMPRESS`, `WEIGHT_STRING`, `UUID_TO_BIN`, `LOAD_FILE`, `ST_GeomFromText`, `b'101'`, and `_binary'a'` -- so the earlier `BINARY_FLAG && charsetnr == 63` branch always intercepts before the cache is consulted. The index-0 path is therefore not constructible against this server and is untested at the integration level; the +1 offset keeps it correct by design should a server ever send charsetnr 63 without BINARY_FLAG.

Specs (`SET character_set_results = NULL` so fields arrive tagged with their own charsetnr rather than converted to the connection charset):

* A 13-field matrix across utf8mb4, latin1, latin2, greek, koi8r, ascii, sjis, cp932, big5, gb2312, binary, a collation id above 255 (utf8mb4_ja_0900_as_cs, id 303 -- exercising the cache beyond one byte), and dec8 (no table mapping -> connection-encoding fallback), asserting bytes and encoding per field, queried twice per example to hit warm cache slots. Run through the text protocol, the prepared-statement path, and streaming. The 0900 column runs on MySQL 8.0+ only: MariaDB has no in-table collation id above 255 (its uca1400 family, ids 2304+, lies beyond the mapping table and takes the connection-encoding fallback), so that pin is skipped there and the MySQL CI legs carry it.
* The same matrix under `Encoding.default_internal = UTF-8` (binary stays binary, everything else transcodes).
* Table columns of differing character sets (utf8mb4/latin1/BLOB) with and without default_internal.
* The deliberate mid-iteration default_internal pin described above.

Verified the specs bite: an off-by-one cache read (slot `charsetnr` instead of `charsetnr - 1`) fails three of them (text, prepared, columns); reverting the per-call default_internal capture to per-row fails the pin.

Benchmark: string-heavy shape (50,000 rows x 6 VARCHAR(32) utf8mb4 columns, multibyte content), `each {}` with `cache_rows: false`, three runs per build alternating base/branch, min of 5 timings per sample, 15 samples per run, per-run median (range) in ms per pass:

| arm          | run 1               | run 2               | run 3               |
|--------------|---------------------|---------------------|---------------------|
| base 82f9e18 | 57.09 (55.48-58.08) | 56.45 (55.33-60.89) | 56.05 (54.96-56.79) |
| this patch   | 48.82 (48.13-51.72) | 48.70 (48.24-54.24) | 48.55 (47.30-52.92) |

Median of medians: 56.45 ms -> 48.70 ms per pass, -13.7% on this workload (190.3 -> 162.3 ns per value), with no overlap between any base and branch range. Ruby 4.0.6, MySQL 9.6 (Homebrew, libmysqlclient.24), macOS arm64.

Independently reproduced on a second platform -- AMD Ryzen 9 7950X, Arch Linux x86_64, Ruby 4.0.6, MySQL 9.6.0 (Docker, TCP loopback), client libmariadb 12.3.2, machine load < 1.0 for every run; same protocol:

| arm          | run 1 | run 2 | run 3 |
|--------------|-------|-------|-------|
| base 82f9e18 | 35.66 | 35.60 | 35.41 |
| this patch   | 31.65 | 31.82 | 31.90 |

Median of medians: 35.60 ms -> 31.82 ms per pass, -10.6% (118.7 -> 106.1 ns per value), non-overlapping ranges in all three pairs.

Implemented and measured at 82f9e18, then rebased onto 96252dcc575e8626026c1e823e06a3d25739d915 (CI-only change -- apt-key to Signed-By swap in MySQL/MariaDB setup scripts, no library code -- so the measurements remain valid).

Full suite on the rebased head: 412 examples, 0 failures, 10 pending (SSL-gated), RuboCop clean, exit 0 verified unpiped.

### DIFF
--- a/ext/mysql2/result.c
+++ b/ext/mysql2/result.c
@@ -7,6 +7,19 @@
 
 static rb_encoding *binaryEncoding;
 
+/* Per-process cache of MySQL charsetnr -> Ruby encoding index, indexed by
+ * charsetnr - 1 to mirror mysql2_mysql_enc_to_rb. Both sides of the mapping
+ * are fixed for the life of the process -- the table is compiled in, and a
+ * Ruby encoding's index never changes once assigned -- so entries are computed
+ * once and never invalidated. Slot values are offset by +1 (0 is the "not yet
+ * cached" sentinel) so that encoding index 0, ASCII-8BIT, remains cacheable;
+ * -1 caches "this charsetnr has no table mapping", whose fallback (the
+ * connection's encoding) is applied per result at use time, never cached.
+ *
+ * Concurrent writers can only race to store the same deterministic value in
+ * an int slot, which is benign everywhere Ruby's GVL-based threading runs. */
+static int mysql2_enc_index_cache[MYSQL2_CHARSETNR_SIZE];
+
 /* on 64bit platforms we can handle dates way outside 2038-01-19T03:14:07
  *
  * (9999*31557600) + (12*2592000) + (31*86400) + (11*3600) + (59*60) + 59
@@ -56,6 +69,10 @@ typedef struct {
   ID db_timezone;
   ID app_timezone;
   int block_given; /* boolean */
+  /* Encoding.default_internal, captured once per #each call rather than once
+   * per row. Changing it mid-iteration therefore no longer affects later rows
+   * of that same call; the next #each call observes the new value. */
+  rb_encoding *default_internal_enc;
 } result_each_args;
 
 extern VALUE mMysql2, cMysql2Client, cMysql2Error;
@@ -312,7 +329,7 @@ static VALUE rb_mysql_result_fetch_field(VALUE self, unsigned int idx, int symbo
   if (rb_field == Qnil) {
     MYSQL_FIELD *field = NULL;
     rb_encoding *default_internal_enc = rb_default_internal_encoding();
-    rb_encoding *conn_enc = rb_to_encoding(wrapper->encoding);
+    rb_encoding *conn_enc = wrapper->conn_enc;
 
     field = mysql_fetch_field_direct(wrapper->result, idx);
     if (symbolize_keys) {
@@ -399,7 +416,7 @@ static VALUE rb_mysql_result_fetch_field_type(VALUE self, unsigned int idx) {
   if (rb_field_type == Qnil) {
     MYSQL_FIELD *field = NULL;
     rb_encoding *default_internal_enc = rb_default_internal_encoding();
-    rb_encoding *conn_enc = rb_to_encoding(wrapper->encoding);
+    rb_encoding *conn_enc = wrapper->conn_enc;
     int precision;
 
     field = mysql_fetch_field_direct(wrapper->result, idx);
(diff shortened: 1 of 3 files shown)


sodabrew:
Ready for rebase

PR #1486, opened 2026-08-11
Add a :force_encoding query option that retags string results

jeremy:
Callers who know better than the column charset need a way to say so: legacy schemas that store UTF-8 bytes in latin1 columns hand back strings tagged ISO-8859-1 that break the moment they meet the rest of the app, and code that hashes or byte-compares values wants raw binary without a `dup.force_encoding` per value. `Client#query` and `Statement#execute` now accept `force_encoding:` with an Encoding object or an encoding name (anything Encoding.find accepts):

    client.query("SELECT * FROM legacy", force_encoding: Encoding::UTF_8)
    stmt.execute(force_encoding: "binary")

Semantics, precisely:

* **Retag, not transcode.** Values keep their bytes; only the encoding tag changes. Force means force: the forced encoding overrides the BLOB/BINARY branch (binary columns get retagged too, instead of tagged ASCII-8BIT) and skips the Encoding.default_internal conversion entirely.
* **String values only.** With cast: true (the default), values cast to Integer/Date/Time and friends are untouched. With cast: false, every non-NULL value arrives as a string and all of them are retagged. Field names and field_types are unaffected.
* **Resolved at query/execute time, fixed for the life of the Result.** The value is canonicalized to an Encoding object via rb_to_encoding at the two command entry points (rb_mysql_query in client.c, rb_mysql_stmt_execute in statement.c) before anything is written to the wire. Invalid values raise ArgumentError/TypeError with the connection untouched and immediately reusable. Nothing may raise once the command is sent -- rb_mysql_result_to_obj in particular must not raise between taking ownership of the MYSQL_RES and returning (the streaming lifecycle, #1463) -- and by then only a plain Encoding unwrap remains. Storing the resolved Encoding back into the per-query options snapshot also makes the Result immune to the caller mutating their options hash, or the encoding-name string itself, after the call.
* **Rejected on Result#each.** `result.each(force_encoding: ...)` raises Mysql2::Error. Non-streaming Statement#execute materializes every row internally by calling #each itself, so a per-#each value could never be honored consistently; rejecting beats silently ignoring. Key presence is the error, even with a nil value. The option is query/execute-only, and the README section documents all of the above.

On the statement path, the options dup/merge moved ahead of bind-buffer setup so canonicalization raises before any bind buffers are xmalloc'd -- an error there now leaks nothing and leaves the connection untouched.

Interaction with the encoding caches underneath (#1485): the forced path returns before the per-process charsetnr cache is consulted, so it never reads or writes it; the non-forced path is unchanged.

This PR is deliberately stacked on #1485: both change mysql2_set_field_string_encoding, and sodabrew's triage of #1422 suggested the two could travel together -- split into two reviewable pieces, second riding on the first. The first commit here is #1485's; review the second commit.

Specs (result_spec, statement_spec):

* Retags string values, bytes unchanged: text protocol, prepared statements, and text-protocol streaming.
* BLOB/binary values retagged under force_encoding: "utf-8" instead of left ASCII-8BIT.
* Wins over Encoding.default_internal: tag forced, bytes untranscoded.
* Accepts an Encoding object and a String name; non-encoding values (integers) raise TypeError.
* Unknown encoding name raises ArgumentError before the command is sent, pinned with an INSERT probe on both paths: the row never lands, and the connection (and the prepared statement handle) stay usable.
* Snapshot immunity: mutating the caller's options hash or the encoding-name string after query returns does not affect the lazily materialized rows.
* `each(force_encoding: ...)` raises, message pinned; nil value included; re-iteration with other per-each options still works, and the merged defaults carrying the query-time value do not re-trip the check.
* Non-string casts (INTEGER, DATE) untouched; cast: false retags everything; plain and symbolized field names unaffected.
* Prepared statements: NULL values and bind parameters combined with the keyword option.
* Streaming prepared statements cannot return string values at all, with or without this option: statement streaming skips mysql_stmt_store_result, so string fetches die with MYSQL_DATA_TRUNCATED ("IMPLBUG...") -- pre-existing upstream behavior, which is why the existing streaming-statement specs only use integers. A spec pins that the option at least threads through that path unharmed.

Verified the specs bite: against the base build (specs applied, extension changes reverted), 15 of 18 fail. The three survivors are the deliberate invariance pins (non-string casts, field names) plus one strengthened after proving vacuous -- String#eql? ignores encoding for ASCII-only strings, so the cast: false spec now asserts the encoding directly.

This is a feature PR, not a performance one: no benchmarks, no performance claims. The forced path costs one pointer test per string value.

Environment: base 82f9e18 + #1485, Ruby 4.0.6, MySQL 9.6 (Homebrew, libmysqlclient.24), macOS arm64.

Full suite: 430 examples, 0 failures, 10 pending (SSL-gated). RuboCop clean.

### DIFF
--- a/README.md
+++ b/README.md
@@ -576,6 +576,22 @@ do_mysql               0.520000   0.010000   0.530000 (  0.574619)
 
 Although Mysql2 performs reasonably well at retrieving uncasted data, it (currently) is not as fast as the Mysql gem.  In spite of this small disadvantage, Mysql2 still sports a friendlier interface and doesn't block the entire ruby process when querying.
 
+### Forcing a string encoding
+
+Pass `:force_encoding` to `Client#query` or `Statement#execute` to retag string results with an encoding of your choosing:
+
+``` ruby
+result = client.query("SELECT * FROM legacy_table", force_encoding: Encoding::UTF_8)
+```
+
+It accepts an `Encoding` object or an encoding name (anything `Encoding.find` accepts). Invalid values raise before anything is sent to the server.
+
+Retag means exactly that: values keep their bytes and only the encoding tag changes — nothing is transcoded. Force means force, so the forced encoding wins over everything else: BLOB/`BINARY` columns are retagged too (instead of being tagged ASCII-8BIT), and the usual `Encoding.default_internal` conversion is skipped. This is useful for reading legacy data stored in a mislabeled character set (say, UTF-8 bytes living in latin1 columns), or for forcing `binary` when you want raw bytes for hashing or byte-wise comparison.
+
+Only values that arrive as strings are affected. With the default `cast: true`, columns cast to Integer/Date/Time and friends are untouched; with `cast: false`, every non-NULL value is a string and all of them are retagged. Field names are never affected.
+
+`:force_encoding` is fixed when the query or execute is issued — `Mysql2::Result#each` raises if you pass it there, because non-streaming `Statement#execute` materializes rows internally with `#each`, so a per-`each` value could never be honored consistently.
+
 ### Partial casting
 
 Between `:cast => true` and `:cast => false` sits `:cast => :fast`: cheap conversions still happen in C, while the expensive ones are skipped and their values returned as encoding-tagged Strings of the raw wire bytes.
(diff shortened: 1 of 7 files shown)


sodabrew:
Rebase needed after #1485 merge

jeremy:
Rebased onto master (466a9ae) — clean-applying again, suite green. The conflict was #1489's new call site needing the widened 5-arg `mysql2_set_field_string_encoding`; `forced_enc` is threaded through it. I'll do the same one-liner rebase as soon as #1490 lands.

jeremy:
Rebased onto ceb90ed (#1490's merge), clean-applying. Full suite green: 500 examples, 0 failures.

PR #1489, opened 2026-08-11
cast: false rows: skip the per-cell type dispatch entirely

jeremy:
Callers that pass cast: false -- raw pipelines, ETL jobs, proxies that re-serialize rows without ever looking inside them -- are asking for the wire bytes untouched. Yet on the text protocol every cell still walks into the full per-type dispatch machinery just to reach the raw-string arm buried inside it, paying the branch tour for a conversion that never happens.

Hoist that arm into its own tight loop at the top of rb_mysql_result_fetch_row: when args->cast is false, every non-NULL cell becomes a String via the same length-based rb_str_new (embedded NULs intact) and the same mysql2_set_field_string_encoding call, and everything else about the cell is kept identical -- MYSQL_TYPE_NULL fields and NULL cells are nil, hash keys are fetched before their value is built (so a raise mid-row leaves the same field-name cache state as before), and the shared setup above the loop is untouched, so hash rows are still pre-sized with rb_hash_new_capa and array-mode field names are still batch-materialized before any value conversion, preserving the ordering guarantee the previous parcel established. The loop runs entirely with the GVL held; the only nogvl region in the function remains the streaming row fetch above it.

The prepared-statement path (rb_mysql_result_fetch_row_stmt) is deliberately untouched: on master, cast: false there warns ":cast is forced for prepared statements" and fully casts anyway, and a spec now pins exactly that.

Stacking disclosure: the first two commits on this branch are #1485 (encoding lookup caches) and #1487 (array-mode field-name batch materialization); review the third commit. The fast path leans on both -- the cached conn_enc/default_internal_enc and the materialize-before-conversion ordering.

Parity is proven in specs, not prose (spec/mysql2/result_spec.rb, "cast: false raw string rows"): a 15-column table covering INT/BIGINT UNSIGNED boundaries, DECIMAL, FLOAT, DATE, DATETIME(6) with fractional seconds, TIME at its -838:59:59 extreme, YEAR, BIT(64) with seven leading NUL bytes, TINYINT(1), multibyte utf8mb4 VARCHAR, a latin1 column, BLOB and VARBINARY with embedded NULs, a row of NULLs in every column, and a row of zero-length strings as distinct from NULL. Expected values are literals -- bytes and encoding both -- never derived from another code path at runtime, and the matrix runs cast: false x {hash, array} x {symbolize_keys on/off} x {buffered, streaming}, plus cache_rows: false re-iteration, charsetnr tagging with character_set_results disabled (latin1 bytes arrive tagged ISO-8859-1), Encoding.default_internal semantics (BINARY_FLAG fields never transcode; charsetnr-63-without-BINARY_FLAG numerics do), and #fields after an abandoned cast: false stream is force-freed. The same 14 examples pass unchanged against the pre-parcel build, pinning that the fast path changed nothing observable.

Verified the specs bite: a strlen-based copy in the fast path fails 11 of the 14 (BIT(64) truncates to ""), and dropping the encoding tagging fails 11 of the 14 (utf8mb4 values arrive ASCII-8BIT).

Benchmark: 50,000 rows x 10 mixed columns -- ints, strings, datetimes, decimal, double, date -- cast: false in hash and array modes with cache_rows: false, min-of-5 per sample, 9 samples per run, three serial alternating control/branch pairs on a quiet machine (AMD Ryzen 9 7950X, Arch Linux x86_64, Ruby 4.0.6, MySQL 9.6.0 in Docker over TCP loopback, client libmariadb 12.3.2, load < 1.4 before every run). Control is this branch's parent commit (the #1487 tip), so the numbers isolate this change alone. Medians of medians, ms per pass:

| arm               | control | this patch | delta |
|-------------------|---------|------------|-------|
| cast:false, hash  |   56.99 |      55.52 | -2.6% |
| cast:false, array |   46.57 |      46.22 | -0.8% |
| cast:true, hash   |  101.35 |     100.82 | noise |
| cast:true, array  |   88.42 |      88.89 | noise |

Sized honestly: the hash-mode win is small but reproducible (branch medians below control minima in effectively every pair); the array-mode delta is directionally consistent but inside the noise floor -- the field-name materialization underneath this stack already removed array mode's larger per-cell cost, leaving little for this hoist to reclaim. The cast:true control arms are flat, confirming the default path is untouched. On its own this reorganization would be borderline; its real role is as the substrate for the `cast: :fast` partial-casting mode stacked directly on top of it (see that PR for the payoff: -41% to -64% vs full casting), plus the byte-exact cast:false parity suite this commit pins for both of them.

Stack base 96252dc + #1485 + #1487, Ruby 4.0.6, MySQL 9.6 (Homebrew, libmysqlclient.24), macOS arm64. Full suite on the stacked baseline before this change: 429 examples, 0 failures, 10 pending (SSL-gated). With this change: 443 examples, 0 failures, 10 pending, RuboCop clean, exit 0 verified unpiped.

### DIFF
--- a/ext/mysql2/result.c
+++ b/ext/mysql2/result.c
@@ -1199,6 +1199,38 @@ static VALUE rb_mysql_result_fetch_row(VALUE self, MYSQL_FIELD * fields, const r
   }
   fieldLengths = mysql_fetch_lengths(wrapper->result);
 
+  /* cast: false wants every non-NULL value as a raw String with the right
+   * encoding, so the per-cell type dispatch below is pure overhead for it.
+   * This loop is the cast: false arm of the general loop with the dispatch
+   * hoisted out; everything else is kept cell-for-cell identical -- the same
+   * length-based rb_str_new (embedded NULs intact), the same
+   * mysql2_set_field_string_encoding, the same MYSQL_TYPE_NULL and NULL-cell
+   * handling, and the same key-before-value evaluation order -- so rows are
+   * byte-identical with just fewer branches per cell. Runs entirely with the
+   * GVL held (the only nogvl region in this function is the streaming row
+   * fetch above). */
+  if (!args->cast) {
+    for (i = 0; i < wrapper->numberOfFields; i++) {
+      /* Hash keys only; array-mode names were batch-materialized above. */
+      VALUE field = args->asArray ? Qnil : rb_mysql_result_fetch_field(self, i, args->symbolizeKeys);
+      VALUE val;
+
+      if (row[i] && fields[i].type != MYSQL_TYPE_NULL) {
+        val = rb_str_new(row[i], fieldLengths[i]);
+        val = mysql2_set_field_string_encoding(val, fields[i], default_internal_enc, conn_enc);
+      } else {
+        val = Qnil;
+      }
+
+      if (args->asArray) {
+        rb_ary_push(rowVal, val);
+      } else {
+        rb_hash_aset(rowVal, field, val);
+      }
+    }
+    return rowVal;
+  }
+
   for (i = 0; i < wrapper->numberOfFields; i++) {
     /* Hash keys only; array-mode names were batch-materialized above. */
     VALUE field = args->asArray ? Qnil : rb_mysql_result_fetch_field(self, i, args->symbolizeKeys);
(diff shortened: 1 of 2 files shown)


sodabrew:
Ready for rebase

sodabrew:
Rebase needed after #1485 merge

jeremy:
Rebased onto master (7bb1d4d) — now a single commit, clean-applying, full suite green. #1490 rebase onto this follows in a moment.

PR #1490, opened 2026-08-11
cast: :fast rows: cast cheap types, defer expensive ones as strings

jeremy:
On a casting-dominated shape (50,000 rows x 10 columns: 4 temporal, 2 decimal, 1 float, 3 int), cast: true costs 152ms where cast: false costs 73ms -- a 51.8% ceiling for partial casting, measured on a quiet Linux x86_64 box (Ryzen 9 7950X, Ruby 4.0.6, MySQL 9.6.0, libmariadb 12.3.2, base 82f9e18; the ceiling is a property of casting cost, not of this stack). Nearly all of that cost is BigDecimal, Date, and Time construction for cells many callers never parse eagerly: raw-mysql2 pipelines, ETL jobs, and applications that pass DECIMAL and temporal values through or parse them lazily. cast: false recovers the whole ceiling but throws away the cheap casts too, forcing those callers to re-parse every integer themselves.

cast: :fast sits between the two, as a third arm of the existing cast dispatch in rb_mysql_result_fetch_row (text protocol only):

- Cast exactly as cast: true casts them: NULL (nil); the integer family
  TINY/SHORT/LONG/INT24/LONGLONG/YEAR via the same rb_cstr2inum arm
  (with an explicit comment on the TINY fallthrough: :cast_booleans
  still wins for TINYINT(1)); FLOAT/DOUBLE via the same
  locale-independent Kernel#Float mechanism (never strtod, whose '.'
  handling follows LC_NUMERIC); BIT, including its :cast_booleans BIT(1)
  arm. - Deferred as Strings of the raw wire bytes, tagged by
  mysql2_set_field_string_encoding exactly as the cast: false loop tags
  them: DECIMAL/NEWDECIMAL (scale-0 included -- the mode is a per-type
  contract, not a per-value one), DATE/NEWDATE, DATETIME, TIMESTAMP,
  TIME. - Everything else (text, blobs, enum, set, json, geometry, ...) is
  already a String under cast: true and is byte- and encoding-identical
  here.

The mode is selected only by the exact symbol :fast, as a query option or per-#each, either place cast: false is accepted today. Any other truthy :cast value keeps meaning full casting, as every truthy value always has -- an unrecognized value must not silently opt into partial casting; a spec pins cast: :bogus to full casting. The cast: false fast path and the shared row setup above it (rb_hash_new_capa pre-sizing, array-mode field names batch-materialized before any value conversion) are untouched; :fast rides the existing casting loop, so cheap types cannot drift from cast: true by construction. Prepared statements keep their cast: false precedent: the binary-protocol fetch always fully casts, and the existing generic ":cast is forced for prepared statements" warning now also fires for :fast (condition widened from !cast to cast != MYSQL2_CAST_ALL; the text already reads as "any non-true :cast is overridden here").

Stacking disclosure: this branch is a 4-commit stack -- #1485 (encoding lookup caches), #1487 (array-mode field-name batch materialization), the cast: false fast-path parcel, then this; review the fourth commit. The deferred-string arm leans on the first and third: the cached conn_enc/default_internal_enc and the hoisted raw-string loop whose tagging it mirrors.

Specs (spec/mysql2/result_spec.rb, "cast: :fast partially-cast rows"): a 20-column table -- BIGINT at LLONG_MIN and BIGINT UNSIGNED at ULLONG_MAX, INT UNSIGNED ZEROFILL, DECIMAL(10,3) and scale-0 DECIMAL(10,0), FLOAT/DOUBLE asserted for exact value, DATE, DATETIME(6) and TIMESTAMP(6) with fractional seconds, TIME at -838:59:59, YEAR, BIT(64) with embedded NULs, BIT(1), TINYINT(1), multibyte utf8mb4 and latin1 VARCHARs, BLOB and VARBINARY with embedded NULs, a row of NULLs, and a row of zero-length strings as distinct from NULL. Expected values are literals -- bytes and encoding both -- never derived from another code path at runtime. The matrix runs cast: :fast x {hash, array} x {buffered, streaming}, plus cache_rows: false re-iteration, per-#each cast: :fast, :cast_booleans interplay, comma-decimal locale (de_DE.UTF-8) float parsing, charsetnr tagging with character_set_results disabled, Encoding.default_internal semantics (temporal BINARY_FLAG fields never transcode; DECIMAL's charsetnr-63-without-BINARY_FLAG does; BIT stays untagged as under cast: true), #fields after an abandoned :fast stream is force-freed, cast: :bogus, and prepared + :fast warning with fully-cast values.

Verified the specs bite: dropping the encoding tagging on the deferred DECIMAL arm fails the default_internal example; parsing :fast floats with strtod fails the comma-locale example (2.7 arrives as 2.0). Both restored and re-verified green.

Benchmark: the prescreen shape above plus a realistic-dilution shape -- 50,000 rows x 10 columns: 2 temporal, 1 decimal, 4 string, 3 int -- each under cast: true, cast: :fast, and cast: false, hash mode, cache_rows: false, min-of-5 per sample, 9 samples per run, three serial alternating control/branch pairs on a quiet machine (AMD Ryzen 9 7950X, Arch Linux x86_64, Ruby 4.0.6, MySQL 9.6.0 in Docker over TCP loopback, client libmariadb 12.3.2, load < 1.0 before every run). Control is this branch's parent commit (the #1489 tip), isolating this change alone; on the control build `cast: :fast` falls through to full casting by the exact-symbol rule, and its :fast arm indeed matched its cast:true arm within 1% -- the fall-through contract, measured. Medians of medians, ms per pass:

| arm         | prescreen ctrl | prescreen this | diluted ctrl | diluted this |
|-------------|----------------|----------------|--------------|--------------|
| cast: true  |         147.82 |         147.90 |        93.66 |        93.05 |
| cast: :fast |  148.61 (=true)|          52.77 |94.38 (=true) |        54.69 |
| cast: false |          52.56 |          52.97 |        54.57 |        54.67 |

Headline: cast: :fast vs cast: true is -64.3% on the temporal/decimal- heavy shape and -41.2% on the diluted one, with distributions separated by ~95 ms and ~38 ms against sub-2 ms ranges. Against cast: false -- the ceiling -- :fast is a statistical tie (-0.4% / +0.04%, fully overlapping ranges): partial casting captures effectively all of the ceiling while still returning real Integers and Floats. Both untouched paths are flat: cast:true within +/-0.7% with overlapping ranges (the new per-expensive-cell guard costs nothing measurable) and cast:false within +0.8% inside its own run-to-run spread.

Two honest bounds on those numbers. The win is definitionally shape-dependent: an all-int/string result set has nothing to defer. And this measures driver-side materialization only -- a caller that immediately converts every deferred string to Time/BigDecimal in Ruby pays the cost back with interest; the option exists for callers that filter, forward, or parse selectively.

Stack base 96252dc + #1485 + #1487 + #1489, Ruby 4.0.6, MySQL 9.6 (Homebrew, libmysqlclient.24), macOS arm64. Full suite on the stacked baseline before this change: 443 examples, 0 failures, 10 pending (SSL-gated), exit 0. With this change: 456 examples, 0 failures, 10 pending, RuboCop clean, exit 0 verified unpiped.

### DIFF
--- a/README.md
+++ b/README.md
@@ -576,6 +576,22 @@ do_mysql               0.520000   0.010000   0.530000 (  0.574619)
 
 Although Mysql2 performs reasonably well at retrieving uncasted data, it (currently) is not as fast as the Mysql gem.  In spite of this small disadvantage, Mysql2 still sports a friendlier interface and doesn't block the entire ruby process when querying.
 
+### Partial casting
+
+Between `:cast => true` and `:cast => false` sits `:cast => :fast`: cheap conversions still happen in C, while the expensive ones are skipped and their values returned as encoding-tagged Strings of the raw wire bytes.
+
+* Cast, exactly as `:cast => true` does: `NULL` (`nil`), the integer types (TINYINT through BIGINT, YEAR), FLOAT/DOUBLE, BIT, and TINYINT(1)/BIT(1) booleans when `:cast_booleans` is enabled.
+* Returned as Strings, exactly as `:cast => false` returns them: DECIMAL, DATE, DATETIME, TIMESTAMP, and TIME -- the types whose casting dominates the cost of materializing a row (BigDecimal, Date, and Time construction).
+* Everything else (CHAR/VARCHAR/TEXT, blobs, ENUM, SET, JSON, ...) is already a String and is identical to `:cast => true`.
+
+``` ruby
+result = client.query("SELECT * FROM table", :cast => :fast)
+```
+
+This is for callers that consume mysql2 results directly -- raw-mysql2 pipelines, ETL jobs, and applications that do their own value parsing -- where DECIMAL and temporal columns are often passed through or parsed lazily, and paying BigDecimal/Time construction for every cell up front is waste. Your code must be prepared to receive Strings for those columns. Note that Active Record does not use this option.
+
+Any `:cast` value other than the exact symbol `:fast` (or `false`/`nil`) keeps meaning full casting. Prepared statement results are always fully cast, as with `:cast => false`.
+
 ### Async
 
 NOTE: Not supported on Windows.
--- a/ext/mysql2/result.c
+++ b/ext/mysql2/result.c
@@ -58,12 +58,23 @@ static int mysql2_enc_index_cache[MYSQL2_CHARSETNR_SIZE];
   mysql2_result_wrapper *wrapper; \
   TypedData_Get_Struct(self, mysql2_result_wrapper, &rb_mysql_result_type, wrapper);
 
+/* How much per-cell casting #each performs, parsed from the :cast option:
+ * false/nil is MYSQL2_CAST_NONE, :fast is MYSQL2_CAST_FAST, and any other
+ * truthy value -- not just true -- is MYSQL2_CAST_ALL, so an unrecognized
+ * value keeps meaning full casting (as every truthy value always has)
+ * instead of silently opting into partial casting. */
+typedef enum {
+  MYSQL2_CAST_NONE = 0, /* cast: false -- every non-NULL value is a raw String */
+  MYSQL2_CAST_ALL  = 1, /* cast: true -- full type casting */
+  MYSQL2_CAST_FAST = 2  /* cast: :fast -- cast cheap types; defer expensive ones as Strings */
+} mysql2_cast_mode;
+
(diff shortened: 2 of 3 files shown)


sodabrew:
Will land this one next, then #1486 last

PR #1491, opened 2026-08-11
CI: Add Ubuntu 26.04 and the MySQL innovation release

fkmy:
This adds two Ubuntu 26.04 jobs to the matrix.

The part I would like a second opinion on is the MySQL innovation release. Innovation releases lead the next LTS, so testing against one surfaces libmysqlclient changes a cycle before they reach it. The cost is that the job moves on its own -- there is no version-specific APT component for innovation releases, so it follows `mysql-innovation` and picks up whatever is newest, 26.7 today. If a future release breaks something, the matrix goes red on pull requests that have nothing to do with it.

Refs:
- https://dev.mysql.com/doc/relnotes/mysql/26.7/en/news-26-7-0.html
- https://dev.mysql.com/blog-archive/introducing-mysql-innovation-and-long-term-support-lts-versions/

### DIFF
--- a/.github/workflows/build-ubuntu.yml
+++ b/.github/workflows/build-ubuntu.yml
@@ -4,16 +4,19 @@ on:
 jobs:
   build:
     name: >-
-      ${{ matrix.os }} ruby ${{ matrix.ruby }} ${{ matrix.db }}
+      ${{ matrix.os }} ruby ${{ matrix.ruby }} ${{ matrix.db }}${{ matrix.allow_failure && ' (allow failure)' || '' }}
     runs-on: ${{ matrix.os }}
     timeout-minutes: 30
+    # allow_failure entries track something that moves outside this repository,
+    # so their failures are informational and must not block unrelated PRs.
+    continue-on-error: ${{ matrix.allow_failure || false }}
     strategy:
       matrix:
         include:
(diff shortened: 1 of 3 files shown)


.github/workflows/build-ubuntu.yml
```diff
@@ -10,8 +10,12 @@ jobs:
     strategy:
       matrix:
         include:
+          # Ruby 4.x on Ubuntu 26.04 LTS
+          # db: mysql-innovation tracks the newest MySQL innovation release, not an LTS.
+          - {os: ubuntu-26.04, ruby: '4.0', db: mysql-innovation}
+          - {os: ubuntu-26.04, ruby: 'head', db: mysql97}
```
sodabrew: Could make these allow-fail, so the failure is informational and not blocking to other PRs

fkmy: Done -- both 26.04 entries take `allow_failure: true`, with `continue-on-error: ${{ matrix.allow_failure || false }}` on the job.

A `continue-on-error` job that fails still reports green, so a real breakage in the innovation release would be easy to miss. As a small mitigation I've appended ` (allow failure)` to the job name, which at least flags it in the checks list.

Another option would be to move these into a workflow with its own trigger (`schedule`, say), which would keep the failure red without touching pull request status.

PR #1492, opened 2026-08-11
Fixes #1033: next_result no longer holds the GVL or blocks Thread#raise

### LINKED ISSUE #1033: client.next_result holds the GVL during a network call
When Mysql2::Client::MULTI_STATEMENTS is enabled and you call client.next_result, the GVL is held while we make a network call out to the MySQL server to obtain the next result set.  

rb_mysql_client_next_result calls mysql_next_result, which makes a call out to the MySQL server.

In the docs for mysql_next_result (https://dev.mysql.com/doc/refman/8.0/en/mysql-next-result.html) it's not totally clear that it reads from the network, though there is the hint that it can return CR_SERVER_GONE_ERROR or CR_SERVER_LOST.

In practice on Mac, it looks like it definitely does;  here's the stack which shows mysql_next_result making a network read call:
```
rb_mysql_client_next_result  (in mysql2.bundle) + 32  [0x10a4c18a0]  client.c:1142
 mysql_next_result  (in libmysqlclient.21.dylib) + 225  [0x10a4db959]
  cli_read_query_result(MYSQL*)  (in libmysqlclient.21.dylib) + 26  [0x10a4e50f8]
   cli_safe_read_with_ok  (in libmysqlclient.21.dylib) + 214  [0x10a4dd9ea]
    my_net_read(NET*)  (in libmysqlclient.21.dylib) + 94  [0x10a4e86d4]
     net_read_packet(NET*, unsigned long*)  (in libmysqlclient.21.dylib) + 47  [0x10a4e892b]
      net_read_raw_loop(NET*, unsigned long)  (in libmysqlclient.21.dylib) + 44  [0x10a4e8a61]
       vio_ssl_read(Vio*, unsigned char*, unsigned long)  (in libmysqlclient.21.dylib) + 47  [0x10a521082]
        ssl3_read_internal  (in libssl.1.0.0.dylib) + 96  [0x10aa157b8]
         ssl3_read_bytes  (in libssl.1.0.0.dylib) + 532  [0x10aa1849f]
          ssl3_read_n  (in libssl.1.0.0.dylib) + 587  [0x10aa1754b]
           BIO_read  (in libcrypto.1.0.0.dylib) + 100  [0x10aaf0fed]
            sock_read  (in libcrypto.1.0.0.dylib) + 50  [0x10aaf2a3f]
             read  (in libsystem_kernel.dylib) + 10  [0x7fff6566e142]
```             

In my case this caused an invisible deadlock (this call held the GVL and was waiting on a MySQL lock, and the Ruby thread with that lock couldn't run because it was in the same process and couldn't get the GVL).

sodabrew:
Fixes #1033. Also closes out #807 and #962, two earlier attempts at the same fix that never merged (see my comment on #1033 for the full history).

## Problem

`client.next_result` calls `mysql_next_result()` directly. When the next statement in a multi-statement batch (`SELECT 1; SELECT SLEEP(3)`) isn't ready yet, that's a real blocking network read -- unlike the initial query, which already releases the GVL and polls an interruptible fd wait (`do_query`/`wait_for_fd`), `next_result` did neither.

## Fix

- On MySQL 8.0.16+, poll `mysql_next_result_nonblocking()` -- a real libmysqlclient API added years after #807/#962 were written -- instead of calling `mysql_next_result()` directly. Each iteration either completes immediately or reports `NET_ASYNC_NOT_READY`, in which case the *wait*, not the poll, is what actually blocks, via the same `rb_wait_for_single_fd` machinery `do_query` already uses, so it's interruptible the same way. Mixing this with the ordinary blocking API elsewhere on the same connection is [explicitly documented as supported by MySQL](https://dev.mysql.com/doc/c-api/8.0/en/c-api-asynchronous-interface-usage.html), so nothing else in `client.c` needed to change -- and #962's `vio_priv.h` detection for already-buffered data isn't needed at all anymore, since the nonblocking function itself reports that.
- Falls back to releasing the GVL around the still-blocking call (`rb_thread_call_without_gvl`) on MariaDB and on MySQL older than 8.0.16, guarded by `have_func('mysql_next_result_nonblocking')`. This fixes the GVL-holding half of the bug unconditionally; the interruptibility half depends on whether the linked libmysqlclient retries internally on the signal `RUBY_UBF_IO` sends on `EINTR`, which this fallback can't control -- verified empirically on MySQL 8.0.34 (macOS) with the fast path force-disabled that both problems actually resolve here too, but that isn't guaranteed on every platform/library build.

## Testing

Two new specs in `spec/mysql2/client_spec.rb`, under "Multiple results sets":
- Measures GVL availability during the wait via a counter thread ticking on `sleep 0.01` -- fails on unfixed code (0 ticks during a 1s block, ~100 expected), passes now.
- Asserts `Timeout.timeout` around a slow `next_result` actually raises `Timeout::Error` -- fails on unfixed code (silently absorbed, returns after the full wait), passes now.

Full suite: 450 examples, 0 failures.

## Known gap

MariaDB Connector/C has its own non-blocking pair for this (`mysql_next_result_start`/`_cont`), not wired up here -- it uses a different `MYSQL_OPT_NONBLOCK`-gated pattern than MySQL's freely-mixable one, and I want to verify it's actually safe to mix with the rest of this codebase's blocking calls before extending the fix to that path rather than guess. MariaDB gets the GVL-release fallback in the meantime, same as older MySQL. If CI turns up a MariaDB-specific issue, will iterate from there.

### DIFF
--- a/ext/mysql2/client.c
+++ b/ext/mysql2/client.c
@@ -1616,6 +1616,142 @@ static VALUE rb_mysql_client_more_results(VALUE self)
     return Qtrue;
 }
 
+#ifndef _WIN32
+/* Waits for the connection's fd to become readable, honoring @read_timeout
+ * the same way do_query does for the initial query. rb_wait_for_single_fd
+ * itself releases the GVL and is interruptible (Thread#raise,
+ * Timeout.timeout), unlike a raw blocking read inside libmysqlclient. */
+static void wait_for_next_result_fd(VALUE self, mysql_client_wrapper *wrapper) {
+  struct timeval tv;
+  struct timeval *tvp;
+  long int sec;
+  int retval;
+  VALUE read_timeout;
+
+  read_timeout = rb_ivar_get(self, intern_read_timeout);
+
+  tvp = NULL;
+  if (!NIL_P(read_timeout)) {
+    Check_Type(read_timeout, T_FIXNUM);
+    tvp = &tv;
+    sec = FIX2INT(read_timeout);
+    if (sec >= 0) {
+      tvp->tv_sec = sec;
+    } else {
+      rb_raise(cMysql2Error, "read_timeout must be a positive integer, you passed %ld", sec);
+    }
+    tvp->tv_usec = 0;
+  }
+
+  retval = rb_wait_for_single_fd(wrapper->client->net.fd, RB_WAITFD_IN, tvp);
+
+  if (retval == 0) {
+    rb_raise(cMysql2TimeoutError, "Timeout waiting for a response from the last query. (waited %d seconds)", FIX2INT(read_timeout));
+  }
+  if (retval < 0) {
+    rb_sys_fail(0);
+  }
+}
+#endif
+
+#if defined(HAVE_MYSQL_NEXT_RESULT_NONBLOCKING) && !defined(_WIN32)
+/* Polls mysql_next_result_nonblocking() (added in MySQL 8.0.16) instead of
+ * calling the blocking mysql_next_result() directly. Unlike the blocking
+ * call, this never holds the GVL for a real network wait: each iteration
+ * either returns immediately (never blocks, by the function's own
+ * contract) or reports NET_ASYNC_NOT_READY, in which case
+ * wait_for_next_result_fd -- not this function -- is what actually waits,
+ * and it does so interruptibly.
+ *
+ * Mixing this with the ordinary blocking API on the same connection is
+ * explicitly documented as supported, so nothing else in this file needs
+ * to change:
+ * https://dev.mysql.com/doc/c-api/8.0/en/c-api-asynchronous-interface-usage.html
+ */
+static enum net_async_status next_result_nonblocking(VALUE self, mysql_client_wrapper *wrapper) {
+  enum net_async_status status;
+
+  for (;;) {
+    status = mysql_next_result_nonblocking(wrapper->client);
+    if (status != NET_ASYNC_NOT_READY) {
+      return status;
+    }
+    wait_for_next_result_fd(self, wrapper);
+  }
+}
+#else
+/* Fallback for MariaDB (mysql_next_result_nonblocking doesn't exist there
+ * under this name -- MariaDB Connector/C has its own mysql_next_result_start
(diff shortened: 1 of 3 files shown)


sodabrew:
## MariaDB Connector/C investigation: status update

Followed up on the "known gap" above by building against a real MariaDB Connector/C and testing directly, rather than reasoning from docs alone.

**Setup**: MariaDB Connector/C 3.4.9 (latest release), built from source on macOS ARM64, tested against a real MariaDB 11.8.8 server -- three standalone C test programs, no changes to `ext/mysql2/`, driving real `SELECT 1; SELECT SLEEP(2)` multi-statement batches over the network.

**Good news**: mixing ordinary blocking calls before *and* after `mysql_next_result_start()`/`_cont()` is safe. Tested three configurations (`MYSQL_OPT_NONBLOCK` set after connect, set before connect, and a fully async connect+query pipeline) -- no crashes, no errors, no corruption in any of them. A blocking `mysql_store_result()`, a fresh blocking `mysql_query()`, even `mysql_ping()`, all worked correctly on a connection that had just used the async pair. That resolves the "is this an invasive, connection-wide commitment" concern from my earlier comment -- it isn't. A small, surgical addition parallel to what's already in this PR for MySQL should be architecturally fine.

**Open problem**: in every configuration tested, `mysql_next_result_start()` itself blocked synchronously for the *entire* `SLEEP(2)` wait on its first call, reporting zero polling iterations -- it never returned `MYSQL_WAIT_READ` to poll on. It behaved like a plain blocking call wearing the async-named API, for exactly the use case this fix needs (waiting on a slow next statement). Whether that's a bug/limitation in Connector/C's implementation of this specific call, or something about how the test harness invoked it, isn't resolved yet -- instrumentation of the vendored source traced the read as far as `ma_pvio_read()` with the async flag correctly set, but not to where it should yield.

**Bottom line**: not ready to extend this fix to MariaDB via `mysql_next_result_start`/`_cont` yet -- the mixing question is answered favorably, but the actual non-blocking behavior for this call isn't confirmed working. MariaDB keeps using the GVL-release fallback in this PR for now (confirmed working, including in CI). Will pick this back up as a follow-up once the yield behavior is understood.

Issue #1415, opened 2025-10-11
Intermittent connection issues with Ruby 3.2 and 3.3.


MrJoy:
This is less a bug report and more a request for some guidance.

We're seeing what appears to be a pair of issues affecting both mysql2 and trilogy -- only under Ruby 3.2 and/or 3.3.

We recently updated an app from Ruby 3.1.6 to 3.3.8 / 3.3.9 (first one, then the other -- both are affected), and started seeing both connection timeouts (only from the mysql2 gem) and dropped connections (both mysql2 and trilogy).  They are infrequent, relative to the workload, but not _rare_: We're seeing about 5-10 occurrences of each per hour.  The connection timeouts are coming from a web service handling about 9k req/sec, and the dropped connections are coming from a Sidekiq service handling about 1 job/sec.

I downgraded to 3.1.6 and confirmed both issues went away.  I upgraded to 3.2.9 to bisect my way to an exact Ruby version that introduced the issue and... the dropped connection issue recurred (at about the same frequency as before) -- but not the connection timeout issue.

Our deploy process is such that I can be certain that no OS packages or configuration were changed as I flipped between Ruby versions, and no application code or configuration changes happened either.  (For the sake of thoroughness:  Our images are all currently Debian 13.  We build Ruby from source.  We're using mysql2 0.5.7.)

So, it sure looks like we're running into not one, but two separate Ruby changes that have perhaps pulled the rug out from under mysql2 / trilogy.

In all cases, the stack traces terminate at the boundary to the C extension, of course.

So, I'm wondering if you can provide any insight on how I might be able to get more useful information to diagnose things further in order to file a useful bug report with the Ruby folks.

Beyond bisecting to exact Ruby versions that introduce the issues (and crossing my fingers and hoping they're part of a patchlevel change that would never the scope of changes that could be at issue), do you have any specific diagnostic steps you'd recommend?

cernyjakub:
I see exact same behaviour on Debian 13 (stock ruby 3.3.8) and MariaDB.

cernyjakub:
Every few seconds, I see this in the syslog of the MariaDB server:

`db-2a mariadbd[265]: 2025-10-15 16:11:58 140041 [Warning] Aborted connection 140041 to db: 'my_database' user: 'myuser' host: '10.0.9.20' (Got an error reading communication packets)`

The exact same config works fine on Debian 12 with stock Ruby 3.1.2

sodabrew:
[...]

PR #1493, opened 2026-08-12
Detect a connection inherited across fork() without reconnecting

sodabrew:
## The problem

`decr_mysql2_client` (the GC finalizer path) only redirects a `Client`'s socket fd to `/dev/null` before closing (`invalidate_fd`) when `automatic_close` is explicitly set to `false`. `automatic_close` defaults to `true`. So by default, whenever a `Client` is garbage collected, its finalizer calls the real `mysql_close()` on the real, live socket, with no awareness of whether this process is the one that actually owns it.

A TCP file descriptor is shared across `fork()` -- parent and child both hold references to the same underlying kernel socket, but each process's in-memory `MYSQL*`/`SSL*` state (sequence numbers, session keys, what command was last sent) is an independent copy that `fork()` duplicated, not shared. If a worker forks (Sidekiq) or a preloaded server forks (Puma `preload_app!`-style) while a `Client` is still alive, and GC later collects that object's leftover reference in either process, `mysql_close()` fires on the shared socket -- sending a real `COM_QUIT`/TCP shutdown that pulls the connection out from under whichever process is still actually using it. Worse on TLS connections, since the finalizer also drives `SSL_shutdown()` over the yanked socket.

Even once that's fixed, a forked child that keeps using its *own* inherited reference (instead of dropping it) is still at risk: its copy of the connection's protocol/TLS state is unsynchronized with whatever the parent's connection is doing, so a query issued from the child can desync or fail outright, with no indication of why.

This looks like the root cause of at least #1415 (Sidekiq/Puma-fork-adjacent intermittent connection drops) -- see that issue for the full symptom writeup.

## The fix

- Record the pid that established the connection on every successful connect.
- In `decr_mysql2_client`, compare it against the current pid: a mismatch means this process inherited the connection across a `fork()` and never reconnected, so its copy of the connection's state can't be trusted regardless of `automatic_close`. Take the same safe `invalidate_fd` path already used for `automatic_close: false` in that case, so the real `mysql_close()`/`shutdown()` is never sent on a socket this process doesn't actually own.
- Extend the same pid check to `#query`, `#ping`, `#prepare`, and `Statement#execute` -- the actions a forked child would actually take with its (untrustworthy) inherited connection before ever reconnecting.
- Warn on stderr whenever any of these checks fires and `automatic_close` is still the default (`true`), since that's the previously-silent case an app likely doesn't know is happening. Deliberately silent when `automatic_close` is already `false`: that's the existing, documented way to tell mysql2 an app expects to inherit a connection across `fork()`, and warning there too would just be the log noise the existing close-across-forks `TODO` already exists to avoid.

No new public API: this PR is scoped to the safety fix and the warning alone. An earlier revision added an opt-in `Client#verify_not_forked!` / `Mysql2::Error::ForkSafetyError` for apps that wanted a hard failure instead of a warning, but nothing in the fix itself ever called it -- it was purely additive, and new public API is a permanent commitment once released. Dropped it to keep this PR to just the fix; it can come back as its own focused proposal later if there's real demand.

## Commits

- Record `connect_pid` at connect time; use it in `decr_mysql2_client` (the GC path) to take the safe `invalidate_fd` close instead of a real `mysql_close()`/`SSL_shutdown()` when a stale reference is collected in a process that didn't establish the connection; warn on stderr when this fires and `automatic_close` is still the default `true`.
- Extend the same check to `#query`, `#ping`, `#prepare`, and `Statement#execute` (warn-only, same `automatic_close` gating).
- Document fork safety in the README (the automatic warning and what `automatic_close: false` opts into); narrow the pre-existing close-across-forks `TODO` in `decr_mysql2_client` to describe what `connect_pid` still doesn't cover.
- Fix the warning/README wording: the original phrasing told the reader to "close and reconnect," but calling `#close` on an inherited connection sends a real `mysql_close()`/QUIT down the socket the parent may still be using -- exactly the corruption this PR exists to prevent. Corrected to "open a new Client instead," and collapsed the warning to a single line (embedded newlines break structured log ingestion).
- Remove the `verify_not_forked!` / `ForkSafetyError` API surface (see above).

## Test plan

- [x] `bundle exec rake compile` -- clean
- [x] `bundle exec rubocop` -- clean
- [x] `bundle exec rspec` -- 504 examples, 0 failures
- [x] Spec forks a child that inherits a live `Client` without reconnecting, drops the reference, and forces GC in the child -- asserts the parent's connection survives
- [x] Confirms `#query`/`#ping`/`#prepare` still work (with a warning) when called from a forked child that hasn't reconnected
- [x] Manually reproduced against the unpatched code first: same repro fails every time with "Lost connection to MySQL server during query" in the parent after the child's GC runs; confirmed the fix resolves it and the existing `automatic_close: false` fork spec stays warning-free even with the new first-use checks added

### DIFF
--- a/ext/mysql2/client.c
+++ b/ext/mysql2/client.c
@@ -16,7 +16,7 @@
 #include "mysql_enc_name_to_ruby.h"
 
 VALUE cMysql2Client;
-extern VALUE mMysql2, cMysql2Error, cMysql2TimeoutError;
+extern VALUE mMysql2, cMysql2Error, cMysql2TimeoutError, cMysql2ForkSafetyError;
 static VALUE sym_id, sym_version, sym_header_version, sym_async, sym_symbolize_keys, sym_as, sym_array, sym_stream;
 static ID intern_brackets, intern_merge, intern_merge_bang, intern_new_with_args,
   intern_current_query_options, intern_read_timeout, intern_values;
@@ -475,6 +475,30 @@ void mysql2_abandon_active_stream(mysql_client_wrapper *wrapper)
   }
 }
 
+/* See client.h. */
+int mysql2_forked_without_reconnect(mysql_client_wrapper *wrapper)
+{
+#ifndef _WIN32
+  return wrapper->connect_pid != 0 && getpid() != wrapper->connect_pid;
+#else
+  return 0;
+#endif
+}
+
+/* See client.h. */
+void mysql2_warn_forked_without_reconnect(mysql_client_wrapper *wrapper, const char *action)
+{
+#ifndef _WIN32
+  fprintf(stderr,
+    "[WARN] mysql2 Client about to %s in process %d, but its connection was "
+    "established in process %d. This process forked after connecting and "
+    "never reconnected, so the connection's state can't be trusted. Call "
+    "Client#close and reconnect after fork() to avoid this warning and the "
+    "intermittent connection errors it can cause.\n",
+    action, (int)getpid(), wrapper->connect_pid);
+#endif
+}
+
 static void *nogvl_close(void *ptr) {
   mysql_client_wrapper *wrapper = ptr;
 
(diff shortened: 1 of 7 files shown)


ext/mysql2/client.c
```diff
@@ -475,6 +475,30 @@ void mysql2_abandon_active_stream(mysql_client_wrapper *wrapper)
...
+  return 0;
+#endif
+}
+
+/* See client.h. */
+void mysql2_warn_forked_without_reconnect(mysql_client_wrapper *wrapper, const char *action)
+{
+#ifndef _WIN32
+  fprintf(stderr,
+    "[WARN] mysql2 Client about to %s in process %d, but its connection was "
```
sodabrew: This warning is way too long! Will trim it down.

PR #1497, opened 2026-08-12
Document that automatic_close cannot protect a TLS connection across fork()

### LINKED ISSUE #1467: automatic_close + fork(): TLS 1.3 session state desync breaks the parent's connection after the child GCs it
## Summary

`spec/mysql2/client_spec.rb`'s `#automatic_close should not close connections when running in a child process` fails reliably (5/5, and repeatedly across many more runs) in a local dev environment where the default connection negotiates **TLS 1.3**, but passes cleanly in every current CI matrix job I checked, including ones where TLS is genuinely available server-side. This issue documents the investigation so the finding isn't lost, since it doesn't currently affect CI and isn't blocking anything.

## The test and what it's protecting

```ruby
it "should not close connections when running in a child process" do
  run_gc
  client = Mysql2::Client.new(DatabaseCredentials['root'])
  client.automatic_close = false

  child = fork do
    client.query('SELECT 1')
    client = nil
    run_gc
  end

  Process.wait(child)

  # this will throw an error if the underlying socket was shutdown by the
  # child's GC
  expect { client.query('SELECT 1') }.to_not raise_exception
  client.close
end
```

This exercises the fork-safety fix from #213 (2014): when a `Client` with `automatic_close = false` is GC'd, `decr_mysql2_client` (`ext/mysql2/client.c`) calls `invalidate_fd()` instead of a plain `mysql_close()`. `invalidate_fd()` does `dup2(/dev/null_fd, clientfd)` — redirecting *that fd number, in the calling process's own fd table only* — before setting `wrapper->client->net.fd = -1`, so that the subsequent `mysql_close()` can't send `COM_QUIT`/`shutdown()` on the real, possibly-shared-via-`fork()` socket.

## What I confirmed is *not* the problem

I instrumented `invalidate_fd()` and `nogvl_close()` with `fprintf`/`lsof` tracing and confirmed the fd-redirection mechanism itself works exactly as designed:

- Before `dup2`: `lsof -p <child_pid> -a -i tcp` shows the real TCP socket, `ESTABLISHED`.
- After `dup2`: the same `lsof` call shows **zero** TCP fds for the child — the redirect to `/dev/null` succeeded.
- `nogvl_close()`'s `mysql_close()` runs with `net.fd == -1`, confirmed via debug print, so it's a safe no-op against the real socket.
- After `Process.wait`, `lsof` on the **parent** still shows the original connection `ESTABLISHED` at the OS/TCP level.

So the socket, at the OS layer, survives the child's teardown correctly. Yet the parent's next `client.query('SELECT 1')` still raises:

```
Mysql2::Error::ConnectionError: Lost connection to MySQL server during query
```

`SHOW GLOBAL STATUS LIKE 'Aborted_clients'` increments by exactly 1 across a run of the minimal repro below, confirming the **server** actively terminates the connection — this isn't a client-side illusion.

## Root cause: TLS record-layer state is duplicated by fork(), not shared

The default `Mysql2::Client.new(DatabaseCredentials['root'])` (no `ssl_mode` set) negotiates TLS automatically against a server with auto-generated certs (`SHOW STATUS LIKE 'Ssl_cipher'` → `TLS_AES_256_GCM_SHA384` — a TLS 1.3-only cipher suite; `tls_version` on this server is `TLSv1.2,TLSv1.3`).

`fork()` duplicates the entire process, including OpenSSL's `SSL*` session object — as two independent copies, one per process, both believing they alone own the connection's record-layer sequence numbers. Sequence:

1. Child does a **real** `client.query('SELECT 1')` over the *actual shared* socket — this happens before `client = nil`/GC, i.e. before any of the fd-invalidation code even runs. The full encrypted round-trip is exchanged for real, over the wire.
2. This advances the encryption sequence counters in lockstep on the server side and in the **child's** local OpenSSL copy.
3. The **parent's** OpenSSL copy has no way to observe that exchange — its sequence counters are frozen at their pre-fork values.
4. `invalidate_fd()`/`mysql_close()` then run in the child and correctly avoid touching the shared socket (confirmed above) — but the damage is already done; this part isn't where the corruption happens.
5. When the parent later performs its own query, it encrypts with stale sequence numbers. The server can't decrypt/verify the record, and terminates the connection — hence `Aborted_clients` +1 and `Lost connection to MySQL server during query` on the client's *next* attempt.

MySQL's own application-level packet-sequence numbering resets to 0 after each complete command, which is exactly why sharing a **plaintext** connection across `fork()` like this has worked reliably since #213 — but TLS adds an independent, session-lifetime sequence-number layer that one round-trip in the child permanently desyncs for the parent, and no amount of client-side fd bookkeeping can repair that after the fact.

### Direct confirmation

Minimal repro (no RSpec/spec_helper involved):

```ruby
require 'mysql2'
client = Mysql2::Client.new(host: '127.0.0.1', username: 'root', password: 'root', database: 'test')
# creds omit ssl_mode -> negotiates TLS automatically against this server
client.automatic_close = false

child = fork do
  client.query('SELECT 1')
  client = nil
  GC.start
  sleep 0.5
end
Process.wait(child)

begin
  client.query('SELECT 1')
  puts "OK"
rescue => e
  puts "FAIL: #{e.class}: #{e.message}"
end
```

- With the default (TLS-negotiating) connection: **fails 5/5** runs — `Lost connection to MySQL server during query`.
- With `ssl_mode: 'disabled'` added to the same repro: **passes 5/5** runs.

## Why this doesn't show up in CI today

I checked the most recent green `Build` run on `master` ([run 31218313039](https://github.com/brianmario/mysql2/actions/runs/31218313039)) job-by-job:

- **`ubuntu-22.04 ruby 3.3 mysql80`** — `ci/ssl.sh` runs and configures real server certs (`spec/ssl/*.pem`) for this job. The `#automatic_close` test still **passes**. I did not get a chance to confirm what TLS version/cipher this job's *default* (non-`ssl_client`) connection actually negotiates — possibly TLS 1.2 rather than 1.3, which wouldn't trigger the post-handshake `NewSessionTicket` / session-state asynchrony that TLS 1.3 introduces, though I want to flag this as unconfirmed rather than settled.
- **`macos-latest ruby 4.0 mariadb@11.4`** — `ci/setup.sh` explicitly skips `ci/ssl.sh` for macOS (`# TODO: get SSL working on OS X in Travis`), so the SSL-specific spec context skips with messages like `DON'T WORRY, THIS TEST PASSES - but /etc/mysql/client-key.pem does not exist.` The `#automatic_close` test passes here too (366 examples, 0 failures).
- Fedora `Container` jobs link against `mariadb-connector-c`, which may not have TLS configured/available by default either in that setup.

So: on every real CI job I checked, either TLS isn't offered at all for the default connection, or (for the Ubuntu jobs) something about that specific TLS negotiation avoids the desync I found locally. The "TLS not available" messages that show up regularly in some CI job types are consistent with — and probably explain — why this specific hazard hasn't surfaced as a real CI failure: wherever TLS isn't negotiated for the plain default connection, there's no session-layer sequence state to desync in the first place, and the original plaintext-safe fork behavior from #213 holds.

## Why I'm filing this rather than sending a fix

This reproduces reliably in one specific local dev setup (mise-installed MySQL 8.0.34 on macOS/arm64, auto-generated certs, TLS 1.3 negotiated by default) but not in any CI job I could check, so it isn't blocking anything today. But the underlying hazard is real and not specific to my machine: **any deployment where the default connection negotiates TLS and code relies on `automatic_close = false` to survive a `fork()`** (a real pattern — e.g. pre-forking app servers keeping a shared DB handle) is exposed to this. I don't think there's a client-side fix possible — the corruption happens the instant the child performs I/O on the shared connection, before any of `invalidate_fd()`'s cleanup runs — so this is likely a documentation issue (the `automatic_close` docs/comments should warn that this only protects plaintext connections) or, at minimum, this spec should pin `ssl_mode: 'disabled'` so it's testing what it claims to test regardless of what the ambient default connection would otherwise negotiate.

Happy to send a PR for the doc/spec clarification if that's a useful next step — wanted to get the investigation written down first since it took a while to pin down.

sodabrew:
## Summary

The `fd`-invalidation trick `automatic_close = false` relies on (`invalidate_fd()` in `client.c`) only patches up the raw socket fd -- it says nothing about a TLS connection's OpenSSL session/record-layer state, which `fork()` duplicates right along with the fd as two independent copies. The moment either side performs a real query, that copy's sequence counters advance independently of the other's, permanently desyncing them; the next query on the stale side gets rejected by the server (`Aborted_clients` +1, `Lost connection to MySQL server during query`). There's no client-side fix -- `invalidate_fd()` can't repair TLS state it never touches in the first place.

The regression spec covering this (child process + `automatic_close`) was already pinned to `ssl_mode: disabled` in #1468, with the same reasoning written in as a comment. This documents the same finding for API consumers, since nothing previously said `automatic_close` only works for plaintext connections.

## Changes

- `README.md`: new "Fork safety and `automatic_close`" section explaining the mechanism and the fix (`ssl_mode: :disabled`).
- `ext/mysql2/client.c`: same caveat added to the `automatic_close=` docstring.

## Test plan

- [x] `spec/mysql2/client_spec.rb`'s `#automatic_close` context passes (151 examples, 0 failures overall)
- [x] No code behavior changes -- docs only

Fixes #1467.

### DIFF
--- a/README.md
+++ b/README.md
@@ -363,6 +363,30 @@ When secure_auth is enabled, the server will refuse a connection if the account 
 The MySQL 5.6.5 client library may also refuse to attempt a connection if provided an older format password.
 To bypass this restriction in the client, pass the option `:secure_auth => false` to Mysql2::Client.new().
 
+### Fork safety and `automatic_close`
+
+By default (`automatic_close = true`), a `Client` closes its connection when
+garbage collected. Set `:automatic_close => false` (or `client.automatic_close
+= false`) to leave the connection open instead -- the classic reason is a
+pre-forking app server that opens a connection before `fork()` and wants the
+parent to go on using it after a forked child's copy of the same `Client` is
+garbage collected.
+
+This only works for a **plaintext** connection. A TLS connection's OpenSSL
+session state -- record-layer sequence numbers -- is duplicated by `fork()`
+right along with the socket, as two independent copies that both believe they
+alone own the connection. If the child performs any real query on its copy,
+that advances the encryption sequence counters on the server and in the
+child's copy, but not in the parent's now-desynced copy; the parent's next
+query on that connection will be encrypted with stale sequence numbers, and
+the server will reject it and drop the connection (`Aborted_clients` +1,
+`Mysql2::Error::ConnectionError: Lost connection to MySQL server during
+query` on the client's next attempt). `automatic_close = false` and
+`invalidate_fd()` (the fd-redirection this relies on) only ever touch file
+descriptors, not TLS session state, so there is no client-side fix for this --
+if you rely on `automatic_close = false` surviving a `fork()`, connect with
+`:ssl_mode => :disabled`.
+
 ### Flags option parsing
 
 The `:flags` parameter accepts an integer, a string, or an array. The integer
--- a/ext/mysql2/client.c
+++ b/ext/mysql2/client.c
@@ -1640,6 +1640,11 @@ static VALUE get_automatic_close(VALUE self) {
  * collected. To avoid "Aborted connection" errors on the server, explicitly
  * call +close+ when the connection is no longer needed.
  *
+ * This only protects a plaintext connection across fork(); a TLS
+ * connection's OpenSSL session state is duplicated (not shared) by fork()
+ * and will desync between parent and child the moment either side performs
+ * a real query, regardless of this setting. See the README for details.
+ *
  * @see http://dev.mysql.com/doc/en/communication-errors.html
  */
 static VALUE set_automatic_close(VALUE self, VALUE value) {


sodabrew:
Folding this into #1493 -- both are documentation-only, and #1493 already carries the Fork safety README section this extends. The TLS/automatic_close paragraph and the client.c docstring caveat from this PR are now in #1493.

PR #1501, opened 2026-08-12
Document that MULTI_STATEMENTS only returns the first statement's result

### LINKED ISSUE #723: Query running in async mode, without it being specified.
I'm loading my rails `structure.sql` file via the mysql2 gem, but the query executes asynchronously, without`async: true` being specified.

The sql file contains the typical rails content,with the exception of including trigger definitions, resulting in the sequence (`CREATE TABLE`s... `CREATE TRIGGER`s... `INSERT INTO schema_migrations`'s).

The connection is established using `Mysql2::Client::MULTI_STATEMENTS`.

The problem is that when I execute `Mysql2::Client#query()`, `nil` is returned immediately, while I can see the sql being executed in the background.
Out of curiosity, I've tried adding `async: true` and waiting for `client.async_result` to return some value, but even when the background work was finished, such call was returning `nil`. 

Based on what I see in the gem documentation, this should not happen.

My O/S is an Ubuntu 64 15.10, with the standard `libmysqlclient` packages installed.
I've tried with the 0.3.17 and 0.4.2, and the behavior is the same.

sodabrew:
## Summary

`client.query()` returning `nil` for a batch that starts with `CREATE TABLE`/`INSERT` statements, while the rest of the batch keeps running, reads as async execution if you don't already know `MULTI_STATEMENTS`' shape: only the first statement's result comes back from `query()` (or `async_result`, if `:async => true` was passed) -- `nil` there just means that particular statement had no result set, same as it would outside `MULTI_STATEMENTS`. The remaining statements aren't deferred -- the server runs the whole batch regardless -- but retrieving their results, or finding out whether any of them errored, requires looping over `next_result`/`store_result`, and skipping that loop leaves the connection undrained (confirmed: reusing it without draining or closing raises `"Commands out of sync"`).

This was asked and answered in #723 back in 2016 ("the implementation is working as intended") but never made it into the README.

## Changes

- `README.md`: new paragraph in the "Multiple result sets" section explaining this.

## Test plan

- [x] Docs only, no code changes
- [x] Manually confirmed the "Commands out of sync" claim against a real connection before writing it

Fixes #723.

### DIFF
--- a/README.md
+++ b/README.md
@@ -442,7 +442,7 @@ while client.next_result
 end
 ```
 
-A single call to `client.query` only ever returns the *first* statement's
+The call to `client.query` returns the *first* statement's
 result -- or `nil`, if that first statement doesn't produce one at all (e.g.
 `CREATE TABLE` or `INSERT`, same as outside of `MULTI_STATEMENTS`). This is
 still true if you pass `:async => true`: `client.async_result` also only
--- a/README.md
+++ b/README.md
@@ -446,14 +446,21 @@ A single call to `client.query` only ever returns the *first* statement's
 result -- or `nil`, if that first statement doesn't produce one at all (e.g.
 `CREATE TABLE` or `INSERT`, same as outside of `MULTI_STATEMENTS`). This is
 still true if you pass `:async => true`: `client.async_result` also only
-returns the first statement's result. The rest of the batch is not deferred
-until you ask for it -- the server runs every statement in the batch to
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -442,6 +442,26 @@ while client.next_result
 end
 ```
 
+A single call to `client.query` only ever returns the *first* statement's
```
sodabrew:
```suggestion
The call to `client.query` returns the *first* statement's
```

PR #1506, opened 2026-08-12
Fix ssl_mode: :verify_ca silently accepting any CA on MariaDB

### LINKED ISSUE #936: CA verification not respected when `ssl_mode` flag is passed
## Prerequisites:
- MYSQL2 GEM is linked against MariaDB Connector/C 3.0.2

## Issue:

When trying to establish a connection as below, even when giving an incorrect CA that did not sign the server certificate, a connection is established successfully:

```
:ssl_mode => "verify_ca",  :sslca => "path-to-incorrect-ca"
```

Any reason why this could be happening ?

sodabrew:
## The problem

`ssl_mode: :verify_ca` silently accepts a connection with a wrong CA on MariaDB Connector/C. Traced to `client.c`'s compatibility shim: MariaDB Connector/C has no `SSL_MODE_VERIFY_CA` enum, so `SSL_MODE_VERIFY_CA` never got a real value distinct from `SSL_MODE_PREFERRED` -- both fell through to the same `INT2NUM(0)` fallback used for any `SSL_MODE_*` constant the shim doesn't explicitly define. `ssl_mode: :verify_ca` therefore resolves to `0`, which `rb_set_ssl_mode_option`'s MariaDB branch doesn't recognize (it only handles `DISABLED=1`, `REQUIRED=3`, `VERIFY_IDENTITY=5`), so it prints a warning and configures no SSL option at all -- the connection proceeds over opportunistic TLS with the CA never checked.

Confirmed live against a real MariaDB 11.8 server: a self-signed, completely unrelated CA is accepted without error under `ssl_mode: :verify_ca`.

This is #936, open since 2018.

## The fix

- Give `SSL_MODE_VERIFY_CA` its own value (`4`, matching real MySQL's own enum ordering) in the compatibility shim, instead of falling through to the shared `0` fallback.
- Map `SSL_MODE_VERIFY_CA` alongside `SSL_MODE_VERIFY_IDENTITY` to `MYSQL_OPT_SSL_VERIFY_SERVER_CERT` in `rb_set_ssl_mode_option`. MariaDB Connector/C has no CA-only verification mode -- `MYSQL_OPT_SSL_VERIFY_SERVER_CERT` is the only verification it offers, checking CA and hostname together. Mapping `verify_ca` to the same full verification `verify_identity` gets is the closest honest behavior available, and strictly safer than the previous silent no-op.
- Extends `#879`'s finding too: `verify_identity` was already correctly wired to this same option, which is why that issue is separately resolved -- this PR only closes the `verify_ca` gap.

## Verification

Built and tested against a real MariaDB 11.8 server (MacPorts `mariadb-11.8`, which self-reports as MariaDB Connector/C 3.4.9 via `mysql_get_client_version()` -- the version numbering this code path actually branches on):

```
Before fix:
  SSL_MODE_VERIFY_CA = 0, SSL_MODE_PREFERRED = 0  (collision)
  ssl_mode: :verify_ca + wrong CA -> CONNECTED (bug)

After fix:
  SSL_MODE_VERIFY_CA = 4  (distinct)
  ssl_mode: :verify_ca + wrong CA   -> Mysql2::Error::ConnectionError: TLS/SSL error: self-signed certificate
  ssl_mode: :verify_ca + correct CA -> connects normally, ssl_cipher=TLS_AES_256_GCM_SHA384
```

## Test plan

- [x] `bundle exec rake compile` -- clean (both MySQL and MariaDB Connector/C)
- [x] `bundle exec rubocop` -- clean
- [x] `bundle exec rspec` -- 501 examples, 0 failures
- [x] Extended the existing per-`ssl_mode` coverage loop to include `verify_ca` on the non-`FULL_SSL_MODE_SUPPORT` (MariaDB) code path, so CI's MariaDB matrix jobs exercise this directly
- [x] Added a new regression spec that generates a throwaway self-signed CA at runtime and asserts `ssl_mode: :verify_ca` rejects it
- [x] Manually verified against a real MariaDB 11.8 server (see above) since this repo's local dev setup here links MySQL by default

Fixes #936.

### DIFF
--- a/README.md
+++ b/README.md
@@ -335,10 +335,14 @@ The `:ssl_mode` option will also set the appropriate MariaDB connection flags:
 | ---                | ---                                  |
 | `:disabled`        | MYSQL_OPT_SSL_ENFORCE = 0            |
 | `:required`        | MYSQL_OPT_SSL_ENFORCE = 1            |
+| `:verify_ca`       | MYSQL_OPT_SSL_VERIFY_SERVER_CERT = 1 |
 | `:verify_identity` | MYSQL_OPT_SSL_VERIFY_SERVER_CERT = 1 |
 
-MariaDB does not support the `:preferred` or `:verify_ca` options. For more
-information about SSL/TLS in MariaDB, see
+MariaDB Connector/C has no CA-only verification mode, so `:verify_ca` gets
+the same full verification (CA and hostname) that `:verify_identity` gets --
+MYSQL_OPT_SSL_VERIFY_SERVER_CERT is the only verification it offers. MariaDB
+does not support the `:preferred` option; there's no equivalent to fall back
+to. For more information about SSL/TLS in MariaDB, see
 [https://mariadb.com/kb/en/securing-connections-for-client-and-server/](https://mariadb.com/kb/en/securing-connections-for-client-and-server/)
 and [https://mariadb.com/kb/en/mysql_optionsv/#tls-options](https://mariadb.com/kb/en/mysql_optionsv/#tls-options)
 
--- a/ext/mysql2/client.c
+++ b/ext/mysql2/client.c
@@ -76,7 +76,14 @@ static ID intern_brackets, intern_merge, intern_merge_bang, intern_new_with_args
  */
 #ifdef HAVE_CONST_MYSQL_OPT_SSL_VERIFY_SERVER_CERT
   #define SSL_MODE_VERIFY_IDENTITY 5
+  /* MYSQL_OPT_SSL_VERIFY_SERVER_CERT is the only verification this client
+   * library offers -- it checks the CA and the hostname together, with no
+   * way to ask for one without the other. SSL_MODE_VERIFY_CA maps to the
+   * same option: full verification is the closest honest behavior available,
+   * and strictly safer than the alternative of accepting any CA silently. */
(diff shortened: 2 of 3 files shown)


sodabrew:
Correction to a claim in this PR's own comment (`client.c`, the `SSL_MODE_VERIFY_CA` definition): "MYSQL_OPT_SSL_VERIFY_SERVER_CERT is the only verification this client library offers -- it checks the CA and the hostname together, with no way to ask for one without the other."

That's not accurate -- it checks CA validity only, not hostname. Confirmed via #1568's new regression spec (fails on every CI leg that runs real certs: `ubuntu-24.04` mariadb11.4/11.8/12.3, both `fedora` legs) and independently via a local repro against a fresh MacPorts MariaDB 11.8.8 + Connector/C 3.4.9 instance built with this repo's own `spec/ssl/*` fixtures -- a hostname the certificate doesn't cover still connects under `ssl_mode: :verify_identity`.

Doesn't change what this PR fixed -- mapping `:verify_ca` onto the same option `:verify_identity` uses is still strictly better than the previous silent no-op, since CA validity is at least checked now. But the comment's description of what that option actually verifies overstated it, and `:verify_identity` doesn't yet deliver on its name on MariaDB. Full writeup and repro on #879, which this doesn't close after all.

PR #1508, opened 2026-08-12
Fixes #1392: Thread#exit mid-query no longer permanently locks the connection

### LINKED ISSUE #1392: With `Thread#exit`, client object becomes incorrect state
From: https://github.com/rails/rails/issues/54441

## Steps to reproduce

- Spawn threads, reusing `client`.
- Call `client#query` in the threads.
- Kill the threads by `Thread#exit`.

In this scenario, C-level `active_fiber` is not cleared, but ruby-level client is still alive.
Therefore the next call of `query` throws `'Mysql2::Client#_query': This connection is in use by: #<Fiber:0x00007fef3d80f2a8 (terminated)>` exception.

```
require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"
  gem "bigdecimal"
  gem "mysql2"
end

client = Mysql2::Client.new(:host => "127.0.0.1", :username => "root", :password => "password", :database => "test_datab
ase")

100.times do
  th = Thread.new do
    loop do
      results = client.query("SELECT * FROM posts")
    end
  end
  sleep(0.1)
  th.exit
end
```

I found this issue on Rails.
ActiveRecord reuses client objects in a connection pool.
My problem was the broken clients were kept in the pool and following requests failed.
I fixed the problem by avoiding `Thread#exit`, but I apprecated if Mysql2 soundly signaled the connection was broken and AR needed to start a new connection.

sodabrew:
This PR previously depended on #1496 (its branch was built directly on top of `fix-777-ping-interrupt-lockout`, so the diff temporarily included #1496's commit as well). #1496 has since merged into `master`, and this branch has been rebased onto it -- the diff now reflects only this PR's own change (`ext/mysql2/client.c` and `spec/mysql2/client_spec.rb`).

## Summary

`query()`'s send phase (`do_send_query`) and read-wait phase (`do_query`) were only protected by `rb_rescue2`, which catches ordinary Ruby exceptions -- but not `Thread#exit`'s non-exception unwind. If a thread was killed with `Thread#exit` while either was in flight, the cleanup that clears `wrapper->active_fiber` never ran, leaving the connection permanently stuck reporting `"This connection is in use by: <dead fiber>"` for every future caller, including the same `Client` object reused later from an ActiveRecord connection pool -- exactly the scenario the original report described.

## Relationship to #777

Same underlying issue as #777 (interrupted `#ping`): `rb_rescue2` only catches actual Ruby exceptions, and `Thread#exit` isn't one. The fix here takes a different shape than #777's, though, since `query()`'s send and read-wait phases don't clear `active_fiber` on their own success path the way `#ping` does -- `active_fiber` has to stay set *between* them for the subsequent read to proceed. So rather than reusing the "check if `active_fiber` is still set" idiom from `#ping`/`#next_result`, each phase now carries an explicit `completed` flag, set only once it reaches its own normal return, and an `rb_ensure` companion that runs the same invalidate-and-clear cleanup as before whenever that flag is still unset -- covering a raised exception exactly as before, and now `Thread#exit` too, without needing to know which kind of unwind actually happened.

`disconnect_and_raise` remains, now as a thin wrapper around a shared `invalidate_after_interrupted_query()` helper: `#ping`'s own `rb_rescue2` still uses it as-is, since a single blocking call has no equivalent mid-phase-boundary window to worry about.

## Related, out of scope here

Audited `statement.c`'s `execute` path (`Statement#execute`'s own send/read split) for the same hazard: it currently has **no** `rb_rescue2` or `rb_ensure` protection at all, a separate and likely larger gap. Out of scope here since the original report is specifically about `Client#query`, but worth a follow-up.

## Changes

- `ext/mysql2/client.c`: add a `completed` flag to both `nogvl_send_query_args` and `async_query_args`, set only on normal completion; switch both call sites from `rb_rescue2` to `rb_ensure` with companions that clean up only when `completed` is still unset, covering both raised exceptions and `Thread#exit`-style unwinds uniformly.
- `spec/mysql2/client_spec.rb`: regression spec using the `FreezableProxy` helper (from #777's fix) to keep a query reliably in flight, calling `Thread#exit` on the thread running it, and asserting the connection recovers with a normal error afterward rather than the permanent lockout message.

## Test plan

- [x] New regression spec added and passing
- [x] Verified the new spec fails against the pre-fix code (the original lockout message) and passes against the fix
- [x] Full `spec/` suite passes (504 examples, 0 failures, on top of `master` post-#1496)
- [x] rubocop clean

Fixes #1392.

### DIFF
--- a/ext/mysql2/client.c
+++ b/ext/mysql2/client.c
@@ -110,6 +110,12 @@ struct nogvl_send_query_args {
   const char *sql_ptr;
   long sql_len;
   mysql_client_wrapper *wrapper;
+  /* Set by do_send_query itself, only once mysql_send_query has actually
+   * returned -- lets disconnect_send_query_if_incomplete (its rb_ensure
+   * companion) tell a normal completion (success or an error already
+   * raised and handled here) apart from any other kind of unwind, e.g.
+   * Thread#exit, which do_send_query never gets a chance to react to. */
+  int completed;
 };
 
 /*
@@ -839,10 +845,12 @@ static VALUE do_send_query(VALUE args) {
   struct nogvl_send_query_args *query_args = (void *)args;
   mysql_client_wrapper *wrapper = query_args->wrapper;
   if ((VALUE)rb_thread_call_without_gvl(nogvl_send_query, query_args, RUBY_UBF_IO, 0) == Qfalse) {
-    /* an error occurred, we're not active anymore */
-    wrapper->active_fiber = Qnil;
+    /* An error occurred: raise it and let disconnect_send_query_if_incomplete
+     * (this call's rb_ensure companion) do the cleanup, same as any other
+     * kind of unwind out of this function. */
     rb_raise_mysql2_error(wrapper);
   }
+  query_args->completed = 1;
   return Qnil;
 }
 
@@ -956,11 +964,18 @@ static VALUE rb_mysql_client_async_result(VALUE self) {
 struct async_query_args {
   int fd;
   VALUE self;
+  /* Set by do_query itself, only once its wait loop has actually finished.
+   * Same purpose as nogvl_send_query_args.completed below: lets this
+   * struct's rb_ensure companion tell a normal completion (success, or an
+   * error already raised and handled here) apart from any other kind of
+   * unwind, e.g. Thread#exit. */
+  int completed;
 };
 
-static VALUE disconnect_and_raise(VALUE self, VALUE error) {
-  GET_CLIENT(self);
-
+/* Shared cleanup for an interrupted send/read/ping: none of them reached
+ * their own normal completion, so the connection may be left
+ * mid-protocol-exchange. */
+static void invalidate_after_interrupted_query(mysql_client_wrapper *wrapper) {
   wrapper->active_fiber = Qnil;
   wrapper->state = MYSQL2_CLIENT_IDLE;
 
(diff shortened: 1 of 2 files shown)


sodabrew:
A few suggestions to trim comment verbosity and reduce duplication between the two new `rb_ensure` trampolines.

**Trim the `completed` field comments** (currently 5 lines each, saying almost the same thing twice):

```c
// ext/mysql2/client.c, nogvl_send_query_args
/* Set once do_send_query finishes normally; lets its rb_ensure
 * companion tell that apart from a Thread#exit-style unwind. */
int completed;
```

```c
// ext/mysql2/client.c, async_query_args
/* Same purpose as nogvl_send_query_args.completed above. */
int completed;
```

**Trim `disconnect_and_raise`'s comment:**

```c
/* rb_rescue2 companion (do_ping): re-raises after cleanup. do_ping is a
 * single blocking call, so it has no Thread#exit window to worry about. */
static VALUE disconnect_and_raise(VALUE self, VALUE error) {
```

**Trim the block comment above the two new trampolines** (currently 9 lines):

```c
/* rb_ensure companions (do_send_query, do_query): unlike disconnect_and_raise,
 * these never raise -- rb_ensure runs them during any unwind, including a
 * non-exception one like Thread#exit, which rb_rescue2 can't see. */
static VALUE disconnect_send_query_if_incomplete(VALUE argsval) {
```

**Share code between the two trampolines:** `async_query_args` doesn't carry a `wrapper` pointer today, so `disconnect_async_query_if_incomplete` has to re-derive it via `GET_CLIENT(async_args->self)`, while `disconnect_send_query_if_incomplete` already has `query_args->wrapper` directly. Giving `async_query_args` a `wrapper` field too (set alongside `self`/`completed` in `rb_mysql_query`, right where `async_args.self = self;` is) removes that asymmetry:

```c
struct async_query_args {
  int fd;
  VALUE self;
  mysql_client_wrapper *wrapper;
  int completed;
};
```

```c
static VALUE disconnect_async_query_if_incomplete(VALUE argsval) {
  struct async_query_args *async_args = (void *)argsval;

  if (!async_args->completed) {
    invalidate_after_interrupted_query(async_args->wrapper);
  }

  return Qnil;
}
```

Both trampolines are then textually identical modulo the struct type, which makes it obvious at a glance that they do the same thing -- smaller diff, no `GET_CLIENT` call to reason about in the async path.

sodabrew:
Done, see 49eb04f — trimmed the comments and gave `async_query_args` a `wrapper` field so both `rb_ensure` trampolines are textually identical modulo struct type.

Issue #1514, opened 2026-08-13
Proposal: build array rows with one rb_ary_new_from_values instead of per-cell rb_ary_push


jeremy:
Part of the roadmap in #1510.
**Category: low-hanging fruit** · `effort:S · risk:low · ready:now · layer:C`

## Problem

Array-mode rows are built cell by cell: `rb_ary_new2(numberOfFields)` up front, then one `rb_ary_push` per cell — live at `result.c:1149` (stmt path), `result.c:1257` (cast-none loop), and `result.c:1481`/`:1487` (general loop). Each push re-pays a capacity check, a frozen check, and a per-element write barrier, N times per row, even though the final length is known before the first cell is cast.

## Proposal

Fill a scratch `VALUE[]` in the cell loop and build each row with a single `rb_ary_new_from_values(numberOfFields, buf)` — exact-capacity allocation, one bulk memcpy, one write-barrier pass.

- Scratch is `ALLOCV_N(VALUE, n)`: alloca for small n, a GC-marked imemo_tmpbuf beyond that, so cells constructed into it are conservatively marked and survive a GC mid-row.
- Allocate the scratch **once per `#each` call** (numberOfFields is fixed for the result), not per row.
- Mid-row raises exist — invalid date at `result.c:1387`/`:1457`, stmt fetch error at `result.c:981` — and a raise that skips `ALLOCV_END` leaks the tmpbuf until GC. Per-`#each` allocation plus `ALLOCV_END` after the row loop bounds that to one buffer per raised iteration; that is the standard ALLOCV trade and I'd document it in the diff rather than armor it.
- **Array mode only.** Hash mode is untouched: `rb_hash_aset` has no batch equivalent, and #1460's `rb_hash_new_capa` pre-sizing (`result.c:1226`) already took the hash-side win.

Sequencing note: this builds the exact GC-visible scratch machinery the splat-row-mode proposal (same roadmap) reuses for `rb_yield_values2`, so it ships first as the substrate — same code region, push vs. yield.

## Evidence and precedent

- **trilogy** PR #280 is precisely this change: three `ALLOCV_N` scratch buffers (`cext.c:952-954`), cells cast into scratch, rows built via `rb_ary_new_from_values(column_count, row_ruby_values)` — with byroot's rationale on the PR that repeated `rb_ary_push` re-checks capacity and frozen state per cell and it's "best to keep values in a scratch space and create arrays all at once".
[...]

PR #1533, opened 2026-08-13
Build array rows with one rb_ary_new4 instead of per-cell pushes

jeremy:
Proposed in #1514.

Array-mode rows were built cell by cell: rb_ary_new2(numberOfFields) up front, then one rb_ary_push per cell, in all three fetch loops (the text cast-none loop, the text general loop, and the binary-protocol stmt loop). Each push re-pays a capacity check, a frozen check, and a per-element write barrier, N times per row, even though the final length is known before the first cell is cast.

Now each row's cells are cast into a VALUE scratch buffer and the row is built with a single rb_ary_new4(numberOfFields, scratch): exact-capacity allocation, one bulk copy, one write-barrier pass. rb_ary_new4 is the long-stable alias of rb_ary_new_from_values -- the modern name only exists on Ruby 2.1+ while the gemspec floor is 2.0.0, so the alias is the one spelling that compiles across the entire supported range; it matches the rb_ary_new2 spelling already used throughout this file, and it carries no deprecation attribute on any current Ruby or on ruby/ruby master (it is a plain macro alias, same status as rb_ary_new2). trilogy made precisely this change in trilogy#280, and pg builds its rows the same way (pg_result.c, VALUE scratch + rb_ary_new4).

The scratch is allocated once per #each call, not per row, in rb_mysql_result_each -- ALLOCV_N(VALUE, holder, mysql_num_fields(...)), so it is alloca for narrow results and a GC-visible imemo_tmpbuf beyond that; either way the VALUEs parked in it between construction and the row build are conservatively marked and survive (and are pinned across) a mid-row GC. Hash mode is untouched: rb_hash_aset has no batch equivalent, and the rb_hash_new_capa pre-sizing already took the hash-side win. Nothing in the scratch outlives the call: the buffer is released with ALLOCV_END when rb_mysql_result_each_ returns.

Safety cases, called out explicitly:

* Freed results: the scratch is only allocated when !wrapper->resultFreed, so mysql_num_fields never touches a freed MYSQL_RES; a result freed from inside the iteration block is caught by the fetch functions' existing freed-result guards before any scratch write.
* Mid-row raises (invalid DATETIME/DATE cast, stmt fetch error) and a `break` from the block skip ALLOCV_END and leak the heap form of the buffer until GC reclaims the holder. That is the standard ALLOCV trade, and it is the trade trilogy shipped in trilogy#280: its merged loops raise Trilogy::CastError mid-row (cast.c invalid date/double arms) and return early past three ALLOCV buffers with no ALLOCV_END and no rb_ensure. pg sidesteps the question structurally -- its scratch is a per-row stack VLA (PG_VARIABLE_LENGTH_ARRAY), which unwinds for free but re-pays the allocation every row. Here the leak is bounded to one buffer per raised iteration precisely because the allocation is per-#each rather than per-row, and is disclosed rather than armored with rb_ensure.
* Rows are byte-identical: every slot is written on every row (including explicit Qnil for NULL cells), so nothing bleeds between rows.

Specs: wide-row (40-column) field-order parity through the text-cast, text-cast:false, streaming, and prepared-statement paths; a NULL-interleave pin that catches stale scratch slots; and a mid-row cast raise (invalid date via sql_mode='') followed by a live query on the same connection. Verified the specs bite: reversing the scratch fill order fails all six parity/interleave pins (the mid-row-raise spec is a regression guard for the leak path and passes on master too, as expected).

Benchmark: wide array-mode rows (20,000 rows x 40 columns: 20 INT + 20 VARCHAR(16)), `each {}` in array mode, three runs per arm alternating base/branch, min of 5 timings per sample, 15 samples per run, per-run median (range) in ms per pass. AMD Ryzen 9 7950X, Arch Linux x86_64, Ruby 4.0.6, MySQL 9.6.0 (Docker, TCP loopback), client libmariadb 12.3.2, machine load < 1.5 for every run:

| mode        | arm          | run 1               | run 2               | run 3               |
|-------------|--------------|---------------------|---------------------|---------------------|
| text cast   | base b4aff39 | 45.98 (45.63-46.40) | 45.93 (45.47-46.66) | 45.65 (45.33-45.97) |
| text cast   | this patch   | 42.46 (42.27-42.75) | 41.67 (41.49-42.35) | 41.85 (41.59-42.24) |
| cast: false | base b4aff39 | 50.02 (49.74-51.71) | 49.82 (49.45-50.39) | 49.76 (49.32-50.19) |
| cast: false | this patch   | 45.50 (45.23-45.85) | 45.41 (44.84-45.78) | 45.26 (45.04-46.09) |
| prepared    | base b4aff39 | 46.35 (45.98-47.23) | 45.47 (44.72-46.21) | 40.87 (40.50-41.19) |
| prepared    | this patch   | 42.09 (41.92-42.35) | 41.41 (41.01-42.22) | 38.89 (38.73-39.67) |

Median of medians: text cast 45.93 -> 41.85 ms (-8.9%, 57.4 -> 52.3 ns per value), cast: false 49.82 -> 45.41 ms (-8.9%), prepared 45.47 -> 41.41 ms (-8.9%). Within every pair the branch range sits entirely below the base range. One drift note, disclosed: the prepared-path base dropped to ~41 ms in run 3 (both arms of that pair ran faster); the branch still beat it pairwise with no overlap (-4.8% for that pair).

Full suite: 509 examples, 0 failures, 10 pending (SSL-gated), RuboCop clean. Ruby 4.0.6, MySQL 9.6 (Homebrew, libmysqlclient), macOS arm64.

### DIFF
--- a/spec/mysql2/statement_spec.rb
+++ b/spec/mysql2/statement_spec.rb
@@ -129,6 +129,7 @@ RSpec.describe Mysql2::Statement do # rubocop:disable Metrics/BlockLength
     stmt = @client.prepare 'SELECT 1 AS a, NULL AS b UNION SELECT NULL, 2'
     expect(stmt.execute(as: :array).to_a).to eq([[1, nil], [nil, 2]])
   end
+
   it "should raise TypeError naming the parameter index, class, and value for an unsupported bind type" do
     stmt = @client.prepare 'SELECT ?, ?'
 
--- a/spec/mysql2/statement_spec.rb
+++ b/spec/mysql2/statement_spec.rb
@@ -119,6 +119,17 @@ RSpec.describe Mysql2::Statement do # rubocop:disable Metrics/BlockLength
     expect(stmt.execute(1, as: :array).first).to eq([1])
   end
 
+  it "should yield every cell in field order for wide array rows" do
+    width = 40
+    stmt = @client.prepare "SELECT #{Array.new(width) { |i| "#{i} AS c#{i}" }.join(', ')}"
+    expect(stmt.execute(as: :array).first).to eq((0...width).to_a)
+  end
+
+  it "should not carry array row cells over between rows" do
+    stmt = @client.prepare 'SELECT 1 AS a, NULL AS b UNION SELECT NULL, 2'
+    expect(stmt.execute(as: :array).to_a).to eq([[1, nil], [nil, 2]])
+  end
+
   it "should keep its result after other query" do
     @client.query 'USE test'
     @client.query 'CREATE TABLE IF NOT EXISTS mysql2_stmt_q(a int)'
--- a/spec/mysql2/statement_spec.rb
+++ b/spec/mysql2/statement_spec.rb
@@ -129,7 +129,6 @@ RSpec.describe Mysql2::Statement do # rubocop:disable Metrics/BlockLength
     stmt = @client.prepare 'SELECT 1 AS a, NULL AS b UNION SELECT NULL, 2'
     expect(stmt.execute(as: :array).to_a).to eq([[1, nil], [nil, 2]])
   end
-  
   it "should raise TypeError naming the parameter index, class, and value for an unsupported bind type" do
     stmt = @client.prepare 'SELECT ?, ?'
 
--- a/ext/mysql2/result.c
+++ b/ext/mysql2/result.c
@@ -84,6 +84,12 @@ typedef struct {
    * per row. Changing it mid-iteration therefore no longer affects later rows
    * of that same call; the next #each call observes the new value. */
   rb_encoding *default_internal_enc;
+  /* Array-mode cell scratch: one slot per field, ALLOCV-allocated (so the
+   * VALUEs written into it are GC-visible) once per #each call and reused for
+   * every row. Each row's cells are cast into it and the row is built with a
+   * single rb_ary_new4 instead of one rb_ary_push per cell. NULL in hash mode
+   * and when the fetch functions can never run (freed result). */
+  VALUE *rowScratch;
 } result_each_args;
 
 extern VALUE mMysql2, cMysql2Client, cMysql2Error;
@@ -904,7 +910,7 @@ static void rb_mysql_result_alloc_result_buffers(VALUE self, MYSQL_FIELD *fields
 
 static VALUE rb_mysql_result_fetch_row_stmt(VALUE self, MYSQL_FIELD * fields, const result_each_args *args)
 {
-  VALUE rowVal;
+  VALUE rowVal = Qnil;
   unsigned int i = 0;
 
   rb_encoding *default_internal_enc;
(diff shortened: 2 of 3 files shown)


sodabrew:
Ready for rebase

spec/mysql2/statement_spec.rb
```diff
@@ -119,6 +119,17 @@ def stmt_count
...
+    width = 40
+    stmt = @client.prepare "SELECT #{Array.new(width) { |i| "#{i} AS c#{i}" }.join(', ')}"
+    expect(stmt.execute(as: :array).first).to eq((0...width).to_a)
+  end
+
+  it "should not carry array row cells over between rows" do
+    stmt = @client.prepare 'SELECT 1 AS a, NULL AS b UNION SELECT NULL, 2'
+    expect(stmt.execute(as: :array).to_a).to eq([[1, nil], [nil, 2]])
+  end
+
```
sodabrew:
```suggestion

```

spec/mysql2/statement_spec.rb
```diff
@@ -119,6 +119,16 @@ def stmt_count
...
+  it "should yield every cell in field order for wide array rows" do
+    width = 40
+    stmt = @client.prepare "SELECT #{Array.new(width) { |i| "#{i} AS c#{i}" }.join(', ')}"
+    expect(stmt.execute(as: :array).first).to eq((0...width).to_a)
+  end
+
+  it "should not carry array row cells over between rows" do
+    stmt = @client.prepare 'SELECT 1 AS a, NULL AS b UNION SELECT NULL, 2'
+    expect(stmt.execute(as: :array).to_a).to eq([[1, nil], [nil, 2]])
+  end
```
sodabrew, line 131:
```suggestion
  end

```

sodabrew:
Reviewed via static inspection of the GVL/GC-safety argument specifically, not just re-reading the description (no prior `ALLOCV` usage exists anywhere else in this codebase, so this is a genuinely new technique here, not an established pattern).

**GVL timing.** Read the full, unchanged surrounding code in both `rb_mysql_result_fetch_row` and `rb_mysql_result_fetch_row_stmt`. Both already carry the comment this PR's safety depends on: "The release is kept as tight as possible around the client-library call because the GVL is required again immediately to build Ruby objects." The GVL is released once per row, strictly before any cell casting begins, and the per-cell loop that writes into `args->rowScratch[i]` runs entirely after it's reacquired. This PR doesn't touch that structure, so it inherits an existing safety invariant rather than introducing a new one.

**GC-visibility.** Confirmed the `ALLOCV_N`/`ALLOCV_END` pattern here, holder declared in the outer frame and touched only at allocation and release, matches how CRuby's own sources (`array.c`, `string.c`, etc.) use this API. For the `alloca` path, the buffer lives in `rb_mysql_result_each`'s own still-live stack frame, covered by ordinary conservative stack scanning. For the `imemo_tmpbuf` path, the contents are marked as part of marking the tmpbuf object itself, which is exactly what that type exists for.

**The mid-row-raise "leak" is more benign than the disclosure implies.** In the heap-buffer case it's not a C-level leak: `imemo_tmpbuf` is a normal GC-managed Ruby object, so once `scratch_holder`'s stack slot is gone, it's reclaimed on the next GC like any other garbage object. `ALLOCV_END` is an eager-release optimization, not a correctness requirement. Worth stating precisely, though the PR's own conservative framing ("disclosed rather than armored with rb_ensure") is a reasonable way to present it.

**Reentrancy** (a cast callback recursively freeing the same result mid-row) is a real category but pre-existing and unaffected by this change: `resultFreed` is checked once per `fetch_row_func` call in both old and new code, not per cell, and array-mode casting doesn't re-touch `wrapper->result` per cell the way hash-mode's per-cell field-name lookups do.

Full suite: 516 examples, 0 failures.

No concerns. This looks correct and the safety analysis in the PR description holds up under independent verification.

Issue #1512, opened 2026-08-13
Proposal: Result#query_time — server round-trip measured in C with a monotonic clock


jeremy:
Part of the roadmap in #1510.
**Category: low-hanging fruit** · `effort:S · risk:low · ready:now · layer:C`

## Problem

There is no timing anywhere in ext/ or lib/. Every consumer that wants query latency wraps `Client#query` in Ruby-level `Process.clock_gettime` pairs (ActiveRecord's log does exactly this), which measures round trip **plus** Ruby row materialization **plus** GVL contention — three different signals blended into one number, at the cost of two Ruby calls per query. There is no way to answer "how slow was the *server*" from the outside.

## Proposal

Two `clock_gettime(CLOCK_MONOTONIC)` stamps in C:

- open when the command is written (`client.c:1128`, stamped on the wrapper),
- close when the first response has been fully read — i.e. after the read_query_result phase completes, **including** the socket wait in `do_query`'s `rb_wait_for_single_fd` loop (`client.c:996-1010`). Bracketing only `nogvl_read_query_result` would miss the server wait entirely, which is most of the signal.

Expose as `Result#query_time` (Float seconds; `nil` when unavailable). Because mysql2's store phase is a separate call (`nogvl_store_result`, `client.c:907`), we can optionally report it as a second reading — a split trilogy's design can't produce.

Monotonic means immune to NTP steps. `gettimeofday` fallback for platforms without `CLOCK_MONOTONIC`, behind a `have_func` probe.

## Evidence and precedent

trilogy has carried exactly this since 2.x: two monotonic stamps bracketing `trilogy_query_recv`, deliberately excluding column/row reading and object construction, surfaced as `Trilogy::Result#query_time` (`cext.c:883-897, :1035`).

## Sizing and risks

- **Cost**: two vDSO `clock_gettime` calls per query — ~20ns. No allocation.
- **Async split**: `Client#query` in async mode returns before the response (`client.c:1131` returns Qnil; the result arrives via `async_result`) — the stamp lives on the wrapper, so the bracket closes wherever read_query_result actually completes. The semantics ("send → first response fully read") hold on both paths and get documented.
- **Prepared statements**: same bracket in `statement.c` execute (`statement.c:573`), same field on the result wrapper.
- **Not a behavior change** anywhere; pure addition.

## Touch points

- `client.c`: stamp at send in `rb_mysql_query`, close in the read path shared with `async_result`.
- `statement.c:573`: same for execute.
[...]

PR #1553, opened 2026-08-13
Result#query_time: time the server round trip in C on a monotonic clock

jeremy:
Proposed in #1512.

There is no timing anywhere in ext/ or lib/. Every consumer that wants query latency wraps Client#query in Ruby-level Process.clock_gettime pairs (Active Record's log does exactly this), which blends three signals -- server round trip, Ruby row materialization, and GVL contention -- into one number, at the cost of two Ruby calls per query. There is no way to ask "how slow was the *server*" from the outside. trilogy has carried exactly this since 2.x: two monotonic stamps bracketing trilogy_query_recv, with its socket waits inside the bracket and column/row reading outside, surfaced as Trilogy::Result#query_time.

Two clock_gettime(CLOCK_MONOTONIC) stamps bracket the round trip:

* Open when the command is written to the wire. For the text protocol the stamp lives on the client wrapper (client.c, rb_mysql_query), because on the async path the bracket closes in a separate Ruby call: Client#query with :async returns right after the send, and #async_result reads the response later.
* Close when the first response has been fully read: right after nogvl_read_query_result in async_result (client.c) -- downstream of the rb_wait_for_single_fd loop in do_query, so the socket wait where a slow query spends its time is inside the bracket. For prepared statements both ends sit in rb_mysql_stmt_execute (statement.c) around nogvl_stmt_execute, since mysql_stmt_execute performs both the write and the first read within one call; the cursor-open handshake of a streaming execute falls inside the same bracket.

The reading rides into the Result as a new rb_mysql_result_to_obj parameter (result.c/result.h) -- a plain double assignment, so nothing new can raise inside to_obj (the streaming lifecycle depends on that) -- and comes out as Result#query_time, a Float in seconds. Server execution and network time are in; the store phase, casting into Ruby objects, and GVL waits after the first response are out. On the :async path the bracket closes inside #async_result, so time between the response becoming readable and that call is included. Closing at readability instead is not implementable: the gem runs no code between the send and the #async_result call (the caller monitors the socket themselves), and a stamp taken at #async_result entry would still sit after that dwell while excluding the read of the response itself. Client#store_result passes a negative sentinel: second and later result sets of a multi-statement command arrive after the bracket that measured the original command closed, so their #query_time is nil rather than a number that belongs to a different result set. #1512 floats an optional second reading for the store phase -- a split trilogy's design can't produce -- which stays a follow-up: this parcel carries the one bracket both drivers agree on.

clock_gettime(CLOCK_MONOTONIC) sits behind a have_func probe (extconf.rb) with a gettimeofday fallback for platforms without it (on Windows, gettimeofday comes from ruby/win32.h). Monotonic means immune to wall-clock steps; on the fallback path a backwards step could yield a negative reading, which the reader reports as nil rather than a negative Float.

Cost: two vDSO clock_gettime calls per query, no allocation. Purely additive -- no existing behavior moves.

Specs: text protocol (SELECT SLEEP(0.1): Float, >= the server sleep, <= a Ruby-level wall bracket around the whole call), async (>= the sleep; non-Windows only, where :async is real), streaming (reading available before any row is read), prepared statements (same three assertions as text), prepared streaming (reading available before the cursor's rows are fetched), and multi-statement (first result carries a Float, store_result nil). Verified they bite: re-stamping the open right before nogvl_read_query_result -- excluding the socket wait -- fails the text and streaming specs; passing the sentinel from statement.c fails the statement specs; passing it from only the streaming execute branch fails exactly the prepared-streaming spec; passing a number from store_result fails the nil spec; a reader hardwired to nil fails everything else, including async. The gettimeofday fallback is not exercised in CI -- every CI platform has CLOCK_MONOTONIC -- but was exercised locally: an #undef at the guard compiles the fallback branch and passes all six specs. (Disabling the extconf probe alone doesn't reach it: ruby/config.h defines HAVE_CLOCK_GETTIME on its own wherever Ruby's build found the function, which also means the probe can only ever widen coverage, not lose it.)

Full suite: 508 examples, 0 failures, 10 pending (SSL-gated), RuboCop clean.

### DIFF
--- a/README.md
+++ b/README.md
@@ -715,6 +715,19 @@ NOTE: Because of the way MySQL's query API works, this method will block until t
 So if you really need things to stay async, it's best to just monitor the socket with something like EventMachine.
 If you need multiple query concurrency take a look at using a connection pool.
 
+### Query timing
+
+Every `Mysql2::Result` carries the server round trip that produced it, measured in C on a monotonic clock:
+
+``` ruby
+result = client.query("SELECT * FROM table")
+result.query_time # => 0.000542 (Float seconds)
+```
+
+The bracket opens when the command is written to the connection and closes when its first response has been fully read. Server execution and network time to the first response are in; buffering the remaining rows, casting values into Ruby objects, and GVL waits after the first response are out -- so it answers "how slow was the server?" without the noise a Ruby-level stopwatch around `#query` picks up. The reading is the round trip as observed by the calling thread: on a quiet process it matches the wire, while under heavy GVL contention it includes time the thread spent waiting to be rescheduled mid-round-trip, like any thread-observed timing in CRuby.
+
+For a query issued with `:async => true` the bracket closes inside `#async_result`, so time between the response becoming readable and that call is included. `#query_time` is `nil` when no reading applies: the second and later result sets of a multi-statement command, retrieved via `#store_result`.
+
 ### Row Caching
 
 By default, Mysql2 will cache rows that have been created in Ruby (since this happens lazily).
--- a/ext/mysql2/client.c
+++ b/ext/mysql2/client.c
@@ -6,6 +6,7 @@
 #ifndef _WIN32
 #include <sys/types.h>
 #include <sys/socket.h>
+#include <sys/time.h>
 #endif
 #ifndef _MSC_VER
 #include <unistd.h>
@@ -598,6 +599,7 @@ static VALUE allocate(VALUE klass) {
   wrapper->closed = 1; /* will be set false after calling mysql_real_connect */
   wrapper->refcount = 1;
   wrapper->affected_rows = -1;
+  wrapper->query_start = 0;
   wrapper->client = (MYSQL*)xmalloc(sizeof(MYSQL));
   wrapper->state = MYSQL2_CLIENT_IDLE;
   wrapper->pending_stmt_closes = NULL;
@@ -841,6 +843,16 @@ static VALUE rb_mysql_client_closed(VALUE self) {
   return CONNECTED(wrapper) ? Qfalse : Qtrue;
 }
 
+double mysql2_monotonic_now(void) {
+#if defined(HAVE_CLOCK_GETTIME) && defined(CLOCK_MONOTONIC)
+  struct timespec ts;
+  return clock_gettime(CLOCK_MONOTONIC, &ts) == 0 ? (double)ts.tv_sec + (double)ts.tv_nsec / 1e9 : -1;
+#else
+  struct timeval tv;
+  return gettimeofday(&tv, NULL) == 0 ? (double)tv.tv_sec + (double)tv.tv_usec / 1e6 : -1;
+#endif
+}
+
 /*
  * mysql_send_query is unlikely to block since most queries are small
  * enough to fit in a socket buffer, but sometimes large UPDATE and
@@ -868,14 +880,24 @@ static VALUE do_send_query(VALUE args) {
   return Qnil;
 }
 
+struct nogvl_read_query_result_args {
+  MYSQL *mysql;
+  /* Round-trip close stamp, taken while the GVL is still released: under
+   * contention, reacquisition means queueing behind every runnable thread,
+   * and a stamp taken after it would charge that wait to the server. */
+  double query_end;
+};
+
 /*
  * even though we did rb_thread_select before calling this, a large
  * response can overflow the socket buffers and cause us to eventually
  * block while calling mysql_read_query_result
  */
 static void *nogvl_read_query_result(void *ptr) {
-  MYSQL * client = ptr;
-  my_bool res = mysql_read_query_result(client);
+  struct nogvl_read_query_result_args *args = ptr;
+  my_bool res = mysql_read_query_result(args->mysql);
+
+  args->query_end = mysql2_monotonic_now();
 
   return (void *)(res == 0 ? Qtrue : Qfalse);
 }
@@ -909,8 +931,9 @@ static void *nogvl_use_result(void *ptr) {
 /* Shared by async_result (a query's first result set) and store_result (a
  * later one, from a multi-statement batch): fetch the current result set as
  * either streamed or fully-buffered, per :stream, track it on wrapper, and
- * wrap it as a Result. */
-static VALUE mysql2_fetch_result_set(VALUE self, mysql_client_wrapper *wrapper) {
+ * wrap it as a Result carrying query_elapsed as its #query_time (negative
+ * for none). */
+static VALUE mysql2_fetch_result_set(VALUE self, mysql_client_wrapper *wrapper, double query_elapsed) {
   MYSQL_RES *result;
   VALUE resultObj, current, is_streaming;
 
@@ -948,7 +971,7 @@ static VALUE mysql2_fetch_result_set(VALUE self, mysql_client_wrapper *wrapper) 
   current = rb_hash_dup(rb_ivar_get(self, intern_current_query_options));
   (void)RB_GC_GUARD(current);
   Check_Type(current, T_HASH);
-  resultObj = rb_mysql_result_to_obj(self, wrapper->encoding, current, result, Qnil);
+  resultObj = rb_mysql_result_to_obj(self, wrapper->encoding, current, result, Qnil, query_elapsed);
 
   /* Track the open cursor so a later command can force-drain it if it's
    * abandoned instead of exhausted -- see mysql2_abandon_active_stream. */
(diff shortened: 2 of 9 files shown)


sodabrew:
Ready for rebase after #1503 moved a few things around

jeremy:
Rebased on master. #1503's shared `mysql2_fetch_result_set` now carries the reading as a parameter: `async_result` closes the bracket right after `nogvl_read_query_result` and passes the elapsed reading, `store_result` passes the negative sentinel so later result sets of a multi-statement batch stay nil — same bracket semantics as before, one call site for `rb_mysql_result_to_obj` instead of two. Full suite: 541 examples, 0 failures, 11 pending (SSL-gated); RuboCop clean.

nuclearspike (commented):
Reviewed and verified empirically, not just read through — macOS 26.5.2 arm64, Ruby 3.3.10, MySQL 9.6.0 via socket, fresh-extconf build of this head: suite 541 examples / 0 failures / 11 pending (environment skips), RuboCop clean, no compiler warnings.

Every path I probed on this build held up:

| claim | measured |
|---|---|
| bracket spans send through first response | `SLEEP(0.5)` → `query_time` 0.5695s, wall 0.5696s |
| store/materialization excluded | 32.8k×20 result: `query_time` 8.3ms vs 194ms wall-with-`each` |
| no stale bracket across commands | 0.2ms after a 0.4s query; clean after an errored query |
| multi-statement | first result Float ≥ sleep; `store_result` result nil |
| async dwell included (as documented) | 0.1s sleep + 0.3s dwell → 0.305s |
| prepared + prepared-streaming | Float ≥ sleep; reading available before rows |
| bracket honesty when quiet | wall − `query_time` = 0.1ms across 7 runs |
| GVL contention (8 spinning threads) | `SLEEP(0.05)` → 2.63s reported; 1.75s with the close stamp moved into the no-GVL read on a scratch build |

Two things I'd ask for before merge, both about the contract rather than the mechanism:

**1. Under GVL contention the bracket reports scheduler wait as server time — and the wording promises otherwise.** With 8 Ruby threads spinning, `SELECT SLEEP(0.05)` reports `query_time` ≈ 2.63s — 52× the requested sleep interval. The close stamp runs after `rb_thread_call_without_gvl` reacquires the GVL, and the wait call's return queues behind the spinners too. I tried the obvious repair on a scratch build — taking the close stamp inside `nogvl_read_query_result`, before reacquisition — and in this run it removed about a third (2.63s → 1.75s). The text path crosses the GVL three times inside the bracket (reacquiring after the send, returning from the single `rb_wait_for_single_fd`, and reacquiring after the read); with N spinning threads each crossing queues behind all of them, and the inflation scales linearly with N (≈200ms/700ms/1.4s/2.8s at 0/2/4/8 burners here) — arithmetic consistent with that shape, though I didn't instrument the per-crossing split. An out-of-process client run during the 8-thread spin completed in 0.225s including process startup, so none of this is server-side. The interruptible-wait design causing it is worth keeping regardless. So: the in-no-GVL close is cheap and real (the prepared path's close after `nogvl_stmt_execute` would want the same treatment; the clock call needs no GVL — return status out and handle failure after reacquiring, no Ruby APIs inside), but the durable fix is mostly wording — this is the round trip *as observed by the calling thread*, and under contention that includes scheduler wait, like every thread-observed timing in CRuby. Quiet, it is exact: wall − `query_time` = 0.1ms.

**2. The trilogy-parity claim doesn't hold for multi-statement, and the nil is a divergence rather than an agreement.** Measured with trilogy 2.12.6 against the same server: for `SELECT SLEEP(0.2); SELECT 2`, trilogy reports `query_time` 0.347 on the first result and 0.0001 on the second — later result sets get a real per-result number, not nil. (Source-wise it's `rb_trilogy_next_result` → `execute_read_query_response` → `read_query_response` at `c25b18e`, bracketing from the `next_result` call to that response — so it can omit server work done before the call, a different contract rather than an obviously better one; trilogy also opens after the send, not before it.) So "the one bracket both drivers agree on" undersells your options: either time the `next_result` read the same way (that call reads/consumes the next response — the comment's "arrived after the bracket closed" isn't quite it), or keep the nil and adjust the claim. Either is defensible; the current text isn't.

Smaller, take or leave:

- README/docstring say "covers server execution and network time"; for a multi-packet result the store phase — genuine network transfer — is outside the bracket, so something like "send to first-response read" is the precise phrase ("time to first response" would overclaim the async path, where the read — and so the close — happens at `#async_result`, after any dwell). (The exclusion itself is the right design and the description already owns it.)
- `mysql2_monotonic_now` ignores both clocks' return values; on a hypothetical failure it reads an uninitialized struct. trilogy checks these. Cheap to return the negative sentinel instead.
- The new specs call `Process.clock_gettime` directly where `spec_helper`'s `clock_time` exists; the suite has precedent for both, so purely a consistency nit.

Explicitly not requesting: any change to the async-dwell behavior — the description's impossibility argument is correct (no gem code runs between the send and `#async_result`), and the disclosure is honest.

The feature itself is exactly the primitive the bench-pipeline discussion needs, and quiet-side it precisely measures its send-to-first-response-read interval. With the contention wording settled and the trilogy claim corrected, happy to see this land.

sodabrew (approved):
LGTM! Good findings by @nuclearspike -- @jeremy lmk if I should hold merge if you plan to update for either concern?

jeremy:
@nuclearspike A review that arrives with its own bench rig is worth a lot — both asks taken, at head d0b837a.

**1. Contention:** the close stamp now lands inside the no-GVL region on both paths — `nogvl_read_query_result` and `nogvl_stmt_execute` each take a small args struct and stamp `query_end` before returning, so the reacquisition queue you measured stays outside the bracket. Elapsed is computed after reacquiring, no Ruby APIs inside — the shape you sketched, prepared path included. And the wording now states the contract: the round trip *as observed by the calling thread* — quiet it matches the wire, under contention it includes time spent waiting to be rescheduled mid-round-trip, like any thread-observed timing in CRuby. README, the `#query_time` docstring, and the commit message all carry it, with the residual wait-loop wakeups named as the price of interruptibility.

**2. Trilogy claim:** conceded — "the one bracket both drivers agree on" was wrong and is gone. The nil stays; the `store_result` comment now says what actually happens (the response is read and consumed by `Client#next_result`, which no bracket surrounds) instead of "arrived after the bracket closed", and the commit message describes trilogy's actual contract: a per-result bracket around each `next_result` read, which can omit server work done before the call — different, not strictly better. Bracketing `next_result` the same way here is a clean follow-up parcel if the per-result numbers earn their keep; happy to write it alongside the #1512 store-phase reading.

Smaller items, all taken: `mysql2_monotonic_now` checks both clock calls and returns the negative sentinel on failure, with both ends of the bracket guarded so a failed stamp yields nil rather than garbage; README/docstring scope the claim to network time *to the first response*, store-phase transfer excluded as documented; the specs use `spec_helper`'s `clock_time`.

Full suite at the new head: 544 examples (the base grew with #1554/#1561 since the earlier 541 count), 0 failures, 11 pending (SSL-gated); RuboCop clean; clean compile, no warnings.

@sodabrew updated in place rather than stacked — d0b837a is the head to merge whenever you're ready.

nuclearspike:
Verified `d0b837a` on the same rig: quiet path unchanged (wall − `query_time` = 0.1ms), contended text path drops 2.63s → 1.74s — within a few ms of what the scratch-build predicted — and the prepared path under the same 8-thread contention reads 93–109ms, i.e. quiet-level: with the stamp inside `nogvl_stmt_execute` that bracket has no mid-flight GVL crossing left at all, which is a nicer property than the sketch aimed for. Suite reproduces at 544 / 0 here too.

One micro-nit in the new wording: "residual wait-loop wakeups" — `wait_for_readable_with_timeout` makes a single `rb_wait_for_single_fd` call, so there's no loop; the residue is the crossings themselves. I only bring it up because we (Claude and Paul) made exactly the same mistake drafting the review — it slipped past Codex, Claude, and DeepSeek, and it was Kimi that caught it.

+1 on the `next_result` bracket as a follow-up parcel alongside the #1512 store-phase reading — happy to bring the rig to that one too.

jeremy:
@nuclearspike The prepared-path number is the best line in the report: 93–109ms under 8 spinning threads means that bracket now has zero mid-flight GVL crossings — open before the release, close inside `nogvl_stmt_execute`, nothing in between that can queue. The text path can't reach that (its interruptible wait is the point), so 2.63s → 1.74s is the floor the design allows, and you measured both landing where the scratch build predicted.

Nit taken, and it's correct twice over: `do_query` makes exactly one `rb_wait_for_single_fd` call, and my comment above repeated the same phantom loop. The word had a plausible neighbor — `next_result_nonblocking` really does loop around `wait_for_readable_with_timeout`, and libmysqlclient's blocking fallback retries internally — but neither is the bracket's wait. The phrase lived only in the commit message (README and the docstring carry the contract without it), so the fix is a message-only amend, now at head 3a4d884: "loop" is gone and the residue is named as what it is — the crossings themselves, reacquiring after the send and returning from the single wait.

@sodabrew 3a4d884 is d0b837a with only the commit message changed — the trees are identical (`git rev-parse HEAD^{tree}`: 53bddb5 on both), so nothing your approval covered moved. Merge whenever ready.

+1 noted for the `next_result` bracket parcel — the rig is welcome there too.

Issue #1530, opened 2026-08-13
Proposal: warn at connect time on incoherent or silently-ignored options


jeremy:
Part of the roadmap in #1510.
**Category: low-hanging fruit** · `effort:S · risk:low · ready:now · layer:lib`

Addresses #879 and #936 (`:ssl_mode` silently ignored on MariaDB Connector/C builds).

## Problem

`Client#initialize` massages its options silently, and the code knows it: `lib/mysql2/client.rb:35` carries a maintainer TODO — "stricter validation rather than silent massaging". Most of the massaging is harmless. One case is not:

**On MariaDB Connector/C builds, `:ssl_mode` can be accepted and then not enforced.** #879 reported it in 2019; #936 discussed the warning; the warning never shipped. A user who writes `ssl_mode: :required` and gets a cleartext connection has suffered a silent security downgrade — the worst kind of option incoherence, because the failure is invisible until someone captures traffic. The C side already does careful version discrimination for ssl_mode (`client.c:124-198`, warning on several known-broken version ranges), but the MariaDB-flavor gap identified in #879 fell through it.

Beyond the flagship case there are smaller incoherences nobody states: `stream: true` with `cache_rows: true` is silently coerced, unknown option keys pass through unremarked, and future options (a callable password, for instance) will interact with `:reconnect`.

## Proposal

Warnings-only pairwise coherence checks at connect, in `Client#initialize` / `check_and_clean_query_options` (`lib/mysql2/client.rb:22`, `:177`):

1. **The #879/#936 warning, finally**: when `:ssl_mode` is set on a build where the underlying library won't enforce it, say so at connect time. This likely needs a C-exported client-library-flavor constant so lib/ can distinguish MariaDB Connector/C from libmysqlclient — the discriminator already exists at compile time (`MARIADB_PACKAGE_VERSION`, `client.c:55`); exporting it is a few lines.
2. A small table of pairwise checks alongside it: `stream: true` + `cache_rows: true` (explain the coercion instead of performing it silently), unknown-key notice, and room to grow one check per future incoherence rather than building generalized capability machinery up front.

Warnings, not raises: existing configurations keep working, and the precedent is already established in the same file — `parse_ssl_mode` warns on an unknown mode (`lib/mysql2/client.rb:106`) and `parse_flags_array` warns on unknown flags (`:117`). This proposal extends a pattern the file already has to the cases where silence costs the most.

[...]

PR #1555, opened 2026-08-13
Warn at connect time on incoherent or silently-ignored option pairs

jeremy:
Proposed in #1530.

Client#initialize massages its options silently, and the file's own TODO (lib/mysql2/client.rb:35) has said "stricter validation rather than silent massaging" for years. This ships the connect-time warning layer proposed there: two checks, each with its evidence cited in a comment beside it, warn-and-proceed only -- no connection behavior changes. They live in a private warn_incoherent_options called from check_and_clean_query_options, the file's existing warn-at-connect spot (deprecated key names already warn there), so each fires at most once per client, before connect:

* :sslkey and :sslcert given singly. The pair only takes effect together; alone, libmysqlclient sends no client certificate (MySQL Reference Manual, "Configuring MySQL to Use Encrypted Connections") while MariaDB Connector/C refuses the connection outright with a TLS error that never names the missing half (verified against Connector/C 3.4.9: "TLS/SSL error: No such file or directory" for a bogus lone key, "TLS/SSL error: no start line" for a real one). The warning precedes connect on both flavors, naming the silence on one and decoding the error on the other. nil-presence is the test (not key?), so an explicit sslkey: nil beside a real sslcert warns too, matching what ssl_set receives.

* Client-wide stream: true with :cache_rows left at its true default. Every query on such a client warns and coerces at materialization time (ext/mysql2/result.c:1828); the connect-time warning names the fix (cache_rows: false) once at the source instead. Per-query query(sql, stream: true) is untouched -- the check reads the connection-level defaults only.

Deliberately not included, per #1530's own sizing: no unknown-option-key notice (unbounded noise surface across adapters and forks), no Client.capabilities roster, no reconnect-vs-callable-password check (no callable passwords exist yet). A write_timeout/streaming pairing was considered and dropped: no issue or PR substantiates an incoherence there.

Also dropped, on review (see below): the :ssl_mode-on-MariaDB warning this PR originally led with, citing the #879/#936 enforcement gaps. #1506 closed those gaps in the extension itself -- verify_ca and verify_identity both map to MYSQL_OPT_SSL_VERIFY_SERVER_CERT on the MariaDB code path, enforced, no :sslverify required -- so the warning would have misfired on every correctly configured verification-mode MariaDB connection and prescribed a redundant option. The CLIENT_LIB_FLAVOR export went with it: its only consumer was that warning.

Behavior delta, disclosed: two new stderr warnings for configurations that were previously silent. Test suites that promote warnings to errors will surface them -- a one-line fix (cache_rows: false, or dropping the lone key/cert) in each case. Also disclosed: Metrics/ClassLength in .rubocop_todo.yml bumps 147 -> 153 for the new method, by hand as in #1482 (a full --regenerate-todo rewrites unrelated entries from RuboCop version drift); the cap stays sized to the class's current line count exactly, as before.

On coverage: the sslkey/sslcert specs tolerate a connect abort (rescue inside the expectation) because the warning fires before connect and the flavors then diverge as described above. Verified the specs bite: with the lib change reverted to master's, the three positive warning specs fail; restored, all 5 in the new context pass.

Full suite on Ruby 4.0.6, rebased on current master (includes #1506): libmysqlclient 8.0.46 (Homebrew) / MySQL 9.6 over a local socket, macOS arm64: 540 examples, 0 failures, 11 pending (SSL-gated), RuboCop clean.

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -39,7 +39,7 @@ Metrics/BlockNesting:
 # Offense count: 1
 # Configuration parameters: CountComments, CountAsOne.
 Metrics/ClassLength:
-  Max: 147
+  Max: 153
 
 # Offense count: 3
 # Configuration parameters: AllowedMethods, AllowedPatterns.
--- a/lib/mysql2/client.rb
+++ b/lib/mysql2/client.rb
@@ -174,7 +174,28 @@ module Mysql2
 
     private
 
+    # Warns once per client, before connecting, about option combinations
+    # that are incoherent or silently ignored. Warnings only: the connection
+    # proceeds exactly as it would have without them.
+    def warn_incoherent_options
+      # The client key and certificate only take effect together. Given one
+      # without the other, libmysqlclient silently sends no client certificate
+      # and MariaDB Connector/C aborts the connection with a bare TLS error
+      # that never names the real problem.
+      # https://dev.mysql.com/doc/refman/en/using-encrypted-connections.html
+      warn ":sslkey and :sslcert only take effect together; alone, libmysqlclient sends no client certificate and MariaDB Connector/C fails to connect" \
+        if @query_options[:sslkey].nil? != @query_options[:sslcert].nil?
+
+      # Streaming results are never cached, so a client-wide :stream default
+      # overrides the :cache_rows default on every query (see the per-query
+      # warning in ext/mysql2/result.c).
+      warn ":cache_rows is ignored on a client with :stream enabled; pass cache_rows: false to acknowledge streaming semantics" \
+        if @query_options[:stream] && @query_options[:cache_rows]
+    end
+
     def check_and_clean_query_options
+      warn_incoherent_options
+
       if %i[user pass hostname dbname db sock].any? { |k| @query_options.key?(k) }
         warn "============= WARNING FROM mysql2 ============="
         warn "The options :user, :pass, :hostname, :dbname, :db, and :sock are deprecated and will be removed at some point in the future."
--- a/spec/mysql2/client_spec.rb
+++ b/spec/mysql2/client_spec.rb
@@ -247,6 +247,46 @@ RSpec.describe Mysql2::Client do # rubocop:disable Metrics/BlockLength
     end
   end
 
+  context "option coherence warnings" do
+    it "warns when :sslkey is given without :sslcert" do
+      expect do
+        begin
+          new_client(sslkey: '/path/to/client-key.pem')
+        rescue Mysql2::Error
+          # MariaDB Connector/C rejects a lone key at connect; the warning precedes it.
+        end
+      end.to output(/:sslkey and :sslcert only take effect together/).to_stderr
+    end
+
+    it "warns when :sslcert is given without :sslkey" do
+      expect do
+        begin
+          new_client(sslcert: '/path/to/client-cert.pem')
+        rescue Mysql2::Error
+          # MariaDB Connector/C rejects a lone certificate at connect; the warning precedes it.
+        end
+      end.to output(/:sslkey and :sslcert only take effect together/).to_stderr
+    end
+
+    it "warns when :stream is enabled with :cache_rows left on" do
+      expect do
+        new_client(stream: true)
+      end.to output(/:cache_rows is ignored on a client with :stream enabled/).to_stderr
+    end
+
+    it "does not warn when :stream is enabled with :cache_rows disabled" do
+      expect do
+        new_client(stream: true, cache_rows: false)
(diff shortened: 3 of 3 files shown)


sodabrew:
Reviewed this alongside #1506 (landing now), since this PR's `:ssl_mode` warning is explicitly framed as complementary to it. That interaction is worth a closer look before this merges.

## The `:ssl_mode` warning's premise doesn't hold up against #1506's actual fix

This warning fires whenever `:ssl_mode` requests verification and `:sslverify` isn't *also* set:

```ruby
warn "...pair it with :sslca and :sslverify to require verification" \
  if CLIENT_LIB_FLAVOR == :mariadb && @query_options[:ssl_mode].to_s.upcase.include?("VERIFY") && !@query_options[:sslverify]
```

Its justifying comment cites #879 and #936 as evidence that MariaDB "accepts the verification modes without reliably enforcing them." **#936 is the exact bug #1506 fixes.**

`:sslverify` and `:ssl_mode` are two independent mechanisms in this codebase: `:sslverify` sets the legacy `CLIENT_SSL_VERIFY_SERVER_CERT` connection flag (`lib/mysql2/client.rb:73`), while `:ssl_mode`'s handling in `rb_set_ssl_mode_option` (what #1506 patches) calls the newer `mysql_options(..., MYSQL_OPT_SSL_VERIFY_SERVER_CERT, ...)` API directly and unconditionally, with no dependency on whether `:sslverify` was also passed. #1506's own PR body demonstrates this working correctly against a real MariaDB 11.8 server: `ssl_mode: :verify_ca` alone, no `:sslverify`, wrong CA rejected.

So once #1506 lands, this warning will fire, and recommend adding a `:sslverify` that #1506 already made unnecessary, on every correctly-configured `verify_ca`/`verify_identity` connection on MariaDB. The PR description asserts the opposite of this directly: "the MariaDB flavor warning stays accurate regardless, since the enforcement gap there is library-side, not fixable from Ruby." Based on #1506's diff, that's not accurate: the gap #936 described was fixable from Ruby, and #1506 fixes it there.

I can't fully rule out a version-range nuance neither PR's diff discusses (e.g. an older MariaDB Connector/C where the OPT-based call doesn't work and only the flag-based one does), which would justify keeping this warning as-is. But nothing in either PR currently establishes that; the warning's own justification comment just cites the bug #1506 fixes. Worth reconciling with the same kind of concrete against-a-real-server evidence #1506 provided, since as written this would ship a stale warning right alongside the fix that makes it inaccurate.

## Everything else checked out

- `CLIENT_LIB_FLAVOR`'s `defined(MARIADB_PACKAGE_VERSION)` check is identical to the discriminator already used for `MYSQL_LINK_VERSION` at `client.c:55` -- genuinely reusing an established, trusted signal, not a new detection mechanism.
- The `:sslkey`/`:sslcert` pairing check (nil-presence XOR) is correct and precise.
- The `:stream`/`:cache_rows` check is correctly scoped to client-wide defaults only (confirmed `cache_rows: true` is in fact the default, so this fires on exactly the case it claims to), and is consistent with the existing per-query runtime warning in `result.c`.
- Full suite passes locally (165 examples in `client_spec.rb`, 0 failures), RuboCop clean.

jeremy:
You're right, and #1506 landing settles it. The warning's own citations were its undoing: #936 is the bug #1506 fixes, and #879's `verify_identity` was already wired to `MYSQL_OPT_SSL_VERIFY_SERVER_CERT` on that path. With both cited gaps closed in the extension itself -- unconditionally, with no dependence on `:sslverify` -- my "library-side, not fixable from Ruby" claim was simply wrong, and the warning would have fired on every correctly configured `verify_ca`/`verify_identity` connection on MariaDB while prescribing a redundant option.

On the version-range nuance: I found nothing establishing an older Connector/C where the OPT-based call fails but the flag-based one works, and I'm not going to keep a warning alive on an undemonstrated gap. If one ever surfaces, it should come back version-gated on its own evidence.

Dropped in bbd40b7:

- The `:ssl_mode` warning and its five specs.
- `CLIENT_LIB_FLAVOR` -- its only consumer was that warning, so it goes too rather than shipping as unconsumed API. (It may yet earn its way back: `rb_set_ssl_mode_option`'s version-range sniff classifies MySQL's year-based releases -- libmysqlclient 26.7.0 reports a version id >= 260000 -- into its MariaDB branch, and the compile-time discriminator is the clean fix. Separate PR if so.)

Kept the two checks you verified, unchanged. Rebased onto current master (includes #1506), commit message and PR body rewritten to match. Re-verified: full suite 540 examples, 0 failures, 11 pending, RuboCop clean; with lib/mysql2/client.rb reverted to master's the three positive warning specs still fail, so the remaining coverage bites.

Issue #1522, opened 2026-08-13
Proposal: finish the fiber-claim audit — guard every command-issuing method


jeremy:
Part of the roadmap in #1510.
**Category: gap filler** · `effort:M · risk:med · ready:after-train · layer:C`

Finishes the audit byroot flagged in the #1433 thread ("and perhaps others?") and fixes the dead-fiber permanent claim in #1392.

## Problem

The `active_fiber` claim — the check that raises "This connection is in use by ..." instead of letting two fibers interleave on one `MYSQL*` — is taken by exactly three methods: `query` (`client.c:1107`), `ping` (`client.c:1485`), and `close` (`client.c:773`). Every other method that issues a command on the shared connection is unguarded:

- `select_db` — releases the GVL at `client.c:1463`, no claim
- `abandon_results!` — GVL-held `mysql_next_result` + GVL-released `nogvl_store_result` in a loop (`client.c:1060-1070`), no claim
- `store_result` — GVL-released `nogvl_store_result` at `client.c:1572`, no claim
- `next_result` — `mysql_next_result` reads the wire at `client.c:1547`, no claim
- `set_server_option` — writes `COM_SET_OPTION` at `client.c:1516`, no claim

A fiber calling `next_result` while another fiber's `query` is in flight interleaves two commands on one protocol stream. The failure is not an error — it is desynced packets and corrupted results inside the client library, with nothing raised at the point of misuse.

Separately, #1392: the claim can be poisoned permanently. If the claiming thread dies (`Thread#exit` mid-query), `active_fiber` stays set and every subsequent call raises forever. Only `close` recovers, via its takeover path — `closing && !rb_fiber_alive_p(wrapper->active_fiber)` at `client.c:591-599`.

## Proposal

Three parts, all extending the existing mechanism rather than adding a new one:

1. **Extend the claim** to the five entry points above — the same `rb_mysql_client_set_active_fiber(self, false)` call `query` makes, placed after any raise-capable setup.
2. **Generalize the dead-fiber takeover**: allow a new claimant when the holding fiber is no longer alive, not only when closing. The shape already exists at `client.c:591-599`; this lifts the `closing` restriction and fixes #1392.
[...]

PR #1563, opened 2026-08-14
Extend the fiber-claim guard to every command-issuing method

jeremy:
Proposed in #1522.

The active_fiber claim -- the GVL-atomic check-and-set that raises "This connection is in use by ..." instead of letting two fibers interleave commands on one MYSQL* -- was taken by exactly three methods: query, ping, and close. Every other method that issues a command on the shared connection ran unguarded: select_db, abandon_results!, store_result, next_result, and set_server_option. A fiber calling any of them while another fiber's query was in flight interleaved two commands on one protocol stream, and the failure was not an error: desynced packets and corrupted results inside the client library, with nothing raised at the point of misuse.

Each of the five now takes the same claim query takes (rb_mysql_client_set_active_fiber -- client.c:1194, 1605, 1698, 1843, 1860), placed after all raise-capable setup so a failed argument conversion can't leak it. Every guard honors the no-raise-between-claim-and-clear rule: any unwind between taking the claim and releasing it -- an ordinary raise, a Timeout interrupt, or a Thread#exit -- releases the claim, reusing the completion-flag + rb_ensure shape #1508 established for query's own send/read phases.

Mechanism per method:

- set_server_option (client.c:1688): claim, run the COM_SET_OPTION round trip (GVL-held, so no interrupt can land mid-guard), release. The raise-capable NUM2INT happens before the claim.
- select_db (client.c:1595): the wire call runs under rb_ensure with a shared companion, release_claim_or_disconnect (client.c:1130): a body that reached its completion mark just releases the claim; one that didn't was interrupted mid-exchange, so the companion also invalidates the socket, same as an interrupted query. A server error reply ("Unknown database") marks completion before raising -- the round trip finished, so the connection stays usable, which the existing select_db specs (a successful select_db right after a failed one) pin.
- next_result (client.c:1833): claim before the existing rb_ensure; the reset-state companion (client.c:1755) now chains to release_claim_or_disconnect. A failed statement in the batch (NET_ASYNC_ERROR / ret > 0) is a complete server reply: completion is marked before the raise, so the connection survives -- pinned by the existing "raise an exception when one of multiple statements fails" spec, which calls next_result again after the raise.
- store_result (client.c:1854): claim on entry; release follows the exact lifecycle a query's own result fetch already has -- nogvl_do_result clears the claim once the result is stored, and the fetch-error path in mysql2_fetch_result_set clears it before raising.
- abandon_results! (client.c:1185): claims once and holds across the whole drain loop -- a new nogvl_store_result_raw (client.c:913) stores without the claim-clearing side effect of nogvl_store_result, so the guard isn't porous between result sets. A failed statement mid-batch marks completion before raising, leaving the connection usable (new spec).

Behavior changes, both changelog-visible:

- Concurrent misuse that previously corrupted silently now raises Mysql2::Error from all five methods, matching what query/ping/close already did. Code relying on (already unsafe) concurrent select_db/next_result/etc. will see new exceptions; same-fiber misuse (e.g. abandon_results! while an async query is outstanding) raises "This connection is still waiting for a result".
- An interrupted next_result / select_db / abandon_results! (Timeout, Thread#raise, or Thread#exit landing mid-exchange) now invalidates the connection, exactly as an interrupted query read always has -- previously the connection was left open mid-protocol, where the next command silently desynced. The new "does not leave the connection claimed after an interrupted #next_result" spec pins the no-lockout half: after the interrupt, the next caller gets a normal connection error, never a permanent "in use by"/"still waiting" claim.

Verification: each new spec fails against a deliberately broken build -- one break per guard (claim removed: the matching "should not allow concurrent use of #X" spec fails), one with abandon_results!'s completion mark dropped before its raise (the failed-statement spec fails because the companion then invalidates a healthy connection), and one with the companion's not-completed branch disabled (the interrupted-next_result spec fails with a leaked same-fiber claim -- which is why that spec matches both claim messages, not just "in use by"). Full suite green: 545 examples, 0 failures (11 pre-existing pendings); rubocop clean. The blocking-mysql_next_result fallback path (MariaDB / MySQL < 8.0.16) was force-compiled locally with HAVE_MYSQL_NEXT_RESULT_NONBLOCKING off and passes the same specs.

Bench (Apple Silicon, localhost server, 3 trials each): 3000 drains of a 3-statement batch via next_result/store_result -- master 0.115-0.136s, guarded 0.108-0.164s; 1000 select_db round trips -- 0.012-0.017s on both builds. Flat within noise, as expected: each guard adds one rb_fiber_current + compare per method call, nothing per-row.

Coverage limits: the closing-only dead-fiber takeover rule in rb_mysql_client_set_active_fiber is untouched -- #1508's ensure-based cleanup already covers the Thread#exit lockout that this issue's takeover generalization originally targeted. Statement#execute's send/read path remains unguarded, as noted in #1508, and is out of scope here. The Windows branch of release_claim_or_disconnect mirrors disconnect_and_mark_inactive's existing socket-close idiom and is compile-verified by CI only. Pre-existing and untouched: calling store_result on a client that has never issued a query crashes (a hash lookup on the unset query-options ivar) on master exactly as before, and deserves a separate fix.

### DIFF
--- a/ext/mysql2/client.c
+++ b/ext/mysql2/client.c
@@ -136,6 +136,7 @@ struct nogvl_send_query_args {
 struct nogvl_select_db_args {
   MYSQL *mysql;
   char *db;
+  struct query_completion completion;
 };
 
 static VALUE rb_set_ssl_mode_option(VALUE self, VALUE setting) {
@@ -906,6 +907,14 @@ static void *nogvl_use_result(void *ptr) {
   return nogvl_do_result(ptr, 1);
 }
 
+/* Unlike nogvl_store_result above, leaves active_fiber alone: the drain loop
+ * in do_abandon_results holds its claim across every result set in the
+ * batch, not just up to the first stored one. */
+static void *nogvl_store_result_raw(void *ptr) {
+  mysql_client_wrapper *wrapper = ptr;
+  return mysql_store_result(wrapper->client);
+}
+
 /* Shared by async_result (a query's first result set) and store_result (a
  * later one, from a multi-statement batch): fetch the current result set as
  * either streamed or fully-buffered, per :stream, track it on wrapper, and
@@ -1109,6 +1118,37 @@ static VALUE disconnect_and_mark_inactive(VALUE self) {
   return Qnil;
 }
 
+/* rb_ensure companion for the fiber-claimed command methods that release the
+ * GVL mid-command (select_db, next_result, abandon_results!): releases the
+ * claim on every kind of unwind, so no raise -- or Thread#exit -- between
+ * claim and release can leave the connection permanently reporting "This
+ * connection is in use by". When the body never reached its completion mark,
+ * an interrupt landed mid-exchange and the connection may be stuck
+ * mid-protocol, so it also gets invalidated, same as an interrupted query.
+ * Bodies mark completion before raising a server-reported error: that reply
+ * finished the round trip, so the connection itself is still usable. */
+static VALUE release_claim_or_disconnect(VALUE completionval) {
+  struct query_completion *completion = (void *)completionval;
+  mysql_client_wrapper *wrapper = completion->wrapper;
+
+  if (completion->completed) {
+    wrapper->active_fiber = Qnil;
+  } else {
+#ifndef _WIN32
+    invalidate_after_interrupted_query(wrapper);
+#else
+    wrapper->active_fiber = Qnil;
+    wrapper->state = MYSQL2_CLIENT_IDLE;
+    if (CONNECTED(wrapper)) {
+      close(wrapper->client->net.fd);
+      wrapper->client->net.fd = -1;
+    }
+#endif
+  }
+
+  return Qnil;
+}
+
 /* call-seq:
  *    client.abandon_results!
  *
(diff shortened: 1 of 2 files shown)


spec/mysql2/client_spec.rb
```diff
@@ -1096,6 +1096,34 @@ def run_gc
...
+        end.to raise_error(Timeout::Error)
+
+        # The interrupted exchange invalidates the connection, same as an
+        # interrupted query's read -- a normal, actionable connection error
+        # is fine, but a permanently leaked claim is not, whether reported
+        # to another fiber ("in use by") or to this one ("still waiting").
+        begin
+          @multi_client.query("SELECT 1")
+        rescue Mysql2::Error => e
+          expect(e.message).not_to match(/This connection is in use by|still waiting for a result/)
```
sodabrew, line 1113: This assertion sits inside a `rescue` clause that may never execute, so the spec can pass without checking anything.

If the follow-up query succeeds without raising (e.g. the connection happened to still be usable, or reconnect silently recovered it), the `rescue` body — the only place the expectation lives — never runs, and the example passes having verified nothing. A regression that leaves the claim-tracking broken in a way that doesn't manifest as an immediate error on this specific follow-up query would slip through this spec undetected.

Worth noting this isn't a new pattern -- the existing Thread#exit-mid-query spec (client_spec.rb:1918-1922, from #1392/#1508) uses the identical shape. So this isn't a one-off mistake, it's following file convention. But compare `result_spec.rb`'s streaming-interrupted-by-net_write_timeout test, which uses the same conditional-rescue shape *with a comment explaining why the failure may or may not fire* (platform/OpenSSL-build-dependent). Neither this spec nor its Thread#exit sibling carries that kind of justification -- they read as "this should always raise," but the code doesn't actually enforce it. Might be worth asserting something unconditionally here instead (e.g. that the follow-up query either succeeds or raises a *different* class of error), and possibly applying the same fix to the older spec while in the area.

ext/mysql2/client.c
```diff
@@ -1109,6 +1118,37 @@ static VALUE disconnect_and_mark_inactive(VALUE self) {
...
+  struct query_completion *completion = (void *)completionval;
+  mysql_client_wrapper *wrapper = completion->wrapper;
+
+  if (completion->completed) {
+    wrapper->active_fiber = Qnil;
+  } else {
+#ifndef _WIN32
+    invalidate_after_interrupted_query(wrapper);
+#else
+    wrapper->active_fiber = Qnil;
```
sodabrew, line 1140: Minor, not a bug -- but sharper version of this comment after checking current master (#1571 has since merged): `invalidate_after_interrupted_query`, `invalidate_socket`, and `invalidate_fd` are all `#ifndef _WIN32`-only. They don't exist as compiled symbols on Windows at all, so this branch isn't skipping an existing shared helper -- there currently isn't one it could call on Windows. `disconnect_and_mark_inactive` (client.c:1092) hand-rolls the same `active_fiber=nil; state=IDLE; if CONNECTED: close(fd); fd=-1;` sequence for the same reason.

So the fix isn't "have `invalidate_after_interrupted_query` branch internally" (it can't -- it's excluded from Windows builds entirely). It'd need a small state-reset helper compiled on *both* platforms (Windows using plain `close()`, POSIX calling into `invalidate_socket()`), so this function and `disconnect_and_mark_inactive` both call one shared thing instead of each re-deriving the Windows fallback by hand. Non-blocking either way -- flagging for whenever someone's next in this area.

spec/mysql2/client_spec.rb
jeremy: Fair. The shape does verify the property -- every leaked-claim manifestation raises a Mysql2::Error matching the guarded pattern on that exact call, and a successful follow-up query is itself proof there's no lockout -- but nothing in the spec said so, and it cost you a re-derivation on first read. #1573 restructures both this spec and the Thread#exit sibling to capture the follow-up error and assert unconditionally, with comments enumerating the acceptable outcomes.

ext/mysql2/client.c
jeremy: Agreed on the shape. One wrinkle for whoever extracts it: the hand-rolled sites aren't quite identical. disconnect_and_mark_inactive also resets client->status = MYSQL_STATUS_READY on both platforms (client.c:1251), while release_claim_or_disconnect (client.c:1267) and discard! (client.c:925, the third hand-roller of the Windows close idiom) don't touch status at all. So the shared helper has to decide whether the status reset is part of the common sequence or stays caller-specific -- not a mechanical extraction. Will fold it into the next client.c change in this area rather than churn the Windows path in a standalone refactor.

Issue #1526, opened 2026-08-13
Proposal: elect result bind types by cast mode — real cast:false/:fast on prepared statements


jeremy:
Part of the roadmap in #1510.
**Category: headline capability** · `effort:L · risk:med · ready:after-#1500 · layer:C+client-lib`

Closes #1314 and supplies the substrate #1276 needs.

## Problem

The prepared-statement result path binds every column as its server type — `buffer_type = fields[i].type` unconditionally (`result.c:842`) — so the binary protocol delivers native ints/doubles/MYSQL_TIME and full casting is the only possible behavior. The cast options are then forcibly overridden with a warning (`result.c:1839`, ":cast is forced for prepared statements", spec-pinned). #1314 has asked for cast modes on statements since 2022.

The consequence grew this year: `cast: false` and `cast: :fast` (#1489/#1490) exist only on the text path, which makes prepared statements the *slow* path for any consumer that wants raw strings or deferred casting — the largest untouched perf/API surface in the gem.

## Proposal

Elect `buffer_type` from the cast mode instead of copying the server type:

- **`cast: false` / `:fast`**: bind `MYSQL_TYPE_STRING` for every column. The client library converts server values to whatever type is bound (the documented `mysql_stmt_bind_result` conversion table), so the fetch produces strings without a Ruby-side cast layer, and the override warning at `result.c:1839` is lifted.
- **`cast: true` stays byte-for-byte today's behavior** — server-type binds, full casting.
- **Rider for #1276**: the same election point makes DOUBLE-for-NEWDECIMAL expressible later (decimal-as-float without BigDecimal), but that's a follow-on option, not this change.

Two verified implementation burdens, neither optional:

1. **Buffer sizing.** For a stored binary result, `max_length` reflects binary wire lengths (4 bytes for an INT), not converted string lengths ("2147483647" is 10 bytes). STRING binds for numeric/temporal columns need type-derived sizing (21 for LONGLONG, 27 for DATETIME(6)) or the truncation-refetch machinery #1500 is building — which is why this sequences after it. Today's `result.c:987-988` IMPLBUG raise fires otherwise.
2. **The per-bind error flag mysql2 already pays for is never read.** `wrapper->error` is xcalloc'd (`result.c:838`), wired into every bind (`result.c:900`), freed (`result.c:180`) — and no code consults it. Conversion under type override is exactly where it earns its keep: check it per fetch and raise on real conversion errors instead of silently truncating.

## Evidence and precedent

[...]

Issue #1525, opened 2026-08-13
Proposal: cache prepared-statement metadata artifacts across executes


jeremy:
Part of the roadmap in #1510.
**Category: gap filler** · `effort:M · risk:med · ready:after-#1500 · layer:C`

Builds directly on the buffer model in the in-flight #1500.

## Problem

Every `Statement#execute` rebuilds everything from scratch, even when the same statement runs ten thousand times against an unchanged result shape:

- `mysql_stmt_result_metadata` is called fresh (`statement.c:581`) and a brand-new Result object is constructed per execute.
- Result buffers are re-xcalloc'd per Result: the MYSQL_BIND array plus is_null/error/length arrays (`result.c:836-839`), then one buffer per column (`result.c:842-903`). #1459's "bind once" is once per *result set*, not cross-execute.
- The field-name VALUE array is rebuilt per Result (`result.c:925-927`); #1487's batch materialization is likewise per-Result.
- `mysql_stmt_wrapper` (`statement.h:4-10`) carries no metadata state at all — `rb_mysql_stmt_mark` marks only `client` (`statement.c:22-27`).

For the workload where prepared statements matter — high-QPS point reads under ActiveRecord's `prepared_statements: true` — this fixed per-execute overhead dominates 1-row results.

## Proposal

Cache the *derived* artifacts on the statement wrapper — field-name VALUEs, decode setup, and the bind buffers — and validate them per execute against fresh metadata.

Two constraints I consider load-bearing:

1. **Metadata must still be read every execute.** `mysql_stmt_result_metadata` is a cheap local call post-execute, and fields go stale after ALTER TABLE; what gets cached is what's built *from* the MYSQL_FIELDs, never the decision to skip reading them. Validation: field count plus per-field name/type/flags/charsetnr against the cache; rebuild on mismatch.
2. **memcmp validation is insufficient for the buffers.** char[] buffers are sized from `fields[i].max_length` (`result.c:893-894`), which is data-dependent per execute (computed by store_result under `STMT_ATTR_UPDATE_MAX_LENGTH`, `statement.c:196-200`) — identical schema, different rows, different sizes. Buffer reuse therefore needs grow-to-fit, which is exactly the xrealloc + `mysql_stmt_fetch_column` refetch model #1500 introduces. This proposal should be written on top of #1500, not master.

## Evidence and precedent

[...]

Issue #1520, opened 2026-08-13
Proposal: drop the STMT_ATTR_UPDATE_MAX_LENGTH prepass once truncation-regrow lands


jeremy:
Part of the roadmap in #1510.
**Category: gap filler** · `effort:M · risk:med · ready:after-#1500 · layer:C+client-lib`

Residual work stacked on #1500, whose truncation-regrow supplies the machinery this deletion depends on.

## Problem

Every prepared statement sets `STMT_ATTR_UPDATE_MAX_LENGTH` at prepare (`statement.c:196-202`), which makes `mysql_stmt_store_result` take an extra pass over the stored result to populate `fields[i].max_length`; the buffered fetch path then sizes every variable-length column buffer to that max_length (`result.c:893-894`).

Two costs:

1. **The extra scan** — the client library walks the entire stored result once before mysql2 reads a single row, on every buffered stmt store.
2. **Pathological allocation** — buffer size is the largest value in the column. One 10MB blob among 10k rows means a 10MB buffer per string-ish column, allocated before the first row is read, even when the caller stops after one narrow row.

It also forks the stmt code: streaming results have no populated max_length at all, which is why buffered and streaming paths diverge and why the IMPLBUG raise at `result.c:988` exists.

## Proposal

Once #1500 lands (on `MYSQL_DATA_TRUNCATED`, xrealloc the column to `wrapper->length[j]` and refetch via `mysql_stmt_fetch_column`), the prepass is obsolete:

- Delete the attr set at `statement.c:196-202`.
- Change `result.c:893-894` to a small fixed initial size; truncation becomes the sanctioned grow path everywhere.
- The IMPLBUG raise becomes unreachable by design; buffered and streaming stmt paths unify, shrinking the test matrix.

Follow-on residual, folded in only if the code is open anyway: #1500 refetches a grown column from offset 0. Passing the already-read byte count as `mysql_stmt_fetch_column`'s offset makes the refetch tail-only — marginal (one redundant prefix memcpy per grow event, and grows happen at most once per column per high-water value), so it rides along rather than standing alone.

## Evidence and precedent

[...]

Issue #1288, opened 2022-12-15
MySQL 8: 0-Bytes in result.fields when field names are truncated


moritzhoeppner:
I noticed that strings in result.fields contain 0-Bytes at the end after executing a GROUP BY-Statement with a very long field name.

Steps to reproduce:

```
client = Mysql2::Client.new(...)
client.query("USE test_db")
[...]

PR #1564, opened 2026-08-14
Cache statement metadata artifacts across executes

jeremy:
Proposed in #1525.

Every Statement#execute rebuilt everything derived from result metadata from scratch, even for the workload prepared statements exist for -- the same statement executed thousands of times against an unchanged result shape. The field-name VALUE array was rebuilt per Result, and the MYSQL_BIND array, the is_null/error/length arrays, and one buffer per column were xcalloc'd per Result and freed with it (rb_mysql_result_alloc_result_buffers, result.c:989).

Now the statement wrapper carries a cross-execute cache of those artifacts (statement.h:36-44). Metadata is still read on every execute -- mysql_stmt_result_metadata is a cheap local call, and fields go stale under ALTER TABLE -- and compared against a snapshot of the shape the artifacts were built from: field count plus per-field name, type, flags, and charsetnr (mysql2_stmt_validate_metadata_cache, statement.c:123, called at statement.c:736). Names are compared NUL-trimmed, the same rule the name derivation uses (#1288). On a match the new Result adopts the cached buffers and a dup of the cached field-name array (rb_mysql_result_to_obj, result.c:2266-2299); on a miss the cache is dropped and rebuilt. When a Result is freed it hands its buffers back to the statement and stores a dup of its field-name array (rb_mysql_result_free_result, result.c:186-231). Statement#close and the statement's final free clear the cache (statement.c:869, :189).

Load-bearing details:

- Buffer sizes are data-dependent -- char[] buffers are sized from fields[i].max_length, computed from whatever rows the previous execute returned -- so size is deliberately not part of validation. An adopted buffer smaller than this execute's widest value takes the existing MYSQL_DATA_TRUNCATED grow-and-refetch path (result.c:1145), exactly as a streaming fetch that never had max_length does. Verified by breaking that path: the widening-data spec then returns truncated bytes.

- The cached field-name array is a private master copy: adoption hands each Result rb_ary_dup(cache) (result.c:2296), harvest stores rb_ary_dup(fields) (result.c:229), so mutating one Result#fields can never reach another execute's keys. Both dups are individually load-bearing -- the mutation-isolation spec fails with either one removed. Steady state is one array dup per execute in exchange for skipping the N per-field interned-string lookups and the array build.

- Field names are only harvested from and adopted by non-streaming Results, and only when built under the same :symbolize_keys as the adopting execute. A non-streaming prepared Result is materialized and freed entirely inside Statement#execute, so its array provably carries the execute-time options and no user code has touched it; a streaming Result's names are built under its first #each's options, which may differ. Buffers, which are never user-visible, are reused in both modes.

- The new VALUE rooted on the statement wrapper (cached_fields) is marked and compacted in rb_mysql_stmt_mark/rb_mysql_stmt_compact (statement.c:29, :48) -- the #1068 bug class. Verified by removing the hooks: the compaction spec segfaults.

- metadata_epoch (statement.h:38, result.h:99): a Result stranded holding buffers by a mid-iteration raise can be GC'd after a later execute has rebuilt the snapshot around an ALTERed shape; its dfree would hand back buffers whose bind types no longer match, and the next execute would adopt them and silently convert values (an INT-shaped bind against a column now VARCHAR converts "1" to 1). The Result records the epoch it was created under and buffers are handed back only on a match (result.c:165). The window only opens when GC collects the strand between a later execute's validation and that execute's own free, so it has no deterministic spec: under GC.stress the strand is always collected before validation, whose mismatch path already clears the stale handback -- verified by removing the epoch check and failing to provoke the bug under stress, which is exactly why the guard stays.

Numbers (thelio: Linux x86_64, MySQL 9.6 over local TCP, ruby 4.0.6; wall = median of 9 samples x 400 ops; allocation counts over 4000 ops; C-heap traffic via an LD_PRELOAD malloc counter; two interleaved base/branch repetitions, deltas identical in both):

```
arm                C mallocs/op          C KB/op        Ruby objs/op   wall
point_read_10col    27.0 -> 13.0  (-52%) 20.3 -> 18.9   28.0 (flat)    flat
point_read_100col  123.0 -> 18.0  (-85%) 56.7 -> 43.9   211.0 (flat)   flat
rows100_10col      128.0 -> 114.0 (-11%) 73.9 -> 72.5   919.0 (flat)   flat
```

This is an allocation-pressure change and wall time is flat, matching go-sql-driver's merged version of the same mechanism (go PR 1711: allocs/op -98.4%, wall flat at p=1.000). One refinement to the issue's sizing: Ruby-object allocations are flat too, because field names were already interned process-wide -- mysql2 was already ahead of go on the axis that produced go's -98%. What this removes is C-heap churn: the per-execute MYSQL_BIND array, the three flag/length arrays, one xmalloc per column, and the field-name array build. That is the -52%/-85% malloc-call reduction above; GC.count deltas were unchanged.

Behavior deltas and coverage limits:

- ALTER TABLE changing a column's type or flags between executes is validated and rebuilt. Specs cover the type change and a signed->unsigned flip, which would otherwise silently mis-decode values >= 2^31. A rename alone never reaches validation on the server tested: MySQL 9.6 pins a prepared handle's result-set names for the life of the handle -- even Statement#fields, which re-reads metadata per call, reports the old name, and master behaves identically. The rename spec therefore pins cached names to whatever fresh metadata reports (row keys == stmt.fields) rather than demanding the new name; on a server that does propagate renames, stale cached names would break that equality.

- ALTER TABLE changing the column count between executes fails inside libmysqlclient ("The number of columns in the result set differs from the number of bound buffers") before metadata can be read. Unchanged from master, verified side by side.

Note for parallel review: #1520 and #1526 touch the same sizing/binding code; this is built against master, and any overlap will surface as a textual conflict in the alloc/free paths at merge time.

### DIFF
--- a/ext/mysql2/result.c
+++ b/ext/mysql2/result.c
@@ -176,7 +176,17 @@ static void rb_mysql_result_free_result(mysql2_result_wrapper * wrapper, int fro
                  && wrapper->client_wrapper;
 
     if (wrapper->stmt_wrapper) {
-      if (!wrapper->stmt_wrapper->closed) {
+      mysql_stmt_wrapper *stmt_wrapper = wrapper->stmt_wrapper;
+      /* Whether the statement's metadata snapshot still describes the shape
+       * these artifacts were built for: an intervening execute can have
+       * rebuilt it (ALTER TABLE) while this Result sat unfreed after a
+       * mid-iteration raise, and handing back buffers bound for the old
+       * shape would make a later execute silently convert values into the
+       * wrong C types. */
+      int cache_compatible = !stmt_wrapper->closed &&
+                             stmt_wrapper->metadata_epoch == wrapper->stmt_metadata_epoch;
+
+      if (!stmt_wrapper->closed) {
         if (defer_free) {
           mysql2_enqueue_pending_result_free(wrapper->client_wrapper, NULL, wrapper->stmt_wrapper->stmt);
         } else {
@@ -194,11 +204,50 @@ static void rb_mysql_result_free_result(mysql2_result_wrapper * wrapper, int fro
         wrapper->stmt_wrapper->stmt->bind_result_done = 0;
       }
 
-      if (wrapper->statement != Qnil) {
-        decr_mysql2_stmt(wrapper->stmt_wrapper);
+      /* Hand the result buffers back to the statement for the next execute
+       * to adopt, instead of freeing them, whenever the cache slot is open
+       * and the shape still matches. Pointer moves only, so this is safe
+       * from the GC dfree path too. Everything cache-related happens before
+       * the decr below, which can free stmt_wrapper outright when this
+       * Result held the last reference. The bind-type election travels with
+       * the buffers (cached_result_buffers_string_binds): an adopting
+       * execute under the other cast mode sees the mismatch at fetch time
+       * and re-elects, exactly as a mid-result #each mode switch does. */
+      if (wrapper->result_buffers &&
+          cache_compatible && stmt_wrapper->cached_result_buffers == NULL) {
+        stmt_wrapper->cached_result_buffers = wrapper->result_buffers;
+        stmt_wrapper->cached_is_null = wrapper->is_null;
+        stmt_wrapper->cached_error = wrapper->error;
+        stmt_wrapper->cached_length = wrapper->length;
+        stmt_wrapper->cached_result_buffers_string_binds = wrapper->result_buffers_string_binds;
+        /* Clue that the next statement execute will need to allocate or
+         * adopt a result buffer. */
+        wrapper->result_buffers = NULL;
+        wrapper->result_buffers_bound = 0;
+      } else {
+        rb_mysql_result_free_result_buffers(wrapper);
       }
 
-      rb_mysql_result_free_result_buffers(wrapper);
+      /* Hand back the field-name array as a private dup, so caller
+       * mutations of this Result's #fields can never reach the cache. Only
+       * from the ordinary free path -- rb_ary_dup allocates, which a GC
+       * dfree callback must not -- and only for a non-streaming Result:
+       * those are materialized and freed entirely inside Statement#execute,
+       * so the array provably carries the execute-time :symbolize_keys and
+       * no user code has run against it. A streaming Result's names are
+       * built under its first #each's options instead, which may differ. */
+      if (!from_dfree_callback && !wrapper->is_streaming && cache_compatible &&
+          wrapper->fields != Qnil &&
+          (my_ulonglong)RARRAY_LEN(wrapper->fields) == wrapper->numberOfFields &&
+          (stmt_wrapper->cached_fields == Qnil ||
+           stmt_wrapper->cached_fields_symbolized != wrapper->fields_symbolized)) {
+        stmt_wrapper->cached_fields = rb_ary_dup(wrapper->fields);
+        stmt_wrapper->cached_fields_symbolized = wrapper->fields_symbolized;
+      }
+
+      if (wrapper->statement != Qnil) {
+        decr_mysql2_stmt(stmt_wrapper);
+      }
     }
 
     /* For prepared statements, wrapper->result is the result metadata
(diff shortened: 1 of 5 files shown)


sodabrew:
This is great. It's a complicated change so I want to read it over a second time.

jeremy:
Thanks — take all the time it needs. For the second pass: the commit message walks each load-bearing detail and how it was verified, but the two spots that most reward scrutiny are

1. **The ownership handoff.** A non-streaming Result frees by handing its buffers back to the statement (`rb_mysql_result_free_result`, result.c), gated by `metadata_epoch` so a Result stranded by a mid-iteration raise can't hand back buffers whose bind types no longer match the current shape.
2. **The validation compare** (`mysql2_stmt_validate_metadata_cache`, statement.c), which defines what "same shape" means: field count plus per-field name/type/flags/charsetnr, names NUL-trimmed per #1288. Buffer *size* is deliberately excluded — an undersized adopted buffer takes the existing `MYSQL_DATA_TRUNCATED` grow-and-refetch path, same as a streaming fetch that never had `max_length`.

One sequencing note: #1566 and #1567 touch adjacent statement/result buffer code, so these three will conflict textually. Whatever order you prefer to land them, I'll rebase the others.

sodabrew (commented):
Reviewed the cross-execute metadata/buffer caching in detail -- the riskiest part by construction (manual buffer lifetime management plus a GC-compaction-sensitive cached VALUE), and it holds up under scrutiny. Checked and confirmed correct:

- The new `cached_fields` VALUE field is added to both `rb_mysql_stmt_mark` and `rb_mysql_stmt_compact` in the same commit -- this is exactly the #1068-class bug this repo has hit before, and the epoch-guard mechanism is clearly built with that history in mind (cites #1068 directly).
- Traced the epoch guard through: buffer/field-count consistency between `cached_field_meta` and `cached_result_buffers` is transitively guaranteed by the epoch check -- any field-count change forces a cache-clear + epoch bump before either could be repopulated, so there's no path to a size-mismatched free.
- `mysql2_stmt_metadata_cache_clear`'s free loops use `cached_field_count` correctly ordered relative to when that count gets overwritten -- no free-with-wrong-size bug.
- Checked `wrapper->numberOfFields` against the full (undiffed) `rb_mysql_result_to_obj`: it's 0-initialized and never touched again in that function except by this PR's adoption code, so nothing later clobbers the adopted value.

One open question I couldn't fully resolve from the diff alone: `wrapper->fields_symbolized` is set inside `rb_mysql_result_fetch_field`'s field-population branch. I wanted to confirm it reliably reflects a single, stable population per Result regardless of how a caller invokes `#each`/`#fields` with varying `symbolize_keys`, since the cache-adoption logic in `rb_mysql_result_free_result` depends on that value being trustworthy. I'll follow up with the answer once I've traced/tested it rather than leave it as a guess.

Everything else -- the hand-back/hand-out cache slot mechanics, the private-dup mutation isolation for `#fields`, the ALTER TABLE invalidation path -- looks solid and the spec coverage for each of these seams is thorough.

sodabrew:
Followed up on my earlier open question above (whether `fields_symbolized` reliably tracks what `wrapper->fields` actually holds) by inspection and a live probe against this branch merged onto current master.

**By inspection**: `wrapper->fields_symbolized = symbolize_keys ? 1 : 0;` sits inside the `if (rb_field == Qnil)` guard in `rb_mysql_result_fetch_field` -- it only fires when actually populating a new array slot, never on a cache-hit read. The array itself is populated exactly once per Result, per the existing comment on `rb_mysql_result_materialize_field_names` ("a per-each :symbolize_keys caches names first-call-wins" -- pre-existing behavior, not something this PR touches). So the flag is set atomically alongside the one real population event and can't diverge from the array's actual contents.

**Empirically**, against a build of this branch:

- Calling `#each` twice on the *same* Result with different `symbolize_keys` reproduces the documented first-call-wins quirk (both calls return the first call's key style) -- confirms it's pre-existing and orthogonal to the caching here.
- Across separate executes on the same statement, the cache correctly refuses to reuse a mismatched `symbolize_keys`: an execute with `symbolize_keys: false` right after one with `true` got plain string keys (not stale symbols), and a following `symbolize_keys: true` execute got symbols again (not stale strings from the string-keyed execute in between). No leakage in either direction.

No bug here -- consider my earlier note resolved. Nice work on the epoch-guard mechanism; traced the buffer/field-count consistency argument through fully and it holds.

Issue #1511, opened 2026-08-13
Proposal: try KILL QUERY before destroying the connection on timeout/interrupt


jeremy:
Part of the roadmap in #1510.
**Category: headline capability** · `effort:M · risk:med · ready:after-train (lib-only version: now) · layer:lib+C`

Addresses #1091 (wall-clock timeouts), and supplies the hard-kill primitive that died with #1433.

## Problem

When a query exceeds `read_timeout` or the calling thread is interrupted, mysql2 tears the connection down unconditionally: the wait loop (`client.c:997`, `rb_wait_for_single_fd`) unwinds through `rb_rescue2` (`client.c:1137`) into `disconnect_and_raise` (`client.c:950-968`), which invalidates the fd and marks the client closed.

Two costs:

1. **The server keeps executing the abandoned query** — holding locks, burning CPU, filling tmp tables — until it finishes or someone notices. The client walked away; nobody told the server.
2. **The connection is always sacrificed**, even though a timeout on an otherwise healthy connection is recoverable in principle.

`#1092`/`#1492` don't touch this: sodabrew has said explicitly that #1492 (next_result GVL work) leaves #1091 open.

## Proposal

An escalation ladder, mirroring what the mysql CLI itself does on Ctrl-C:

1. On timeout/interrupt with a query in flight, **issue `KILL QUERY <thread_id>` from a short-lived helper connection** (we already expose the id: `Client#thread_id`, `client.c:1429`). The helper is bounded: small connect timeout, one statement, closed immediately.
2. **Resume waiting on the original connection for a bounded grace period.** The expected outcome is the server abandoning the query and the client reading `ER_QUERY_INTERRUPTED` (1317) as an ordinary error packet — the protocol stream stays in sync, and **the connection survives** to run its next query.
3. Only if the grace period expires (server unreachable, packet truly lost mid-stream) fall back to today's behavior: invalidate and raise.

A `KILL CONNECTION` variant of the same machinery gives pools a sanctioned way to hard-kill a stuck connection — the capability that was lost when #1433's interrupting-close was closed unmerged. Because the helper never touches the victim's `MYSQL*`, this has no interaction with the `active_fiber` claim model.

### Degraded version that ships first

A pure-`lib/` patch: rescue `Mysql2::Error::TimeoutError`, fire the helper-connection `KILL QUERY`, re-raise. The connection is still lost (no C changes), but the server-side query stops burning. That's most of the operational value for a fraction of the risk, and it's independently revertible.

[...]

Issue #1091, opened 2019-12-09
[BUG] Timeout does not apply when buffer is large


damianthe:
While looking into making mysql2 support non-integer timeouts, @pushrax and I found a bug in the client.

There is a specific edge case where neither the [do_query file descriptor](https://github.com/brianmario/mysql2/blob/master/ext/mysql2/client.c#L665) timeout nor the `MYSQL_OPT_READ_TIMEOUT ` apply.

The error can be reproduced by running a query that responds quickly with the first reply but fills the query response buffer, therefore taking a while to read.

As you can see from the run below, this causes the query the run successfully, despite going over the 1 second read timeout time.

### Instructions to reproduce

Set the client configuration in the following test and run it. 

```ruby
  it "timeout overflow buffer timeout edge case" do
    client = Mysql2::Client.new(
      host: '<insert>',
      port: '<insert>',
      username: '<insert>',
      database: '<insert>',
      connect_timeout: 1,
      read_timeout: 1,
      write_timeout: 1,
    )

    client.query("CREATE TABLE IF NOT EXISTS test_table (c1 MEDIUMTEXT, c2 MEDIUMTEXT)")

    s = StringIO.new
    1500.times { s << "sdjsdjsdjssadaskdhasfhasjdhaskdjhasdjkhsadjkh" }
    3000.times do
    client.query("INSERT INTO test_table (c1, c2) VALUES ('#{s.string}', '#{s.string}')")
    end

    start_time = Time.now
    client.query("SELECT * from test_table")
    query_time = Time.now - start_time

    expect(query_time).to be <= 1.0
  ensure
    client.query("DROP TABLE IF EXISTS test_table")
  end
```

The result on my local machine:
![image](https://user-images.githubusercontent.com/19417045/70471798-e7d57300-1a9b-11ea-8b2c-e9e57f6df8ce.png)

damianthe:
cc @sodabrew

jeremy:
Looks like #807 / #962.

sodabrew:
Investigated whether #1492 (fixed #1033/#807/#962) also resolves this one -- it doesn't, they're different bugs despite looking similar.

#1033/#807/#962 were about `Client#next_result`, the call used *after* a multi-statement query to move to the next result set. It did a raw blocking `mysql_next_result()` that held the GVL and ignored `Thread#raise`/`read_timeout`. #1492 fixed exactly that function.

[...]

PR #1565, opened 2026-08-14
Add :kill_on_timeout, escalating timeouts via KILL QUERY before teardown

jeremy:
Proposed in #1511. Addresses the operational half of #1091, and supplies the hard-kill primitive that died with #1433.

When a query outlives :read_timeout or its thread is interrupted (Timeout.timeout, Thread#raise), mysql2 tears the connection down unconditionally -- and tells the server nothing. The server keeps executing the abandoned query until it finishes on its own, holding locks and burning CPU, and the connection is always sacrificed even though a timeout on an otherwise healthy connection is recoverable.

With kill_on_timeout: true (per connect or per query), Client#query instead runs the escalation ladder the mysql CLI's own Ctrl-C handler uses:

1. Issue KILL QUERY <thread_id> from a short-lived helper connection opened with the victim's own credentials (KILL requires same-user, SUPER, or CONNECTION_ADMIN, so same-user is the default path). The helper is bounded -- 5s connect/read/write timeouts, one statement, closed immediately (lib/mysql2/client.rb:252).
2. Resume waiting up to :kill_grace seconds (default 5) on the original connection. The expected outcome is the server abandoning the query and the client reading ER_QUERY_INTERRUPTED (1317) as an ordinary error packet -- the protocol stream stays in sync and the connection survives to run its next query. If the query finished before the KILL landed, its result is read and discarded (multi-statement leftovers via abandon_results!). Either way the caller still gets the original TimeoutError/interrupt, so Timeout.timeout semantics are preserved.
3. Only if the helper can't connect, the KILL fails, or the grace period expires does the connection get invalidated exactly as before.

### Mechanism

The C surface is deliberately thin; the policy lives in lib:

- The one recoverable window is the read-wait phase (do_query, client.c:1073): the command is fully sent and none of its response read, so the stream is intact. For kill_on_timeout queries that phase runs under rb_rescue2 (do_query_deferring_teardown, client.c:1100) whose companion defers teardown: it marks the connection MYSQL2_CLIENT_CANCEL_PENDING (client.h:20) instead of invalidating the fd, and re-raises (client.c:1092). Interrupted sends and mid-result reads keep today's unconditional teardown -- KILL can't resync a half-written or half-read stream.
- Thread#exit is a non-exception unwind: rb_rescue2 never sees it, so the existing rb_ensure companion tears down exactly as without the option -- no rescue upstream could ever run the escalation (spec: "Thread#exit interrupts an armed query").
- Client#query (lib/mysql2/client.rb:181) rescues, checks cancel_pending? (client.c:1140), and settles the connection under Thread.handle_interrupt(Object => :never): kill_and_resume_cancelled_query (lib/mysql2/client.rb:229) fires the KILL and calls resume_cancelled_query(grace) (client.c:1153), which re-waits on the fd and reuses the ordinary async_result/disconnect_and_mark_inactive read path. An ensure'd abort_cancelled_query (client.c:1187) is rung 3 as a callable: invalidate if still pending, no-op otherwise.
- The helper never touches the victim's MYSQL*, so there is no interaction with the active_fiber claim model; concurrent callers stay fenced off by active_fiber for the whole escalation.

The same machinery is public API:

- Client#kill_query -- cancel this connection's in-flight query from another thread; the blocked #query returns its error/result normally and the connection survives.
- Client#kill_connection -- KILL CONNECTION hard-kill, the sanctioned way for pools to unstick a connection they're about to discard (the capability lost when #1433 was closed unmerged).

Thread-id width: the KILL targets Client#thread_id, the id the server assigned at handshake. The handshake field is 32 bits, so on servers with wider connection ids the README documents SELECT CONNECTION_ID() as the alternative.

### Specs

All in the existing non-Windows guard, each verified to fail against deliberately broken builds:

| deliberate break | specs that catch it |
|-------------------------------------------------|---------------------|
| C deferral disabled (always plain teardown) | read_timeout-survive, ER_QUERY_INTERRUPTED, interrupt-survive, grace-expiry (4) |
| KILL statement never sent | all three keep-alive specs, #kill_query, #kill_connection (5) |
| ladder's rescue Mysql2::Error removed | ER_QUERY_INTERRUPTED, helper-unavailable fallback (2) |
| ensure-based deferral (defers Thread#exit too) | armed Thread#exit spec: fails with "in use by <dead fiber>" wedge (1) |

Server-side verification is direct: after the timeout raises, SHOW PROCESSLIST from a supervisor connection must show the victim session registered but idle (Info NULL) before the victim connection is touched again. Probed against MySQL 9.6: a killed SLEEP reports interruption by returning a result early (exercising the completed-result rung); a killed GET_LOCK wait raises 1317 (exercising the expected-error rung).

### Performance

Overhead on the unarmed hot path is one rb_hash_aref per synchronous query plus a Ruby rescue frame: tight SELECT 1 loop (7 samples x 20k queries, localhost, median qps) -- master 45,497 vs branch 45,286 (-0.5%, within run-to-run noise; sample ranges overlap fully).

### Behavior deltas

All opt-in unless noted:

- With kill_on_timeout enabled, the timeout/interrupt raise is delayed by the escalation: bounded by helper connect+kill (10s worst case) + :kill_grace. Defaults keep this ~15s worst case; typical is tens of milliseconds.
- A query that completes in the KILL race has its result discarded; the deadline contract wins over late delivery.
- Always-on (unarmed paths): Client#query gains a rescue-all that re-raises the same exception object after two cheap no-op C calls; initialize builds a helper-factory lambda per client (closes over the connect options, including the password -- held in the lambda rather than an ivar so #inspect doesn't print it).

### Coverage limits

- Client#query only. Statement#execute has no equivalent wait phase to resume (its interrupt handling is unchanged), and next_result's timeout path is likewise untouched. async: true callers own their own waits, so the ladder never arms there.
- Windows keeps plain teardown: the deferral lives in the non-Windows wait phase; cancel_pending? is always false there and the README documents the limitation.
- The ER_QUERY_INTERRUPTED spec asserts outcomes (timeout raised, connection survives), not the error number, so it stays valid on MariaDB even if a killed GET_LOCK reports differently there.
- Residual interrupt windows in the ladder itself are masked with handle_interrupt(Object => :never); an interrupt landing in the remaining one-call window before the mask (or on the ensure's own call site) degrades to rung-3 teardown -- parity with today -- except the ensure-call-site case, which can leave the connection fenced until its owning thread closes it. That window is the same class every Ruby ensure clause has.

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -23,13 +23,13 @@ Lint/MissingSuper:
 # Offense count: 2
 # Configuration parameters: AllowedMethods, AllowedPatterns, CountRepeatedAttributes.
 Metrics/AbcSize:
-  Max: 92
+  Max: 94
 
 # Offense count: 35
 # Configuration parameters: CountComments, CountAsOne, AllowedMethods, AllowedPatterns.
 # AllowedMethods: refine
 Metrics/BlockLength:
-  Max: 495
+  Max: 516
 
 # Offense count: 1
 # Configuration parameters: CountBlocks, CountModifierForms.
@@ -39,7 +39,7 @@ Metrics/BlockNesting:
 # Offense count: 1
 # Configuration parameters: CountComments, CountAsOne.
 Metrics/ClassLength:
-  Max: 195
+  Max: 236
 
 # Offense count: 3
 # Configuration parameters: AllowedMethods, AllowedPatterns.
@@ -49,7 +49,7 @@ Metrics/CyclomaticComplexity:
 # Offense count: 6
 # Configuration parameters: CountComments, CountAsOne, AllowedMethods, AllowedPatterns.
 Metrics/MethodLength:
-  Max: 54
+  Max: 55
 
 # Offense count: 2
 # Configuration parameters: AllowedMethods, AllowedPatterns.
--- a/README.md
+++ b/README.md
@@ -278,6 +278,8 @@ Mysql2::Client.new(
   :read_timeout = seconds,
   :write_timeout = seconds,
   :connect_timeout = seconds,
+  :kill_on_timeout = true/false,
+  :kill_grace = seconds,
   :connect_attrs = {:program_name => $PROGRAM_NAME, ...},
   :reconnect = true/false,
   :local_infile = true/false,
@@ -302,6 +304,57 @@ type of connection to make, with special interpretation you should be aware of:
 * A value of `"."` on Windows specifies a named-pipe connection.
 * An IPv4 or IPv6 address will result in a TCP connection.
 * Any other value will be looked up as a hostname for a TCP connection.
+
+### Cancelling a runaway query: kill_on_timeout
+
+By default, when a query outlives `:read_timeout` or its thread is interrupted
+(`Timeout.timeout`, `Thread#raise`), mysql2 abandons the connection -- and the
+server keeps executing the query anyway, holding locks and burning CPU until it
+finishes on its own. With `:kill_on_timeout` enabled, mysql2 instead escalates
+the way the mysql CLI's Ctrl-C handler does:
+
+1. Issue `KILL QUERY <thread_id>` from a short-lived helper connection opened
+   with the victim connection's own credentials.
+2. Resume waiting up to `:kill_grace` seconds (default 5) for the server to
+   answer the killed query -- normally `ER_QUERY_INTERRUPTED`, or the query's
+   own result if it finished before the KILL landed. Either way the protocol
+   stream stays in sync and **the connection survives** to run its next query.
+   The original `Mysql2::Error::TimeoutError` (or interrupting exception) is
+   still raised.
+3. Only if the helper can't connect, the KILL fails, or the grace period
+   expires is the connection torn down as before.
+
+``` ruby
+client = Mysql2::Client.new(read_timeout: 5, kill_on_timeout: true)
+```
+
+Both options may also be passed per query:
+
+``` ruby
+client.query("SELECT ...", kill_on_timeout: true, kill_grace: 2)
+```
+
+Notes:
+
+* `KILL QUERY` requires the same user, `SUPER`, or `CONNECTION_ADMIN`; the
+  helper connects with the victim's own credentials, so the same-user path is
+  the default.
+* The thread id used is the one the server assigned at handshake
+  (`Client#thread_id`). The handshake field is 32 bits wide, so on servers
+  whose connection ids exceed 32 bits, run `SELECT CONNECTION_ID()` yourself
+  and issue the KILL from your own connection instead.
+* Not supported on Windows, where the plain teardown behavior is unchanged.
+
+The same machinery is exposed directly, usable from another thread while the
+connection is stuck mid-query:
+
+``` ruby
+client.kill_query      # KILL QUERY <thread_id>: the blocked #query call
+                       # returns its error/result and the connection survives
+client.kill_connection # KILL CONNECTION <thread_id>: hard-kill for a
+                       # connection you are about to throw away, e.g. from a
+                       # connection pool reaper
+```
 
 ### Secure connections with SSL/TLS
 
--- a/ext/mysql2/client.c
+++ b/ext/mysql2/client.c
@@ -18,7 +18,7 @@
 
 VALUE cMysql2Client;
 extern VALUE mMysql2, cMysql2Error, cMysql2ConnectionError, cMysql2TimeoutError;
-static VALUE sym_id, sym_version, sym_header_version, sym_async, sym_symbolize_keys, sym_as, sym_array, sym_stream;
+static VALUE sym_id, sym_version, sym_header_version, sym_async, sym_symbolize_keys, sym_as, sym_array, sym_stream, sym_kill_on_timeout;
 static ID intern_brackets, intern_merge, intern_merge_bang, intern_new_with_args,
   intern_current_query_options, intern_read_timeout, intern_values;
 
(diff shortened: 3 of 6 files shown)


sodabrew (commented):
Reviewed the escalation ladder in detail -- the C-side state machine (`MYSQL2_CLIENT_CANCEL_PENDING`, `cancel_pending?`/`resume_cancelled_query`/`abort_cancelled_query`) is careful and I couldn't find a fault in it. Specifically checked and confirmed correct:

- The new functions are placed inside the pre-existing `#ifndef _WIN32` block that also gates `do_query` itself (verified against the full file, not just the diff) -- no compile break on Windows, and the call site in `rb_mysql_query` is correctly confined to the non-Windows branch.
- The apparent double-invalidate risk in `abort_cancelled_query` (called unconditionally from Ruby's `ensure`, after `resume_cancelled_query` may have already invalidated on grace-expiry) isn't real: `invalidate_after_interrupted_query` resets `wrapper->state` to `IDLE`, so the state guard correctly no-ops the redundant call.
- State transitions between the C extension's `CANCEL_PENDING` and the Ruby-side trio are consistent in both the success (query finished before KILL landed) and grace-expiry paths.

One real bug, in `lib/mysql2/client.rb#query`:

```ruby
rescue Exception => e
  begin
    Thread.handle_interrupt(Object => :never) do
      kill_and_resume_cancelled_query(opts) if cancel_pending?
    end
  ensure
    abort_cancelled_query
  end
  raise e
end
```

`kill_and_resume_cancelled_query` only rescues `Mysql2::Error` internally. Anything else it raises -- e.g. `Float(opts[:kill_grace] || 5)` raising `ArgumentError` on a bad `kill_grace:` value (confirmed: `Float("abc")` raises `ArgumentError: invalid value for Float(): "abc"`), or any bug surfacing in the escalation path itself -- propagates out of the `begin/ensure` block and skips `raise e` entirely. The caller then sees that exception instead of the original `TimeoutError`/interrupt.

Concretely: `client.query(sql, kill_on_timeout: true, kill_grace: "oops")` on a query that actually times out would raise `ArgumentError` instead of `Mysql2::Error::TimeoutError` -- a caller doing `rescue Mysql2::Error::TimeoutError` around it would never see the timeout at all, just an unrelated crash from the safety net that's supposed to be transparent.

Narrow trigger (needs a bug or bad input specifically inside the escalation path while a real timeout is also in flight), but worth closing before merge -- the escalation's own failures shouldn't be able to shadow the exception it exists to preserve. A `rescue StandardError` (or narrower, matching whatever the escalation can legitimately raise) around the `Thread.handle_interrupt(Object => :never)` block, falling through to the existing `ensure`/`raise e`, would fix it without changing the happy-path behavior.

Everything else here -- the KILL/grace/re-elect protocol, the Thread#exit non-exception-unwind handling, the helper connection factory closing over credentials instead of storing them in an ivar, the disclosed residual-interrupt-window caveats -- looks solid and thoroughly tested.

jeremy:
Thanks for the thorough read, @sodabrew — the rescue-path bug is real, and your `kill_grace: "oops"` repro is exactly the failure: `Float()` raising out of `kill_and_resume_cancelled_query` skipped `raise e`, so a caller waiting on `rescue Mysql2::Error::TimeoutError` got an `ArgumentError` from the safety net instead. Fixed in 8f830dc (branch also rebased onto current master, see below), two parts:

1. **Your suggested shape, adopted**: the `Thread.handle_interrupt(Object => :never)` block now has a `rescue StandardError` that falls through to the existing `ensure`/`raise e`. I kept it at `StandardError` rather than narrower — the net's job is to stay transparent for escalation-path bugs we *haven't* predicted, and anything narrower re-opens the same hole for the next one. (Not `Exception`: async interrupts are already deferred by the `:never` mask, and what remains above `StandardError` — `NoMemoryError`, `SystemExit` — shouldn't be suppressed.) The internal `rescue Mysql2::Error` stays as the expected-path resolution, so the outer net remains a should-never-fire backstop.

2. **Eager `:kill_grace` resolution**: `Float(opts[:kill_grace] || 5)` moved to the top of `#query`, resolved when `:kill_on_timeout` is armed, before the query is in flight. Malformed input now raises `ArgumentError` deterministically at the call site — with no timeout in hand to shadow — instead of surfacing mid-escalation (where, pre-fix, it shadowed the timeout, and post-fix-1 alone, it would silently degrade the KILL to a plain teardown without ever telling you your option is broken). The escalation path itself no longer parses input at all.

Specs added, both verified to fail against the pre-fix code:

- *"rejects a malformed :kill_grace at the call site, before the query runs"* — expects `ArgumentError` and asserts the connection is untouched and still usable. The second assertion is the discriminating one: the buggy code also raised `ArgumentError`, but only after the timeout fired and the connection was torn down.
- *"re-raises the original timeout when the escalation itself fails"* — your concrete scenario with the input error generalized to an injected escalation-path bug (`kill_query` stubbed to raise `ArgumentError`) while a real `read_timeout` fires: the caller sees `Mysql2::Error::TimeoutError`, and the connection is forfeited exactly as if `kill_on_timeout` were off.

---

**Separately, the three failing CI legs on the previous head (ubuntu ruby 2.7/3.0 × mariadb 10.6/10.11) are pre-existing on master, not from this PR.** All three fail on the same single spec from #1572:

```
rspec ./spec/mysql2/client_spec.rb:524 # Mysql2::Client TLS option validation
  maps sslverify: true onto ssl_mode: :verify_identity when no ssl_mode is given

  ssl_mode: :verify_identity cannot be enforced by this mysql2 build (client library 3.3.19) ...
```

i.e. the new `sslverify: true → verify_identity` mapping trips the #1570 fail-closed guard on legs that build against MariaDB Connector/C 3.3.19 (< 3.4, no hostname-verification support). Master's own Build runs show the identical spec failing on the identical three legs: [run 32062147833](https://github.com/brianmario/mysql2/actions/runs/32062147833) on 3df3179 — the #1572 merge commit itself — and [run 32064123836](https://github.com/brianmario/mysql2/actions/runs/32064123836) on 9dab159. That spec needs the same enforceability guard its sibling specs at client_spec.rb:269 use; I'll leave that to a follow-up against master rather than smuggling it into this PR.

(Edit: the head was rebased onto current master while addressing this -- 9dab159 grew the same spec block RuboCop's todo caps, so the cap is now set against the merged count.)

sodabrew (approved):
The exception-masking bug from my earlier review is fixed, and fixed more thoroughly than I'd suggested:

- `kill_grace = Float(opts[:kill_grace] || 5) if opts[:kill_on_timeout]` is now resolved *before* the query runs, at the call site -- a malformed `:kill_grace` raises immediately instead of ever having a chance to surface mid-escalation. That's stronger than a rescue: it removes the failure mode by construction rather than catching it after the fact.
- The escalation attempt itself now sits under `rescue StandardError; nil`, so any other failure inside it also falls through to the existing `ensure`/`raise e` instead of replacing the original exception.

Verified against a build of the rebased branch (`d339285`, now past #1578's TLS-spec guard and current on master otherwise):

- Both new specs pass -- "rejects a malformed :kill_grace at the call site, before the query runs" and "re-raises the original timeout when the escalation itself fails" (stubs `kill_query` to raise `ArgumentError`, confirms the caller still sees the original `Mysql2::Error::TimeoutError`).
- Full `:kill_on_timeout`/`kill_query`/`kill_connection` suite: 10/10.
- Full spec suite: 630 examples, 0 failures, 23 pending (all pre-existing environment-skip pendings, none new).

The C side is untouched (line-number shift is purely rebase offset from new master content above it) -- consistent with this being a Ruby-level fix for a Ruby-level bug, exactly where I'd expect it. No further findings from me. LGTM.

Issue #1314, opened 2023-03-07
cast: false doesn't work for prepared statements


D1353L:
`cast: false` option works for the following code
```
connection.query(query, symbolize_keys: true, cast: false).each do |row|
   row
end
```
but it doesn't work for the prepared statement
```
statement = connection.prepare(query)
statement.execute(*sanitized_params, symbolize_keys: true, cast: false)
```
shows `warning: :cast is forced for prepared statements`
but still returns `invalid date (Date::Error)` if the date is all zeroes in DB

sodabrew:
Fixed by #1567 (merged): `cast: false` is now fully honored for prepared statements, including the all-zero-date case here. Verified against current master:

```ruby
statement = client.prepare("SELECT d FROM t") # d DATE, value '0000-00-00'
statement.execute(cast: false).first
# => {"d"=>"0000-00-00"}  -- no warning, no Date::Error, raw string like text-protocol queries

PR #1567, opened 2026-08-14
stmt results: elect bind types by cast mode, honoring cast: false and :fast

jeremy:
Proposed in #1526.

Prepared-statement results bind every column as its server type -- buffer_type = fields[i].type unconditionally (formerly result.c:959) -- so the binary protocol delivers native ints/doubles/MYSQL_TIME and full casting is the only behavior the fetch loop can produce. The cast option is then forcibly overridden with ":cast is forced for prepared statements" (formerly result.c:2024, spec-pinned). cast: false and cast: :fast (#1489/#1490) exist only on the text path, which has made prepared statements the slow path for any consumer that wants raw strings or deferred casting. Requested in #1314 since 2022.

This elects buffer_type from the cast mode instead of copying the server type (rb_mysql_result_alloc_result_buffers):

* cast: false / :fast bind MYSQL_TYPE_STRING for every column. The client library converts each value to its string form on fetch (the documented mysql_stmt_bind_result conversion behavior), so the fetch loop receives text-shaped cells: cast: false materializes them with the exact twin of the text path's cast-none loop, and cast: :fast dispatches on the server type to cast the cheap types from their string bytes with the same mechanisms the text path uses (mysql2_cast_integer, Kernel#Float, the :cast_booleans checks) while deferring DECIMAL and temporals as tagged Strings.
* cast: true keeps today's server-type binds and decode arms byte for byte.
* The forced-cast warning is lifted, and the fetch path re-elects (frees, reallocates, rebinds) the buffers if a later #each switches cast mode mid-result -- reachable only when an earlier iteration stopped short, since non-streaming Statement#execute materializes every row up front with the execute-time options. Per-#each options never apply to already-materialized rows; that pre-existing semantic is unchanged.

Buffer sizing (issue burden 1): for a stored result fields[i].max_length is the binary wire length (4 bytes for an INT), not the converted string width, and for a streaming result it is 0. String binds for fixed-width types therefore get type-derived sizes -- the widest value the type can print -- widened to fields[i].length for ZEROFILL display padding, with FLOAT/DOUBLE given enough headroom (512) that libmysql's my_gcvt never has to round to fit: it silently reduces precision to the bind's buffer_length rather than reporting truncation, so the undersized case has to be unreachable, and is. Variable-length types keep max_length sizing and the MYSQL_DATA_TRUNCATED grow-and-refetch backstop (#1500), which works under string binds exactly as under server-type binds; the new streaming spec pins that byte-for-byte, because a truncated fetch still reports the full length and a size-only assertion passes with garbage bytes.

Per-bind error flag (issue burden 2): under string binds the only condition the client library flags is truncation, which the per-column wrapper->error[] consult in the MYSQL_DATA_TRUNCATED arm already reads (#1500); server-type-to-string conversion is total, so no reachable "real conversion error" is left to raise on, and no new error path is added.

Formatting parity -- the honest risk in the proposal -- was measured, not assumed: a parity matrix comparing text cast: false/:fast against stmt cast: false/:fast came back byte-identical, values and encodings, for every probed column type on both client libraries.

    matrix: TINYINT/SMALLINT/MEDIUMINT/INT/BIGINT (signed and unsigned
    boundaries, ZEROFILL), DECIMAL(10,3)/(10,0), FLOAT and FLOAT(10,3),
    DOUBLE (DBL_MAX/DBL_MIN shortest-round-trip formatting) and
    DOUBLE(30,20), DATE, DATETIME(0)/(6), TIMESTAMP(6), TIME(0)/(6)
    including -838:59:59, YEAR, BIT(1)/(64), CHAR/VARCHAR (utf8mb4 and
    latin1), BLOB/VARBINARY with embedded NULs, ENUM, SET, JSON, GEOMETRY,
    zero and partial-zero dates (sql_mode=''), NULLs, empty strings

    libs: libmysqlclient 9.6.0 (macOS/arm64) and MariaDB Connector/C 3.4.9
    (Linux/x86_64), against MySQL Server 9.6

Byte parity is an empirical property of those client libraries, not a documented guarantee; the contract this change promises is value equivalence. The spec parity contexts now run their literal byte-and-encoding expectations through prepared statements as well as text queries, so a future client-library divergence fails a named spec instead of shifting bytes silently.

Benchmarks (Ryzen 9 7950X Linux host, MariaDB Connector/C 3.4.9, MySQL 9.6 in Docker, load < 0.5, three serial alternating base/branch pairs, 20k rows, each figure the median of 15 samples of min-of-5 full materializations; base = c7a682a):

shape "heavy" (4x DATETIME(6), 4x DECIMAL(10,3), 4x INT, 4x VARCHAR(16))
| mode              | base        | branch      | delta        |
|-------------------|-------------|-------------|--------------|
| stmt cast: true   | 58.7-61.6ms | 59.9-63.2ms | unchanged    |
| stmt cast: :fast  | 58.5-63.7ms | 33.0-35.7ms | about -44%   |
| stmt cast: false  | 58.8-62.9ms | 33.2-37.7ms | about -44%   |
| text cast: false  | 21.8-22.3ms | 20.9-23.0ms | reference    |

shape "ints" (16x INT)
| mode              | base        | branch      | delta        |
|-------------------|-------------|-------------|--------------|
| stmt cast: true   | 10.9-11.1ms | 10.9-11.8ms | unchanged    |
| stmt cast: :fast  | 10.9-11.3ms | 20.9-21.6ms | about +90%   |
| stmt cast: false  | 11.1-11.4ms | 22.4-22.8ms | see below    |

Two honest readings of those numbers. String binds cost about 2x on integer-heavy shapes -- the client library renders text that the fetch loop then re-parses -- so cast: :fast pays off on the decimal/temporal-heavy shapes it exists for rather than as a universal default; the same election point could bind cheap types natively under :fast as a follow-on. (The ints cast: false base column is a forced full cast producing Integers, so its delta is not like-for-like.) And stmt cast: false (~35ms) still trails text cast: false (~22ms) on the heavy shape: this makes the cast modes work on statements, it does not yet make statements the fastest raw path.

Behavior changes to call out in release notes: the ":cast is forced for prepared statements" warning is gone (its two pinned specs now assert the new behavior), and code passing cast: false / :fast to Statement#execute -- or carrying it in query defaults -- now receives Strings where it previously got fully cast values plus a warning.

Out of scope, unchanged: stmt cast: true still raises ArgumentError for TIME values outside 00..23 hours (Time.local(838, ...)), a pre-existing binary-path limitation the text path does not share; the new modes return such values as strings, byte-identical to the text path.

Built against master c7a682a; #1520/#1525 touch neighboring result.c code and whichever lands second will need a mechanical rebase.

### DIFF
--- a/README.md
+++ b/README.md
@@ -690,7 +690,9 @@ result = client.query("SELECT * FROM table", :cast => :fast)
 
 This is for callers that consume mysql2 results directly -- raw-mysql2 pipelines, ETL jobs, and applications that do their own value parsing -- where DECIMAL and temporal columns are often passed through or parsed lazily, and paying BigDecimal/Time construction for every cell up front is waste. Your code must be prepared to receive Strings for those columns. Note that Active Record does not use this option.
 
-Any `:cast` value other than the exact symbol `:fast` (or `false`/`nil`) keeps meaning full casting. Prepared statement results are always fully cast, as with `:cast => false`.
+Any `:cast` value other than the exact symbol `:fast` (or `false`/`nil`) keeps meaning full casting.
+
+Prepared statements honor `:cast => false` and `:cast => :fast` too: result columns are bound as strings, so the client library delivers each value's string form from the binary protocol. Those strings are value-equivalent to the text protocol's, and byte-identical for every column type in the spec suite's parity matrix, under both libmysqlclient and libmariadb. Pass `:cast` to `Statement#execute` (or set it client-wide): non-streaming `#execute` materializes rows internally with `#each`, so a `:cast` passed to a later `Result#each` call only applies to rows not already materialized.
 
 ### Async
 
--- a/ext/mysql2/result.c
+++ b/ext/mysql2/result.c
@@ -121,6 +121,27 @@ static void rb_mysql_result_mark(void * wrapper) {
   }
 }
 
+/* Free the statement result bind buffers so the next fetch (or the next
+ * statement execute) allocates and binds fresh ones. Called from
+ * rb_mysql_result_free_result and when a fetch needs buffers of the other
+ * bind-type election (see result_buffers_string_binds in result.h). */
+static void rb_mysql_result_free_result_buffers(mysql2_result_wrapper *wrapper) {
+  if (wrapper->result_buffers) {
+    unsigned int i;
+    for (i = 0; i < wrapper->numberOfFields; i++) {
+      if (wrapper->result_buffers[i].buffer) {
+        xfree(wrapper->result_buffers[i].buffer);
+      }
+    }
+    xfree(wrapper->result_buffers);
+    xfree(wrapper->is_null);
+    xfree(wrapper->error);
+    xfree(wrapper->length);
+  }
+  wrapper->result_buffers = NULL;
+  wrapper->result_buffers_bound = 0;
+}
+
 /* this may be called manually or during GC.
  *
  * from_dfree_callback must be true when called from rb_mysql_result_free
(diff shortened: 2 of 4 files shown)


sodabrew (approved):
This is great, def no reason why a prepared statement result _needs_ to be cast to type.

Issue #1480, opened 2026-08-11
MariaDB TLS: add MARIADB_OPT_TLS_VERSION / TLS_PEER_FP / TLS_PASSPHRASE support


sodabrew:
## Summary

Investigated current MariaDB TLS best practices (11.8 LTS and 12.3 LTS) against what mysql2 exposes today. The headline capability — Zero-Configuration TLS (MariaDB Server 11.4+, Connector/C 3.4+) — mysql2 already gets automatically, with no code changes, whenever it's built against a sufficiently new MariaDB Connector/C. But MariaDB Connector/C exposes several `MARIADB_OPT_*` connection options with no mysql2-facing equivalent, which a security-conscious MariaDB deployment might want and currently can't reach through this gem.

## Background: Zero-Configuration TLS

MariaDB 11.4+ servers auto-generate and serve a self-signed certificate; instead of CA-chain validation, client and server each derive a signature from the certificate's SHA-256 fingerprint combined with the account's password (already shared via the auth exchange) and compare them. MITM-resistant without any manual cert provisioning. Requires MariaDB Server 11.4+ and Connector/C 3.4+; carried forward unchanged through 11.8 LTS and 12.3 LTS (12.3's own TLS-adjacent changes are server-side/audit-log only — nothing new for client connections).

**mysql2 already meets this transparently:**
- It never overrides `ssl_verify_server_cert` unless the caller explicitly sets `:ssl_mode`/`:sslverify`, so the connector's own secure-by-default behavior isn't fought.
- `default_auth` (`ext/mysql2/client.c`, `set_default_auth`) is a free-text passthrough with no allowlist, so PARSEC — the auth plugin MariaDB recommends pairing with zero-config TLS — negotiates automatically if the linked connector supports it.

**Caveat:** this only holds when mysql2 is built against MariaDB's own Connector/C ≥ 3.4, not MySQL's `libmysqlclient`. Since mysql2 supports building against either (exercised by the `cross-version` CI job), a MySQL-linked build talking to a MariaDB 11.8/12.3 server gets an ordinary CA-based or unverified handshake instead — the mechanism lives in the connector, not the wire protocol.

## Gap: three `MARIADB_OPT_*` options with no mysql2 equivalent

[...]

PR #1568, opened 2026-08-14
Add regression coverage for #879: verify_identity must reject a hostname mismatch

sodabrew:
## Why

#879 has been open since 2018: does mysql2 actually reject a connection under `ssl_mode: :verify_identity` when the server's certificate doesn't cover the hostname you connected with? #1506 (merged) fixed the sibling `:verify_ca` gap and, in tracing that fix, established that `:verify_identity` is wired to the right underlying call on both engines (`MYSQL_OPT_SSL_MODE` on MySQL, `MYSQL_OPT_SSL_VERIFY_SERVER_CERT` on MariaDB Connector/C) -- but "the option gets set" was never actually proven to mean "a mismatch gets rejected." Discussion: #1480 (comment).

This PR is test-only. It doesn't re-chase the original 2018-era Connector/C 2.3.0 report -- scope is current versions only (MariaDB Connector/C 3.4+, MySQL 8.0+/9.x), per the standard set in the #1480/#1506 thread.

## What

New spec in the `SSL` context (`spec/mysql2/client_spec.rb`): connects via `127.0.0.1` under `ssl_mode: :verify_identity`, reusing the existing `spec/ssl/server-cert.pem` fixture, whose subject is `CN=mysql2gem.example.com` with no SAN extension. CI's own `/etc/hosts` entry (`build-ubuntu.yml`/`build-macos.yml`) points `mysql2gem.example.com` at `127.0.0.1`, so both hostnames reach the identical server -- one matches the certificate, one doesn't. No new fixtures needed.

Expected: `Mysql2::Error` when connecting via the mismatched hostname. If CI shows otherwise, that's the actual answer to #879 and tells us where real code needs to change.

## Verification

- `bundle exec rubocop spec/mysql2/client_spec.rb` -- clean.
- `bundle exec rspec` (full suite, without forcing the SSL cert-dir env var so the SSL context's own skip-if-missing-fixture guard behaves normally) -- 546 examples, 0 failures, 12 expected pending.
- **Could not verify the new assertion itself locally.** This machine's local `mysqld` serves its own auto-generated self-signed certificate rather than `spec/ssl/*`, so every SSL-context test -- not just this one -- fails with a CA verification error before ever reaching hostname-check logic. Reconfiguring the local server to serve the repo's fixture certs would mean restarting an instance other local sessions may depend on, so I didn't. CI installs the server with `spec/ssl/*` as its actual cert and is the real verification venue for this test.

### DIFF
--- a/spec/mysql2/client_spec.rb
+++ b/spec/mysql2/client_spec.rb
@@ -236,6 +236,22 @@ RSpec.describe Mysql2::Client do # rubocop:disable Metrics/BlockLength
       end
     end
 
+    it "should reject a connection when ssl_mode is verify_identity and the hostname doesn't match the server's certificate" do
+      # Regression coverage for #879: verify_identity being wired to the right
+      # mysql_options() call (MYSQL_OPT_SSL_MODE on MySQL, MYSQL_OPT_SSL_VERIFY_SERVER_CERT
+      # on MariaDB) was never actually proven to reject a hostname mismatch --
+      # only that the option gets set, which isn't the same thing.
+      #
+      # spec/ssl/server-cert.pem has CN=mysql2gem.example.com and no SAN
+      # extension. option_overrides above connects via exactly that hostname
+      # (ssl_cert_host) so it matches; CI's /etc/hosts also points
+      # mysql2gem.example.com at 127.0.0.1 (see build-ubuntu.yml/build-macos.yml),
+      # so connecting via 127.0.0.1 instead reaches the identical server with a
+      # hostname the certificate doesn't cover -- verify_identity must reject it.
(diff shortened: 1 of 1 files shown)


sodabrew:
More Claude credits tomorrow! 😭

sodabrew:
Closing in favor of #1570, which supersedes this. That PR fixes #879 for real (this PR was test-only, proving the bug existed but not fixing it) and its own spec coverage is a superset of the regression spec here -- including this exact mismatch-by-IP leg plus a mismatch-by-name leg, `tls_info` proof of which verification rung ran, and CA-required/pinning validation. No need to carry a narrower, now-redundant version of the same test separately.

PR #1570, opened 2026-08-15
Enforce hostname verification for ssl_mode: :verify_identity on MariaDB (#879)

### LINKED ISSUE #879: Gem does not verify hostname when `ssl_mode: :verify_identity`

jeremy:
Design agreed in #1480; fixes #879.

ssl_mode: :verify_identity on MariaDB Connector/C never verified the server's hostname. MYSQL_OPT_SSL_VERIFY_SERVER_CERT -- the option mysql2 maps both verify modes to -- only clears the connector's skip-everything gate (tls_allow_invalid_server_cert); which checks actually run is chosen by the connector's local-peer heuristic (plugins/auth/my_auth.c, is_local_connection), and the hostname check is added only for peers it considers non-local. 127.0.0.1, ::1, and socket transports never get it, so the canonical Rails-style config -- verify_identity against a local address or a tunneled port -- reported success with the server's identity never checked: X509_check_host never executed. Confirmed by sodabrew's CI evidence and MacPorts repro on #879 and corroborated from a Connector/C 3.4.9 source read on the same thread. Worse, a failed HOST/TRUST verification with no CA configured falls through to the 3.4 self-signed leniency path (my_auth.c) and the connection can still succeed, so even a forced hostname check cannot fail closed without a CA.

This parcel fulfills the #1480/#879 commitments as one cohesive security change, four parts:

1. Enforcement (the core). On MariaDB Connector/C 3.4+ builds, :verify_identity registers a TLS verification callback (MARIADB_OPT_TLS_VERIFICATION_CALLBACK, CONC-724) that replaces the connector's default verifier and verifies the session directly with OpenSSL: chain trust via the handshake's recorded status plus SSL_get_verify_result, hostname via X509_check_host/X509_check_ip_asc (real SAN, wildcard, and IP-address matching), unconditionally -- local peers included -- inside the handshake, before credentials are sent. Delegating to the connector's own ma_pvio_tls_verify_server_cert with forced HOST|TRUST flags would have been simpler, but distro builds hide that symbol behind the version script's local:* (verified against Arch's libmariadb.so.3: not in the dynamic symbol table), so the callback links only public API. It reaches the session through the installed MARIADB_PVIO struct plus a mirrored 3-pointer MARIADB_TLS head (the installed ma_tls.h can't be included: it pulls in the not-installed ma_hash.h), validates that layout via pointer round-trips and SSL app_data identity before trusting it, and fails closed on any mismatch. On failure it also forces MARIADB_TLS_VERIFY_ERROR into net.tls_verify_status so the caller's no-CA leniency path cannot resurrect the connection, and the refusal names the specifics -- the hostname checked, the CA in effect, OpenSSL's chain-failure reason -- plus the remedy, since this error is what a misconfigured caller actually sees. Both option registrations are checked and refuse at configuration time if the runtime client library rejects them: enforcement must not silently demote to a weaker layer. A post-connect tripwire in rb_mysql_connect then re-proves chain+hostname against the live session and refuses (close + raise) any connection where enforcement was promised but verification cannot be shown to have run -- e.g. a connect that never negotiated TLS, the class of silent skip #879 is about.

2. Fail-closed configuration. :sslca/:sslcapath left unset means the TLS backend resolves its default trust store natively (OpenSSL's default verify paths, SSL_CERT_FILE/SSL_CERT_DIR, or the platform certificate store); mysql2 neither guesses system CA paths client-side nor pre-flights their existence. The connect attempt is the source of truth: under :verify_identity the verification callback and the post-connect tripwire refuse any connection whose chain no trust store vouches for, rather than letting it fall through to the connector's no-CA self-signed leniency. All of the pre-connect refusals in this change raise Mysql2::Error::ConnectionError -- the class the equivalent SSL/CA failures have always raised from the C connect path -- so code narrowly rescuing ConnectionError keeps seeing these as connection failures. On MariaDB builds where the callback is unavailable (Connector/C < 3.4, or no OpenSSL headers at gem build time), :verify_identity raises a clear error instead of connecting -- connecting with the check silently skipped is exactly the #879 bug class. The MariaDB-family version gate is open-ended (any Connector/C 3.x+ or 10.x+ server library, not a closed version window), so a future Connector/C major version lands on this enforce-or-refuse pair instead of silently regressing to the unverified path. Mysql2::Client::TLS_PEER_IDENTITY_VERIFICATION reports the mechanism in effect (:native / :callback / nil).

3. Pinning passthrough (#1480). :tls_peer_fingerprint and :tls_peer_fingerprint_list map to MARIADB_OPT_TLS_PEER_FP/_LIST for CA-less pinned TLS. Unsupported builds raise rather than connect unpinned (a pin not applied is a control silently dropped -- deliberate departure from the warn-and-degrade additive model of #1405, for the same fail-closed reason as above). Combining pinning with a verifying ssl_mode raises: the connector runs the FINGERPRINT rung instead of HOST|TRUST, so one of the two requested models would silently not run.

4. Introspection (#1480). Client#tls_info returns the negotiated protocol and cipher, the peer certificate actually presented (subject, issuer, validity, SHA-256 fingerprint), the connector's recorded verify-status bitmask (decodable via new TLS_VERIFY_* constants), and :identity_verified -- whether mysql2's enforcement confirmed chain+hostname for this connection. nil for non-TLS connections and on builds without the API (libmysqlclient, Connector/C < 3.4); on a returned hash every key is present, nil for anything the library could not report. Note verify-status records failures among checks that ran: a CA-less pinned connection legitimately carries TLS_VERIFY_TRUST (the pin, not the chain, is its trust anchor), and a clean status alone cannot prove the hostname rung executed -- that is what :identity_verified is for.

The #879 spec matrix: MariaDB mismatch-by-IP refusal (the regression leg: previously connected), mismatch-by-name refusal (guards the connector's own non-local-name topology), match-accepted with tls_info proving which rung ran, MySQL-tier refusal parity, fingerprint pin accept/refuse legs, option-validation legs (pin+verify mutual exclusion, unsupported-build raises), and a CA-less leg proving verify_identity fails closed at connect time -- not via any Client.new pre-flight -- when no CA is configured. Environment-gated like the existing SSL specs; the mismatch-by-name leg needs a wrong.mysql2gem.example.com hosts alias, added to CI alongside the existing one, and skips where unresolvable. The mismatch-by-IP leg supersedes the regression spec from #1568, which fails on real-certificate MariaDB CI legs until this enforcement lands.

Behavior changes, loudly:

* Connections that previously "succeeded" under ssl_mode: :verify_identity on MariaDB with the hostname never verified are now REFUSED unless the certificate actually covers the connected hostname. That includes 127.0.0.1/tunnel topologies -- connect via a name the certificate covers, or pin the certificate instead. This is the fix, not a side effect.
* Unset :sslca is no longer replaced by a guessed system bundle path: the find_default_ca_path probe of four hardcoded paths is no longer called (the helper remains, uncalled, for API compatibility), and the TLS backend's own default trust-store resolution applies instead.
* :verify_identity on MariaDB builds that cannot enforce it now raises (previously: silent CA-only verification at best).
* :verify_identity on MariaDB now implies a required TLS connection (MYSQL_OPT_SSL_ENFORCE), as it always did on MySQL.

Verification: full suite green on Linux/MariaDB 11.8 server with Connector/C 3.4.9 (the exact version from the #879 repro) and on macOS/libmysqlclient; TLS matrix additionally green against MySQL 8.4 with libmysqlclient 8.0.46 on Ubuntu 24.04. Specs verified to bite on deliberately broken builds: with the callback registration removed the post-connect tripwire still refuses the mismatch legs, and with both layers removed the mismatch-by-IP leg reproduces #879 (connects) and fails the suite. An instrumented build confirms the callback executes during the handshake, pre-auth. The CA-less verify_identity leg verified on both tiers: Connector/C 3.4 refuses via the callback ("certificate verification failed (self-signed certificate in certificate chain; CA in effect: none)") and libmysqlclient 8.0.46 refuses natively ("CA certificate is required if ssl-mode is VERIFY_CA or VERIFY_IDENTITY"), both as Mysql2::Error::ConnectionError.

Coverage limits: the callback verifies via OpenSSL only -- on Schannel/GnuTLS Connector/C builds it fails closed (refuses) rather than verifying, and Windows behavior (including the vendored MSYS2 build's TLS backend and the named-pipe transport, where the connector itself disables TLS) is compile-gated by the same extconf probes but untested here; CI's Windows legs skip the SSL context entirely, a pre-existing blind spot flagged on #879. The MARIADB_TLS head mirror is validated at runtime but is still a layout assumption against installed headers, re-verified for Connector/C 3.4.x and 12.3-bundled headers.

### DIFF
--- a/.github/workflows/build-fedora.yml
+++ b/.github/workflows/build-fedora.yml
@@ -20,8 +20,8 @@ jobs:
           # (nothing's been dnf-installed yet at that point in the job), so
           # these rows test whatever Ruby version Fedora currently packages
           # rather than a chosen version.
-          - {os: ubuntu-24.04, ruby: 'system', db: 'fedora:latest', container: {image: 'fedora:latest', options: '--cap-add=SYS_PTRACE --security-opt seccomp=unconfined --add-host=mysql2gem.example.com:127.0.0.1'}}
-          - {os: ubuntu-24.04, ruby: 'system', db: 'fedora:rawhide', container: {image: 'fedora:rawhide', options: '--cap-add=SYS_PTRACE --security-opt seccomp=unconfined'}, ssl_cert_dir: '/tmp/mysql2', ssl_cert_host: 'localhost'}
+          - {os: ubuntu-24.04, ruby: 'system', db: 'fedora:latest', container: {image: 'fedora:latest', options: '--cap-add=SYS_PTRACE --security-opt seccomp=unconfined --add-host=mysql2gem.example.com:127.0.0.1 --add-host=wrong.mysql2gem.example.com:127.0.0.1'}}
+          - {os: ubuntu-24.04, ruby: 'system', db: 'fedora:rawhide', container: {image: 'fedora:rawhide', options: '--cap-add=SYS_PTRACE --security-opt seccomp=unconfined --add-host=mysql2gem.example.com:127.0.0.1 --add-host=wrong.mysql2gem.example.com:127.0.0.1'}, ssl_cert_dir: '/tmp/mysql2'}
       fail-fast: false
     env:
       BUNDLE_WITHOUT: development
--- a/.github/workflows/build-macos.yml
+++ b/.github/workflows/build-macos.yml
@@ -26,7 +26,7 @@ jobs:
         with:
           ruby-version: ${{ matrix.ruby }}
           bundler-cache: true # runs 'bundle install' and caches installed gems automatically
-      - run: sudo echo "127.0.0.1 mysql2gem.example.com" | sudo tee -a /etc/hosts
+      - run: sudo echo "127.0.0.1 mysql2gem.example.com wrong.mysql2gem.example.com" | sudo tee -a /etc/hosts
       - run: echo 'DB=${{ matrix.db }}' >> $GITHUB_ENV
       - run: bash ci/setup.sh
       # Set the verbose option in the Makefile to print compiling command lines.
--- a/.github/workflows/build-ubuntu.yml
+++ b/.github/workflows/build-ubuntu.yml
@@ -60,7 +60,7 @@ jobs:
         with:
           ruby-version: ${{ matrix.ruby }}
           bundler-cache: true # runs 'bundle install' and caches installed gems automatically
-      - run: sudo echo "127.0.0.1 mysql2gem.example.com" | sudo tee -a /etc/hosts
+      - run: sudo echo "127.0.0.1 mysql2gem.example.com wrong.mysql2gem.example.com" | sudo tee -a /etc/hosts
       - run: echo 'DB=${{ matrix.db }}' >> $GITHUB_ENV
       - run: bash ci/setup.sh
       # Set the verbose option in the Makefile to print compiling command lines.
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -39,7 +39,7 @@ Metrics/BlockNesting:
 # Offense count: 1
 # Configuration parameters: CountComments, CountAsOne.
 Metrics/ClassLength:
-  Max: 153
+  Max: 164
 
 # Offense count: 3
 # Configuration parameters: AllowedMethods, AllowedPatterns.
--- a/README.md
+++ b/README.md
@@ -331,20 +331,71 @@ the `:sslverify` boolean. For details on each of the `:ssl_mode` options, see
 
 The `:ssl_mode` option will also set the appropriate MariaDB connection flags:
 
-| `:ssl_mode`        | MariaDB option value                 |
-| ---                | ---                                  |
-| `:disabled`        | MYSQL_OPT_SSL_ENFORCE = 0            |
-| `:required`        | MYSQL_OPT_SSL_ENFORCE = 1            |
-| `:verify_ca`       | MYSQL_OPT_SSL_VERIFY_SERVER_CERT = 1 |
-| `:verify_identity` | MYSQL_OPT_SSL_VERIFY_SERVER_CERT = 1 |
+| `:ssl_mode`        | MariaDB option value                                |
+| ---                | ---                                                 |
+| `:disabled`        | MYSQL_OPT_SSL_ENFORCE = 0                           |
+| `:required`        | MYSQL_OPT_SSL_ENFORCE = 1                           |
+| `:verify_ca`       | MYSQL_OPT_SSL_VERIFY_SERVER_CERT = 1                |
+| `:verify_identity` | MYSQL_OPT_SSL_VERIFY_SERVER_CERT = 1 + mysql2's own hostname verification |
 
-MariaDB Connector/C has no CA-only verification mode, so `:verify_ca` gets
-the same full verification (CA and hostname) that `:verify_identity` gets --
-MYSQL_OPT_SSL_VERIFY_SERVER_CERT is the only verification it offers. MariaDB
-does not support the `:preferred` option; there's no equivalent to fall back
-to. For more information about SSL/TLS in MariaDB, see
+MYSQL_OPT_SSL_VERIFY_SERVER_CERT verifies the CA at most: despite its
+MySQL-derived name, MariaDB Connector/C decides whether to check the
+hostname from the peer address, and never checks it for peers it considers
+local (127.0.0.1, ::1, or socket connections). mysql2 enforces
+`:verify_identity` on MariaDB itself, by registering a TLS
+verification callback (MariaDB Connector/C 3.4+) that verifies the
+certificate chain and the hostname (SAN, wildcard, and IP-address matching
+via OpenSSL) on every connection, local peers included.
+
+Two things worth knowing:
+
+* With `:sslca`/`:sslcapath` left unset, the TLS backend resolves its
+  default trust store natively (OpenSSL's default verify paths,
+  `SSL_CERT_FILE`/`SSL_CERT_DIR`, or the platform certificate store) --
+  publicly-CA-signed servers verify with no CA option passed. Under
+  `:verify_identity`, an unverifiable chain is refused at connect time.
+* If the mysql2 build cannot enforce the hostname check (MariaDB
+  Connector/C older than 3.4, or no OpenSSL headers at build time),
+  `ssl_mode: :verify_identity` raises instead of connecting.
+  `Mysql2::Client::TLS_PEER_IDENTITY_VERIFICATION` reports the mechanism
+  in effect: `:native` (libmysqlclient), `:callback` (MariaDB + mysql2's
+  verification callback), or `nil` (unenforceable).
+
+MariaDB does not support the `:preferred` option; there's no equivalent to
+fall back to. For more information about SSL/TLS in MariaDB, see
 [https://mariadb.com/kb/en/securing-connections-for-client-and-server/](https://mariadb.com/kb/en/securing-connections-for-client-and-server/)
 and [https://mariadb.com/kb/en/mysql_optionsv/#tls-options](https://mariadb.com/kb/en/mysql_optionsv/#tls-options)
+
+On MariaDB Connector/C 3.4+, the server certificate can be pinned by
+fingerprint instead of a CA -- an alternative trust model to
+`:verify_ca`/`:verify_identity` (the connector runs the fingerprint check
+*instead of* the CA/hostname checks, so combining them raises):
+
+``` ruby
+Mysql2::Client.new(
+  # ...options as above...,
+  :tls_peer_fingerprint => 'c3ab8ff13720e8ad9047dd39466b3c8974e592c2fa383d4a3960714caef0c4f2', # SHA-256 hex
+  # or a file of acceptable fingerprints, one per line:
+  :tls_peer_fingerprint_list => '/path/to/fingerprints',
+  )
+```
+
+On client libraries without pinning support (libmysqlclient, MariaDB
+Connector/C before 3.4) these options raise rather than silently
+connecting unpinned.
+
+After connecting, `Client#tls_info` describes the TLS session as the
+client library observed it: negotiated protocol and cipher, the peer
+certificate the server actually presented (subject, issuer, validity,
+SHA-256 fingerprint), the bitmask of verification checks the connector
+recorded as failed (`:verify_status`, decodable with the
+`Mysql2::Client::TLS_VERIFY_*` constants -- note a CA-less pinned
+connection legitimately carries `TLS_VERIFY_TRUST`, since the pin rather
+than the chain is its trust anchor), and whether mysql2's
+`:verify_identity` enforcement confirmed chain and hostname verification
+for this connection (`:identity_verified`). It returns `nil` for non-TLS
+connections and on client libraries without the introspection API
+(libmysqlclient, MariaDB Connector/C before 3.4).
 
 To set the TLS Server Name Indication (SNI) hostname sent during the TLS
 handshake, e.g. when connecting through a proxy that routes by hostname:
--- a/ext/mysql2/client.c
+++ b/ext/mysql2/client.c
@@ -17,7 +17,7 @@
 #include "mysql_enc_name_to_ruby.h"
 
 VALUE cMysql2Client;
-extern VALUE mMysql2, cMysql2Error, cMysql2TimeoutError;
+extern VALUE mMysql2, cMysql2Error, cMysql2ConnectionError, cMysql2TimeoutError;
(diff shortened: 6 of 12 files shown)


sodabrew (changes requested):
Reviewed the full diff, including an empirical check of the callback approach against a real MariaDB 11.8.8 / Connector-C 3.4.9 build (the same combination from the original #879 repro). This is careful, well-diagnosed work -- the root-cause writeup (connector's local-peer heuristic, not a missing flag) matches what I independently confirmed, the fail-closed design throughout is the right call, the struct-layout mirror is validated defensively rather than trusted blindly, and the post-connect tripwire is a genuinely good piece of defense-in-depth (it's independent of whether the callback ran at all, so it also covers session-resumption-shaped edge cases). Spec coverage is thorough.

One change requested, plus two items to settle explicitly before merge (not code defects, decisions).

## Requested change: stop having mysql2 guess CA paths; let the connect attempt (and the existing post-connect tripwire) be the source of truth

`configure_tls_verification` now hard-raises at `Client.new` if `:sslca`/`:sslcapath` is unset and `find_default_ca_path`'s four hardcoded paths (`/etc/ssl/certs/ca-certificates.crt`, `/etc/pki/tls/certs/ca-bundle.crt`, `/etc/ssl/ca-bundle.pem`, `/etc/ssl/cert.pem`) don't exist. That helper predates this PR (`d136164`, 2021, "Setup default CA path if not provided") and was always a soft convenience default -- if it found nothing, `:sslca` just stayed unset and OpenSSL fell back to its own default trust-store resolution (`SSL_CTX_set_default_verify_paths`, `SSL_CERT_FILE`/`SSL_CERT_DIR`, or libmysqlclient's native Windows cert-store path). This PR turns "found nothing, let OpenSSL handle it" into "found nothing, refuse to connect" -- before OpenSSL gets a chance to try. That's a real regression for a very common case: `ssl_mode: :verify_identity` against a publicly-CA-signed server (RDS, Cloud SQL, etc.) with no `:sslca` ever passed, relying entirely on the system trust store -- and it hits both MySQL and MariaDB, not just the MariaDB path this PR is otherwise scoped to. It also guarantees a hard failure on Windows (none of the four paths exist there, and CI's Windows legs skip the SSL context, so this wouldn't be caught before release).

Requesting: don't encode mysql2's own list of system CA paths at all, and don't pre-flight-check for one. Either the caller passes an explicit `:sslca`/`:sslcapath`, or nothing is set and the system default is used as OpenSSL/libmysqlclient would resolve it natively. Let the actual connect attempt happen, and lean on the verification you've already built -- the post-connect tripwire this PR adds -- to prove the connection was genuinely validated before handing it back, rather than trying to predict from a client-side path guess whether it will be. Concretely: drop the `find_default_ca_path` call and its raise from `configure_tls_verification` entirely; the tripwire already fails closed on anything that didn't actually verify, which is the authoritative check anyway.

## Please don't change the error class for these raises

The new validation raises (CA-required, fingerprint/verify_mode mutual exclusion, "cannot be enforced") use plain `Mysql2::Error`. Existing SSL/CA failures have always surfaced as `Mysql2::Error::ConnectionError` (raised from the C `connect()` call). Please keep these on `Mysql2::Error::ConnectionError` (or whatever the pre-#1570 equivalent failure raised) rather than the bare parent class, so code that narrowly rescues `ConnectionError` to distinguish connection failures from other errors isn't silently broken by the switch to earlier, Ruby-side validation.

## Not blocking, flagging for the release decision

This is an intentional, clearly-disclosed breaking change: `ssl_mode: :verify_identity` connections that previously "succeeded" against a local/tunneled peer with the hostname never checked will now be refused. Given mysql2 is heading toward a major release, that's an acceptable tradeoff for a real security fix -- but it's sodabrew's call on packaging (major-version timing, CHANGELOG/migration note, whether to fold it in with other breaking changes already queued for the same release) rather than something to resolve in this review. Flagging it here so it's an explicit decision rather than an implicit one.

## Not a concern

The MARIADB_PVIO/`mysql2_mariadb_tls_head` struct-layout mirror is inherently version-fragile by nature, but it's the right kind of fragile: validated at runtime via pointer round-trips before being trusted, and fails closed (refuses) rather than misbehaving on a mismatch. Only tested against Connector/C 3.4.9 specifically rather than the full 3.4.x range, but that's fine to leave as-is for now -- if a specific packaged version turns out to disagree, that's a `.1` patch fix, not something to hold up this PR for.

jeremy:
@sodabrew Both requested changes landed, amended in place (11439c8f075de70a231a7a92374b9d9e97b46b96, still one commit).

**CA pre-flight dropped.** `configure_tls_verification` no longer calls `find_default_ca_path` and no longer raises on unset `:sslca`/`:sslcapath` — unset stays unset, no path guessing at all, and the backend's native default trust-store resolution applies. Before accepting that the connect attempt plus tripwire could carry the guarantee alone, I proved the no-CA case fails closed on the review's own combination (MariaDB 11.8 / Connector/C 3.4): CA-less `verify_identity` against the suite's self-signed CA refuses inside the handshake via the callback — `Mysql2::Error::ConnectionError`, "certificate verification failed (self-signed certificate in certificate chain; CA in effect: none); pass the CA that signed the server certificate as :sslca" — and on libmysqlclient 8.0.46 the library refuses natively ("CA certificate is required if ssl-mode is VERIFY_CA or VERIFY_IDENTITY"). A new spec leg pins this down: CA-less `verify_identity` fails closed at connect time, not via any `Client.new` pre-flight. You were right about Windows too — none of the four probed paths exist there, so the pre-flight guaranteed a raise the SSL-skipping CI legs would never have caught. `find_default_ca_path` itself predates this PR (d136164), so the method stays defined, now uncalled; delete it here or in release housekeeping as you prefer. One data point past the PR's scope: CA-less `verify_ca` on this Connector/C build also refuses at handshake ("self-signed certificate in certificate chain") rather than reaching the no-CA leniency path — though `verify_ca` has no tripwire, so that leg rests on the connector's own behavior, same as pre-PR.

**Error class.** Confirmed the pre-#1570 equivalent: SSL/CA failures surface from the C connect path as CR_SSL_CONNECTION_ERROR (2026), which `Error::CODES` maps to `Mysql2::Error::ConnectionError`. Every pre-connect refusal this PR adds now raises `ConnectionError`: the fingerprint/verify_mode mutual exclusion, the "cannot be enforced" raises, and the fingerprint-unsupported raises (new options, but the same failure class — kept consistent). The callback and tripwire refusals already carried errno 2026 through `Error.new_with_args`, so they were `ConnectionError` from the start. Specs now assert the subclass on every leg, mismatch refusals included.

**Release packaging**: your call, agreed. The migration note is the PR description's "Behavior changes, loudly" section (CHANGELOG.md is a pointer to GitHub Releases) — updated to match the above, ready to lift into release notes wherever you slot this.

**Struct mirror**: agreed on the framing — version-fragile by nature, validated before trusted, refuses on mismatch; a packaged version that disagrees is a `.1` fix.

And thanks for the empirical re-check — your #879 follow-up confirming the shipped two-arg contract (`(MARIADB_TLS *, unsigned int)` by value, no third argument) against the same MacPorts build that crashed the commit-message signature is what makes the callback approach safe to rest on.

Suite re-verified green on MariaDB 11.8 server / Connector/C 3.4 client and on libmysqlclient 8.0.46 / MySQL 8.4; full matrix running in CI on the amended sha.

Ready for re-review (I can't formally re-request from this side of the fork).

sodabrew (approved):
Both requested changes verified directly in the diff, not just the description:

- **CA pre-flight**: `configure_tls_verification` no longer calls `find_default_ca_path` or raises on an unset CA -- that path is gone entirely, not softened. The new "fails closed at connect time, not Client.new, when no CA is configured" spec proves the CA-less case is now caught by the actual connect attempt (callback + tripwire), not a client-side guess. `find_default_ca_path` is left defined but unused, as offered.
- **Error class**: every new raise this PR adds -- the fingerprint/verify_mode mutual exclusion, the "cannot be enforced" raises (C and Ruby), the fingerprint-unsupported raises -- now raises `Mysql2::Error::ConnectionError`, matching the pre-existing `CR_SSL_CONNECTION_ERROR` mapping.

CI green across the full matrix on the amended commit. Approving.

sodabrew:
Follow-up on the README section specifically (code/tests/design already approved and unaffected by this).

README should describe the gem's current behavior, full stop -- not issue numbers or before/after framing. A reader new to the project has no "previously" to compare against, and doesn't need to chase `#879` to understand what `:verify_identity` does today; that context belongs in the commit message, not the docs. Found four spots in this section that read as changelog/migration-note material rather than a description of current behavior:

1. `"...so on its own it cannot deliver what :verify_identity promises (see issue #879)."` -- drop the issue pointer, just state the fact.
2. `"...connecting with the check silently skipped is exactly the failure mode of #879."` -- same; justifies the design by its bug history instead of stating current behavior.
3. The `"Note that this means connections that previously 'succeeded' ... are now refused"` paragraph -- pure migration-note framing, and redundant once (1)/(2) are fixed: the opening paragraph already states `:verify_identity` verifies the hostname on every connection, local peers included, which already tells a new reader what will and won't connect.
4. Minor: `"Two consequences worth knowing:"` -- "consequences" frames the bullets as fallout of a change; `"Two things worth knowing:"` describes the same two facts without implying a before/after.

Suggested diff:

```diff
 MYSQL_OPT_SSL_VERIFY_SERVER_CERT verifies the CA at most: despite its
 MySQL-derived name, MariaDB Connector/C decides whether to check the
 hostname from the peer address, and never checks it for peers it considers
-local (127.0.0.1, ::1, or socket connections) -- so on its own it cannot
-deliver what `:verify_identity` promises (see issue #879). mysql2 therefore
-enforces `:verify_identity` on MariaDB itself, by registering a TLS
+local (127.0.0.1, ::1, or socket connections). mysql2 enforces
+`:verify_identity` on MariaDB itself, by registering a TLS
 verification callback (MariaDB Connector/C 3.4+) that verifies the
 certificate chain and the hostname (SAN, wildcard, and IP-address matching
 via OpenSSL) on every connection, local peers included.

-Two consequences worth knowing:
+Two things worth knowing:

 * With `:sslca`/`:sslcapath` left unset, the TLS backend resolves its
   default trust store natively (OpenSSL's default verify paths,
   `SSL_CERT_FILE`/`SSL_CERT_DIR`, or the platform certificate store) --
-  publicly-CA-signed servers verify with no CA option passed. Under
-  `:verify_identity`, whether verification actually succeeded is proven at
-  connect time: a chain that no trust store vouches for is refused rather
-  than falling through to MariaDB's no-CA self-signed leniency.
+  publicly-CA-signed servers verify with no CA option passed. Under
+  `:verify_identity`, an unverifiable chain is refused at connect time.
 * If the mysql2 build cannot enforce the hostname check (MariaDB
   Connector/C older than 3.4, or no OpenSSL headers at build time),
-  `ssl_mode: :verify_identity` raises instead of connecting -- connecting
-  with the check silently skipped is exactly the failure mode of #879.
+  `ssl_mode: :verify_identity` raises instead of connecting.
   `Mysql2::Client::TLS_PEER_IDENTITY_VERIFICATION` reports the mechanism
   in effect: `:native` (libmysqlclient), `:callback` (MariaDB + mysql2's
   verification callback), or `nil` (unenforceable).

-Note that this means connections that previously "succeeded" under
-`ssl_mode: :verify_identity` on MariaDB without their hostname ever being
-verified are now refused unless the certificate actually matches.
-
 MariaDB does not support the `:preferred` option; there's no equivalent to
 fall back to. For more information about SSL/TLS in MariaDB, see
```

The migration-note content isn't lost -- it's already covered by the PR's own "Behavior changes, loudly" section, which is the right home for it.

jeremy:
README fixed with your diff as-is — all four spots, amended in place (01a22db, still one commit). Agreed on the principle: the README describes what `:verify_identity` does today; the before/after story stays in the commit message and the PR's "Behavior changes, loudly" section for release notes.

While amending I also rebased onto master — the branch had gone conflicting after the recent merges. One trivial conflict (adjacent new fields in the client wrapper struct in client.h); range-diff against the approved sha shows context drift only, nothing in the reviewed diff changed. Suite re-verified green at the new sha on both tiers: MariaDB 11.8 server / Connector/C 3.4 client with the full #879 and fingerprint legs running, and libmysqlclient 8.0.46 / MySQL 8.4.

PR #1572, opened 2026-08-16
Add :tls_version option and map :sslverify onto :ssl_mode

### LINKED ISSUE #973: It is not possible to explicitly set the SSL/TLS protocol
I need to set the TLS version to TLSv1.2 via the MYSQL_OPT_TLS_VERSION MySQL connection option.

https://dev.mysql.com/doc/refman/5.7/en/mysql-options.html
https://dev.mysql.com/doc/refman/5.7/en/encrypted-connection-protocols-ciphers.html

Is this possible with Mysql2?  Maybe with :connect_attrs?

sodabrew:
Rebased directly onto `master` now that #1570 has merged -- no longer stacked on anything, diff is just this PR's own commits.

## Fixes #973

`MYSQL_OPT_TLS_VERSION` / `MARIADB_OPT_TLS_VERSION` (the same enum member, both MySQL 5.7.10+ and MariaDB Connector/C 3.4.3+ handle it identically per Connector/C's own source) had no mysql2-facing option. Requested in #973 since 2020.

New `:tls_version` option, wired through the existing `_mysql_client_options()` dispatcher the same way `default_auth`/`init_command`/etc. already are. Takes a comma-separated protocol list, e.g. `tls_version: 'TLSv1.2,TLSv1.3'`. Gated on `have_const('MYSQL_OPT_TLS_VERSION')`; raises `Mysql2::Error` on older client libraries rather than silently ignoring the option, matching the `tls_sni_name` precedent (#1479). Exposes `Mysql2::Client::TLS_VERSION_SUPPORTED`.

## Also: map legacy `:sslverify` onto `:ssl_mode`

`:sslverify => true` sets `CLIENT_SSL_VERIFY_SERVER_CERT` directly as a `mysql_real_connect()` connect-flag. MySQL's client library reads that flag during the handshake -- it's literally what `ssl_mode: :verify_identity` sets internally on MySQL, so `:sslverify` works there. MariaDB Connector/C never reads that flag anywhere in its connect path (confirmed by grepping `mariadb_lib.c` end to end): `:sslverify => true` was a complete, silent no-op on MariaDB, with none of the enforcement #1570 built for `ssl_mode: :verify_identity`. Landed here rather than as a separate PR since the README changes above already document (and now need to stop describing as a limitation) exactly this gap.

`configure_tls_verification` now fills in `ssl_mode: :verify_identity` from `:sslverify => true` when the caller didn't give an explicit `:ssl_mode` -- so the legacy option gets the same enforcement (and the same fail-closed guarantees, including refusing to connect on builds that can't enforce it) `ssl_mode: :verify_identity` already has. An explicit `:ssl_mode` always wins over `:sslverify`. Conversely, `:sslverify => false` alongside a verifying `ssl_mode` (`:verify_ca` or `:verify_identity`) now raises `Mysql2::Error::ConnectionError` before any connection attempt, rather than silently picking a side -- the two options would otherwise directly contradict each other. Follows the option-coherence precedent #1555 established, though #1555's own checks are warn-only; this one raises since it's a hard contradiction of stated intent, not a silent-degradation case.

## Verification

Since this machine's `mysqld` doesn't serve `spec/ssl/*`, I verified `:tls_version` directly with two standalone scripts against real, separately-configured servers rather than through the spec suite alone:

- **MySQL 8.0.34** (local): requesting `TLSv1.2` and `TLSv1.3` each genuinely negotiated that exact version, confirmed via `SHOW STATUS LIKE 'Ssl_version'`.
- **MariaDB 11.8.8 / Connector-C 3.4.9** (MacPorts, isolated instance configured with this repo's own `spec/ssl/*` fixtures): same result -- both versions negotiated correctly, and requesting an unsupported version (`TLSv1.1`, which the server had disabled) was cleanly rejected.

For `:sslverify`: six option combinations verified at the logic level against a `Klient` subclass (already in the spec file) that records connect args without dialing out. Live end-to-end against MySQL 8.0.34: `sslverify: true` now genuinely fails closed against a self-signed local server (`CA certificate is required if ssl-mode is VERIFY_CA or VERIFY_IDENTITY`), matching `ssl_mode: :verify_identity`'s own behavior; `sslverify: false` + a verifying `ssl_mode` raises before any connection attempt. **Could not do live end-to-end verification against MariaDB** -- blocked by a pre-existing, unrelated crash filed separately as #1575 (dual-OpenSSL-linking segfault in #1570's `MYSQL2_VERIFY_IDENTITY_SHIM`). Confirmed that crash reproduces identically with plain `ssl_mode: :verify_identity` and zero `:sslverify` involved, so it isn't specific to this change.

Full suite (`bundle exec rspec`, `bundle exec rubocop`) green -- 605 examples, 0 failures, 23 pending (all pre-existing environment skips). `.rubocop_todo.yml`'s `Metrics/ClassLength` bumped 164 -> 170 by hand for the new lines, matching the precedent in #1482/#1555.

New spec (`spec/mysql2/client_spec.rb`, in the `SSL` context) mirrors the `tls_sni_name` spec's shape for `:tls_version`: accepts-and-negotiates when supported, raises cleanly when not. One thing this caught along the way, worth noting: the first version of the `:tls_version` spec merged `option_overrides`, which hardcodes `:sslcipher => 'DHE-RSA-AES256-SHA'` (a TLS 1.2-only cipher name) -- forcing that while restricting to `tls_version: 'TLSv1.3'` is a real, self-inflicted handshake failure on strict OpenSSL builds (caught by CI, not locally). Fixed by dropping `:sslcipher` for this spec specifically; not a bug in the feature itself.

### DIFF
--- a/README.md
+++ b/README.md
@@ -368,9 +368,8 @@ fall back to. For more information about SSL/TLS in MariaDB, see
 [https://mariadb.com/kb/en/securing-connections-for-client-and-server/](https://mariadb.com/kb/en/securing-connections-for-client-and-server/)
 and [https://mariadb.com/kb/en/mysql_optionsv/#tls-options](https://mariadb.com/kb/en/mysql_optionsv/#tls-options)
 
-`:sslverify` still works on MySQL 8.0+ (it's what `ssl_mode: :verify_identity`
-sets internally there) but is a complete no-op on MariaDB Connector/C -- the
-connect-flag it sets is never read there -- so prefer `:ssl_mode` on MariaDB.
+On MySQL 8.0+, `:sslverify` is identical to `ssl_mode: :verify_identity`.
+On MariaDB Connector/C, only the :ssl_mode flag is available.
 
 On MariaDB Connector/C 3.4+, the server certificate can be pinned by
 fingerprint instead of a CA -- an alternative trust model to
--- a/README.md
+++ b/README.md
@@ -413,9 +413,8 @@ Mysql2::Client.new(
 ```
 
 A comma-separated list of TLS versions, e.g. `'TLSv1.2,TLSv1.3', that are available on your SSL client library. Works
-identically on MySQL (5.7.10+) and MariaDB Connector/C (3.4.3+); raises
-`Mysql2::Error` on older client libraries rather than silently ignoring the
-option.
+identically on MySQL (5.7.10+) and MariaDB Connector/C (3.4.3+). raises `Mysql2::Error` on older client libraries rather than silently
+ignoring the option.
 
 To set the TLS Server Name Indication (SNI) hostname sent during the TLS
 handshake, e.g. when connecting through a proxy that routes by hostname:
--- a/README.md
+++ b/README.md
@@ -414,10 +414,10 @@ Mysql2::Client.new(
   )
 ```
 
-A comma-separated list of TLS versions, e.g. `'TLSv1.2,TLSv1.3', that are available on your SSL client library. Works
-identically on MySQL (5.7.10+) and MariaDB Connector/C (3.4.3+).
-Raises `Mysql2::Error` on older client libraries rather than silently
-ignoring the option.
+A comma-separated list of TLS versions available on your SSL client library,
+e.g. `'TLSv1.2,TLSv1.3'`. Works identically on MySQL (5.7.10+) and MariaDB
+Connector/C (3.4.3+). Raises `Mysql2::Error` on older client libraries rather
+than silently ignoring the option.
 
 To set the TLS Server Name Indication (SNI) hostname sent during the TLS
 handshake, e.g. when connecting through a proxy that routes by hostname:
--- a/README.md
+++ b/README.md
@@ -416,7 +416,8 @@ Mysql2::Client.new(
 ```
 
 A comma-separated list of TLS versions, e.g. `'TLSv1.2,TLSv1.3', that are available on your SSL client library. Works
-identically on MySQL (5.7.10+) and MariaDB Connector/C (3.4.3+). raises `Mysql2::Error` on older client libraries rather than silently
+identically on MySQL (5.7.10+) and MariaDB Connector/C (3.4.3+).
+Raises `Mysql2::Error` on older client libraries rather than silently
 ignoring the option.
 
 To set the TLS Server Name Indication (SNI) hostname sent during the TLS
--- a/README.md
+++ b/README.md
@@ -362,12 +362,11 @@ Two things worth knowing:
   in effect: `:native` (libmysqlclient), `:callback` (MariaDB + mysql2's
   verification callback), or `nil` (unenforceable).
 
-`:sslverify => true` is equivalent to `ssl_mode: :verify_identity` on
-every client library mysql2 supports -- mysql2 maps it internally, so it
-carries the same enforcement described above on MariaDB rather than being
-a no-op there. An explicit `:ssl_mode` always takes precedence over
-`:sslverify`; passing `:sslverify => false` alongside a verifying
-`:ssl_mode` (`:verify_ca` or `:verify_identity`) raises, since the two
+`:sslverify => true` is equivalent to `ssl_mode: :verify_identity`.
+`:ssl_mode` is preferred as it has more nuanced options available.
+An explicit `:ssl_mode` takes precedence over `:sslverify`,
+however conflicting values, e.g. `:sslverify => false` alongside
+`:ssl_mode => :verify_ca` raises an exception, as the two
 would otherwise contradict each other.
 
 MariaDB does not support the `:preferred` option; there's no equivalent to
--- a/README.md
+++ b/README.md
@@ -412,8 +412,7 @@ Mysql2::Client.new(
   )
 ```
 
-A comma-separated list of one or more versions (`TLSv1.2`, `TLSv1.3`, etc. --
-availability depends on the SSL library the client was built against). Works
+A comma-separated list of TLS versions, e.g. `'TLSv1.2,TLSv1.3', that are available on your SSL client library. Works
 identically on MySQL (5.7.10+) and MariaDB Connector/C (3.4.3+); raises
 `Mysql2::Error` on older client libraries rather than silently ignoring the
 option.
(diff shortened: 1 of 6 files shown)


sodabrew:
Correction to the "Also investigated for this PR, but NOT included" section above: the segfault was my bug, not a defect in `MARIADB_OPT_TLS_VERIFICATION_CALLBACK` itself.

I used the 3-argument signature documented in the CONC-724 commit message (`MYSQL*, unsigned int*, my_bool`). The callback actually invokes with a different, 2-argument signature matching the connector's own internal verifier (`void *ctls, unsigned int flags -> int`). Registering the wrong-shaped function is what produced garbage arguments and crashed -- confirmed by testing the correct signature directly against the same MacPorts MariaDB 11.8.8 / Connector-C 3.4.9 build: no crash, sane arguments.

#879 has a real fix using the correct signature: **#1570** (Jeremy), which I've reviewed and requested changes on (unrelated to this signature question -- a CA-path-handling issue). Full correction posted on #879 itself.

This PR's own content -- `:tls_version` (#973) -- stands on its own and isn't affected; only the "investigated but not included" side note above was wrong. I'll update the `client.c` comment that repeats the same incorrect claim in a follow-up commit, and rebase this branch onto #1570 once it merges.

README.md
```diff
@@ -397,6 +403,21 @@ for this connection (`:identity_verified`). It returns `nil` for non-TLS
...
+
+``` ruby
+Mysql2::Client.new(
+  # ...options as above...,
+  :tls_version => 'TLSv1.2,TLSv1.3',
+  )
+```
+
+A comma-separated list of one or more versions (`TLSv1.2`, `TLSv1.3`, etc. --
+availability depends on the SSL library the client was built against). Works
```
sodabrew:
```suggestion
A comma-separated list of TLS versions, e.g. `'TLSv1.2,TLSv1.3', that are available on your SSL client library. Works
```

README.md
```diff
@@ -397,6 +403,20 @@ for this connection (`:identity_verified`). It returns `nil` for non-TLS
...
+Mysql2::Client.new(
+  # ...options as above...,
+  :tls_version => 'TLSv1.2,TLSv1.3',
+  )
+```
+
+A comma-separated list of TLS versions, e.g. `'TLSv1.2,TLSv1.3', that are available on your SSL client library. Works
+identically on MySQL (5.7.10+) and MariaDB Connector/C (3.4.3+); raises
+`Mysql2::Error` on older client libraries rather than silently ignoring the
+option.
```
sodabrew:
```suggestion
identically on MySQL (5.7.10+) and MariaDB Connector/C (3.4.3+). raises `Mysql2::Error` on older client libraries rather than silently
ignoring the option.
```

README.md
```diff
@@ -366,6 +368,10 @@ fall back to. For more information about SSL/TLS in MariaDB, see
 [https://mariadb.com/kb/en/securing-connections-for-client-and-server/](https://mariadb.com/kb/en/securing-connections-for-client-and-server/)
 and [https://mariadb.com/kb/en/mysql_optionsv/#tls-options](https://mariadb.com/kb/en/mysql_optionsv/#tls-options)
 
+`:sslverify` still works on MySQL 8.0+ (it's what `ssl_mode: :verify_identity`
+sets internally there) but is a complete no-op on MariaDB Connector/C -- the
+connect-flag it sets is never read there -- so prefer `:ssl_mode` on MariaDB.
```
sodabrew:
```suggestion
On MySQL 8.0+, `:sslverify` is identical to `ssl_mode: :verify_identity`.
On MariaDB Connector/C, only the :ssl_mode flag is available.
```

README.md
```diff
@@ -397,6 +406,19 @@ for this connection (`:identity_verified`). It returns `nil` for non-TLS
...
+
+``` ruby
+Mysql2::Client.new(
+  # ...options as above...,
+  :tls_version => 'TLSv1.2,TLSv1.3',
+  )
+```
+
+A comma-separated list of TLS versions, e.g. `'TLSv1.2,TLSv1.3', that are available on your SSL client library. Works
+identically on MySQL (5.7.10+) and MariaDB Connector/C (3.4.3+). raises `Mysql2::Error` on older client libraries rather than silently
```
sodabrew:
```suggestion
identically on MySQL (5.7.10+) and MariaDB Connector/C (3.4.3+).
Raises `Mysql2::Error` on older client libraries rather than silently
```

README.md
```diff
@@ -361,6 +362,14 @@ Two things worth knowing:
   in effect: `:native` (libmysqlclient), `:callback` (MariaDB + mysql2's
   verification callback), or `nil` (unenforceable).
 
+`:sslverify => true` is equivalent to `ssl_mode: :verify_identity` on
+every client library mysql2 supports -- mysql2 maps it internally, so it
+carries the same enforcement described above on MariaDB rather than being
+a no-op there. An explicit `:ssl_mode` always takes precedence over
+`:sslverify`; passing `:sslverify => false` alongside a verifying
+`:ssl_mode` (`:verify_ca` or `:verify_identity`) raises, since the two
+would otherwise contradict each other.
```
sodabrew:
```suggestion
`:sslverify => true` is equivalent to `ssl_mode: :verify_identity`.
`:ssl_mode` is preferred as it has more nuanced options available.
An explicit `:ssl_mode` takes precedence over `:sslverify`,
however conflicting values, e.g. `:sslverify => false` alongside
`:ssl_mode => :verify_ca` raises an exception, as the two
would otherwise contradict each other.
```

lib/mysql2/client.rb
```diff
@@ -133,7 +135,28 @@ def configure_tls_verification(opts, mode)
...
+          if (mode.nil? || mode.zero?) && !SSL_MODE_VERIFY_IDENTITY.zero?
+            mode = SSL_MODE_VERIFY_IDENTITY
+            verify_mode = true
+          end
+        elsif verify_mode
+          # :sslverify => false and a verifying ssl_mode contradict each
+          # other: one says the connection doesn't need to be verified, the
+          # other asks mysql2 to refuse it unless it is. Refuse the ambiguity
+          # instead of silently picking a side.
+          raise Mysql2::Error::ConnectionError, "sslverify: false conflicts with ssl_mode: #{opts[:ssl_mode]}, which requires verification; pick one"
```
sodabrew:
```suggestion
          raise Mysql2::Error::ConnectionError, "sslverify: false conflicts with ssl_mode: #{opts[:ssl_mode]}"
```

Issue #1575, opened 2026-08-17
ssl_mode: :verify_identity segfaults on MariaDB when mysql2.bundle and libmariadb link different OpenSSL builds


sodabrew:
## Summary

`ssl_mode: :verify_identity` on MariaDB Connector/C (`MYSQL2_VERIFY_IDENTITY_SHIM`, added in #1570) segfaults when `mysql2.bundle` and the linked `libmariadb` are built against two different OpenSSL installations. Reproduced deterministically, not a hypothetical.

## Environment

- macOS arm64 (darwin24), Ruby 3.3.7 installed via `mise` (ruby-build), which vendors its own OpenSSL for the `openssl` gem.
- MariaDB 11.8.8 / Connector-C 3.4.9 installed via MacPorts (`mariadb-11.8` port), linked against MacPorts' own `openssl3` port.
- mysql2 built with `--with-mysql-config=/opt/local/lib/mariadb-11.8/bin/mysql_config`.

## Root cause

```
$ otool -L lib/mysql2/mysql2.bundle | grep -i ssl
	/Users/aaron/.local/share/mise/installs/ruby/3.3.7/openssl/lib/libssl.3.dylib
	/Users/aaron/.local/share/mise/installs/ruby/3.3.7/openssl/lib/libcrypto.3.dylib

$ otool -L /opt/local/lib/mariadb-11.8/mysql/libmariadb.3.dylib | grep -i ssl
	/opt/local/libexec/openssl3/lib/libssl.3.dylib
	/opt/local/libexec/openssl3/lib/libcrypto.3.dylib
```

`mysql2.bundle` links against Ruby's own bundled OpenSSL; `libmariadb.3.dylib` links against MacPorts' separate OpenSSL. Both get loaded into the same process. `mysql2_tls_check_peer_identity` (`ext/mysql2/client.c`, added in #1570) calls `SSL_get_verify_result`/`SSL_get_app_data`/`SSL_get_peer_certificate` -- resolved against Ruby's OpenSSL -- on an `SSL*` pointer that was actually allocated and initialized by `libmariadb.3.dylib` using MacPorts' OpenSSL. `SSL`/`SSL_CTX` are opaque structs precisely because their internal layout isn't guaranteed stable across builds; handing one library's object to a different build's accessor functions is undefined behavior. Segfaults inside `SSL_get_app_data` -> `CRYPTO_get_ex_data`.

[...]

PR #1576, opened 2026-08-17
Map legacy :sslverify onto :ssl_mode instead of leaving it a MariaDB no-op

sodabrew:
## The problem

`:sslverify => true` sets `CLIENT_SSL_VERIFY_SERVER_CERT` directly as a `mysql_real_connect()` connect-flag. MySQL's client library reads that flag during the handshake -- it's literally what `ssl_mode: :verify_identity` sets internally on MySQL, so `:sslverify` works there. MariaDB Connector/C never reads that flag anywhere in its connect path (confirmed by grepping `mariadb_lib.c` end to end): `:sslverify => true` is a complete, silent no-op on MariaDB, with none of the enforcement #1570 built for `ssl_mode: :verify_identity`.

## The fix

`configure_tls_verification` now fills in `ssl_mode: :verify_identity` from `:sslverify => true` when the caller didn't give an explicit `:ssl_mode` -- so the legacy option gets the same enforcement (and the same fail-closed guarantees, including refusing to connect on builds that can't enforce it) `ssl_mode: :verify_identity` already has. An explicit `:ssl_mode` always wins over `:sslverify`.

Conversely, `:sslverify => false` alongside a verifying `ssl_mode` (`:verify_ca` or `:verify_identity`) is a direct contradiction -- one says the connection doesn't need verification, the other requires it -- so that combination now raises `Mysql2::Error::ConnectionError` before any connection attempt, rather than silently picking a side. This follows the option-coherence precedent #1555 established, though #1555's own checks are warn-only; this one raises since it's a hard contradiction of stated intent, not a silent-degradation case.

## Verification

Logic-level: six `:sslverify`/`:ssl_mode` combinations exercised against a `Klient` subclass (already in the spec file) that records connect args without dialing out -- covers the synthesize-on-true-and-unset case, explicit-mode-wins in both directions, and both raise/no-raise sides of the false+conflicting-mode case.

Live end-to-end against MySQL 8.0.34: `sslverify: true` now genuinely fails closed against a self-signed local server (`CA certificate is required if ssl-mode is VERIFY_CA or VERIFY_IDENTITY`), matching `ssl_mode: :verify_identity`'s own behavior exactly; `sslverify: false` + a verifying `ssl_mode` raises before any connection attempt.

**Could not do live end-to-end verification against MariaDB** -- blocked by a pre-existing, unrelated crash I found while testing this and filed separately as #1575 (dual-OpenSSL-linking segfault in `MYSQL2_VERIFY_IDENTITY_SHIM`). Confirmed that crash reproduces identically with plain `ssl_mode: :verify_identity` and zero `:sslverify` involved, so it isn't specific to this change -- it's a pre-existing hazard in already-merged #1570 code, just newly surfaced by exercising the callback path on a machine with two different OpenSSL builds loaded (Ruby's own bundled OpenSSL + MacPorts' separate OpenSSL, in this case). Logic-level MariaDB coverage (the `Klient`-based specs) doesn't touch the crashing code path at all, since it never actually connects.

Full suite (`bundle exec rspec`, `bundle exec rubocop`): 603 examples, 0 failures, 21 pending (all pre-existing environment skips). `.rubocop_todo.yml`'s `Metrics/ClassLength` bumped 164 -> 170 by hand for the new lines, matching the precedent in #1482/#1555.

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -39,7 +39,7 @@ Metrics/BlockNesting:
 # Offense count: 1
 # Configuration parameters: CountComments, CountAsOne.
 Metrics/ClassLength:
-  Max: 164
+  Max: 170
 
 # Offense count: 3
 # Configuration parameters: AllowedMethods, AllowedPatterns.
--- a/lib/mysql2/client.rb
+++ b/lib/mysql2/client.rb
@@ -50,7 +50,7 @@ module Mysql2
       self.charset_name = opts[:encoding] || 'utf8mb4'
 
       mode = parse_ssl_mode(opts[:ssl_mode]) if opts[:ssl_mode]
-      configure_tls_verification(opts, mode)
+      mode = configure_tls_verification(opts, mode)
 
       ssl_options = opts.values_at(:sslkey, :sslcert, :sslca, :sslcapath, :sslcipher)
       ssl_set(*ssl_options) if ssl_options.any? || opts.key?(:sslverify)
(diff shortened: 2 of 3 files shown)


sodabrew:
Rolling into #1572 instead -- #1572's own README changes describe `:sslverify` as a MariaDB no-op, and this PR fixes exactly that, so keeping them separate would mean documenting a limitation in the same breath as removing it. Closing in favor of a combined PR.

PR #1579, opened 2026-08-17
Match the verify_identity callback's OpenSSL to the client library's (#1575)

### LINKED ISSUE #1575: ssl_mode: :verify_identity segfaults on MariaDB when mysql2.bundle and libmariadb link different OpenSSL builds

jeremy:
Fixes #1575.

## Root cause

The #1570 `verify_identity` enforcement callback verifies the connector's TLS session by calling OpenSSL directly — `SSL_get_app_data`, `SSL_get_verify_result`, `SSL_get_peer_certificate`, `X509_check_host` — on the `SSL*` the connector created. Those calls resolve against whatever OpenSSL `mysql2.bundle` linked; the session belongs to whatever OpenSSL `libmariadb` linked. `SSL`/`X509` are opaque types whose layout is private to each build, so when those are two different builds the accessors misread the session — the deterministic segfault inside `CRYPTO_get_ex_data` in the issue. Nothing tied the two together: the extconf probes took whatever `-lssl -lcrypto` resolved to off the default search path.

I could reproduce this on a stock setup without MacPorts at all: a mise-built Ruby 4.0.6 keeps **static** `libssl.a`/`libcrypto.a` (OpenSSL 3.5.5) in Ruby's own libdir, which mkmf always searches first (`$DEFLIBPATH`) — ahead of every `-L` extconf can add, including `--with-openssl-dir`'s. A default build against Homebrew's mariadb-connector-c silently static-linked Ruby's OpenSSL into `mysql2.bundle` while the connector loaded Homebrew's 3.6.3. That pairing happened to work by layout luck; pairing with openssl@1.1 instead crashes with the issue's exact stack (`CRYPTO_get_ex_data` ← `mysql2_tls_check_peer_identity` ← `mysql2_tls_verification_callback`), no options beyond `ssl_mode: :verify_identity`.

## Fix

Both directions suggested in the issue, each covering what the other cannot see:

**1. Link time (extconf.rb, macOS).** Link the callback against the exact OpenSSL the client library links. `mariadb_config` doesn't say where that is (`--libs_sys` names `-lssl` with no `-L`), but the connector dylib's recorded install names do: read them with `otool -L` and pass those dylib **paths** straight to the linker. Full paths rather than `-L`/`-l` because no search-path ordering can beat mkmf's `$DEFLIBPATH` when Ruby's libdir holds OpenSSL static archives — which is also why `--with-openssl-dir` now selects its dylibs the same way. Headers come from the include directory installed next to the chosen dylibs, keeping compile and link on the same build. macOS-only by scope: two-level namespace binding is what lets two loaded OpenSSLs coexist with each image resolving separately, while ELF resolves shared symbols by SONAME process-wide (the first loaded `libssl.so.3` wins for every image), so Linux cannot bind mysql2 and libmariadb to two different loaded OpenSSLs this way.

**2. Runtime (client.c).** Before registering the callback, `:verify_identity` asks the dynamic loader — the only authority on what actually resolved — whether this extension and the client library's image agree on one address per OpenSSL library (`SSL_get_verify_result` for libssl, `X509_check_host` for libcrypto; `dladdr` to find the connector's image, `dlopen(RTLD_NOLOAD)` + `dlsym` to resolve through the connector's own dependency scope). Disagreement refuses the connection before anything touches the network, with an error naming both images and versions and the rebuild remedy — the same enforce-or-refuse posture as #1570. This catches what extconf cannot see: gems built before this change, a connector repackaged against a different OpenSSL after the gem was built, and the static-archive absorption above.

Loader identity was chosen over comparing version strings (`OpenSSL_version` vs the connector's `MARIADB_TLS_LIBRARY` report): two distinct same-version installs compare equal while remaining formally incompatible builds, and a version string can't name the images for the remedy. The strict form knowingly refuses same-version-different-install pairings that would have worked by layout luck; for a crash-or-verify security path that trade is deliberate, and the link-time layer makes common builds converge on one library anyway.

The check needs `dlopen`/`dlsym`/`dladdr`/`RTLD_NOLOAD` (glibc gates the latter two behind `_GNU_SOURCE`, added when needed); where that surface is missing (Windows) the probe compiles the check out and behavior is unchanged. `Mysql2::Client::TLS_OPENSSL_LINKAGE_CHECK` reports whether a build carries it. GnuTLS/Schannel connectors skip the check — the callback's own OpenSSL-backend refusal already covers them with an accurate message.

## What was verified, and where

On the repro machine (macOS arm64, mise Ruby 4.0.6 with vendored static OpenSSL 3.5.5, Homebrew mariadb-connector-c 3.4.7 against openssl@3 3.6.3):

* master + openssl@1.1 pairing: deterministic segfault, same stack as the issue.
* this branch, default build: links the connector's own `libssl.3`/`libcrypto.3` (otool-verified, no static OpenSSL absorbed), the runtime check passes, and `verify_identity` proceeds to real certificate verification.
* this branch, deliberately mismatched openssl@1.1 build: refuses at `Client.new` — `mysql2 calls SSL_get_verify_result in .../libssl.1.1.dylib (OpenSSL 1.1.1w ...) but the client library .../libmariadb.3.dylib resolves it in .../libssl.3.dylib (OpenSSL 3.6.3 ...)` — instead of crashing.
* `client_spec` compared against master on the same build config: no new failures (the pre-existing environment-specific Timeout-interrupt failures are identical on both).

What CI can and cannot verify: CI images carry a single OpenSSL install, so they **cannot produce the real mismatch** (that's why #1570's matrix never saw this). The refusal spec therefore forces the loader's verdict through a fail-closed-only env seam (`MYSQL2_TEST_FORCE_OPENSSL_LINKAGE_MISMATCH` — it can only refuse a connection, never permit one). What CI does prove: the real check runs, and must keep passing, inside every existing `verify_identity` leg (macOS mariadb@11.8 with the explicit `--with-openssl-dir` path, Ubuntu/Fedora MariaDB legs with the `have_library` fallback + `_GNU_SOURCE` probe), and the macOS legs exercise the new full-path linking.

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,9 +100,12 @@ If you have not done so already, you will need to install the XCode select tools
 `xcode-select --install`.
 
 Later versions of MacOS no longer distribute a linkable OpenSSL library. It is
-common to use Homebrew or MacPorts to install OpenSSL. Make sure that both the
-Ruby runtime and MySQL client libraries are compiled with the same OpenSSL
-family, 3.x, since only one can be loaded at runtime.
+common to use Homebrew or MacPorts to install OpenSSL. Where mysql2 itself
+needs OpenSSL (the `verify_identity` hostname verification callback on MariaDB
+builds), it automatically links the same OpenSSL the MySQL client library
+links, since the two must be the same build; `--with-openssl-dir` overrides
+that choice, and a mismatched pairing is refused at connect time rather than
+crashing (see issue #1575).
 
 ``` sh
 $ brew install openssl@3 zstd
@@ -387,6 +390,7 @@ Runtime client library and TLS feature introspection:
 |`Mysql2::Client::TLS_PEER_IDENTITY_VERIFICATION` | :native (MySQL), :callback (MariaDB Connector/C 3.4.3+), or nil (unenforceable)
 |`Mysql2::Client::TLS_VERSION_SUPPORTED`          | true/false
 |`Mysql2::Client::TLS_SNI_SUPPORTED`              | true/false
+|`Mysql2::Client::TLS_OPENSSL_LINKAGE_CHECK`      | true/false -- whether the build checks that mysql2 and the client library resolve to the same loaded OpenSSL before enforcing `:verify_identity`
 
 #### TLS verification modes
 
@@ -412,6 +416,13 @@ Notes:
   Either upgrade the client library, or downgrade the connection to `:verify_ca`
   and accept some risk of hostname uncertainty. Relying on the CA certificate
   validation only may be acceptable in your use case at your judgment.
+- The `:verify_identity` callback verifies the connector's TLS session with the
+  OpenSSL mysql2 linked, which is only sound when both resolve to the same
+  loaded library. If the process holds two different OpenSSL builds -- commonly
+  a Ruby with its own vendored OpenSSL next to a package manager's MariaDB
+  (#1575) -- `:verify_identity` refuses to connect with an error naming both
+  libraries instead of crashing. Rebuilding the gem fixes the pairing, since
+  mysql2 links the client library's own OpenSSL.
 
 #### Certificate fingerprint pinning
 
--- a/ext/mysql2/client.c
+++ b/ext/mysql2/client.c
@@ -201,6 +201,92 @@ typedef struct {
   void *ssl;
 } mysql2_mariadb_tls_head;
 
+/* The client library's TLS backend identity string ("OpenSSL 3.4.1 ..."),
+ * or NULL when the backend is not OpenSSL/LibreSSL -- under Schannel or
+ * GnuTLS, ctls->ssl is a different animal entirely and no OpenSSL-based
+ * verification can apply to it. */
+static const char *mysql2_tls_openssl_library(MYSQL *mysql) {
+  const char *tls_library = NULL;
+
+  if (mariadb_get_infov(mysql, MARIADB_TLS_LIBRARY, (void *)&tls_library) != 0 ||
+      tls_library == NULL ||
+      (strncmp(tls_library, "OpenSSL", 7) != 0 && strncmp(tls_library, "LibreSSL", 8) != 0))
+    return NULL;
+  return tls_library;
+}
+
+#ifdef MYSQL2_OPENSSL_LINKAGE_CHECK
+#include <dlfcn.h>
+#include <openssl/crypto.h>
+
+#if defined(OPENSSL_VERSION_NUMBER) && OPENSSL_VERSION_NUMBER >= 0x10100000L
+#define MYSQL2_OPENSSL_VERSION_STRING OpenSSL_version(OPENSSL_VERSION)
+#else
+#define MYSQL2_OPENSSL_VERSION_STRING SSLeay_version(SSLEAY_VERSION)
+#endif
+
+/* Compare where this extension and the client library each resolve the
+ * OpenSSL symbols the verification callback calls. SSL and X509 are opaque
+ * types whose layout is private to each OpenSSL build, so handing the
+ * client library's TLS session to another build's accessors is undefined
+ * behavior -- in practice a segfault inside libcrypto (#1575). Two builds
+ * commonly coexist in one process (a language runtime's vendored OpenSSL
+ * plus the package manager's build the client library links), and macOS
+ * two-level namespace binding resolves each image's references separately,
+ * so nothing else forces agreement. The dynamic loader is the authority on
+ * what actually resolved: for one representative symbol per OpenSSL
+ * library, the address this extension linked must be the address the
+ * client library's image resolves through its own dependencies. Returns
+ * NULL when they agree, or a description naming both images.
+ * connector_openssl is the client library's backend identity string, NULL
+ * when unknown. */
+static const char *mysql2_tls_openssl_linkage_mismatch(const char *connector_openssl, char *detail, size_t detail_len) {
+  static const char *const symbol_names[] = { "SSL_get_verify_result", "X509_check_host" };
+  void *const local_symbols[] = { (void *)SSL_get_verify_result, (void *)X509_check_host };
+  Dl_info connector_image, local_image, resolved_image;
+  void *handle;
+  size_t i;
+  const char *mismatch = NULL;
+
+  /* Test seam: exercising the refusal below needs two OpenSSL builds
+   * resolved into one process, which CI images (a single OpenSSL install)
+   * cannot produce, so the spec suite forces the loader's verdict instead.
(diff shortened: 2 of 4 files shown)


jeremy:
CI triage for this run:

* The three failing legs — ubuntu-22.04 ruby 2.7 mariadb10.11, ruby 3.0 mariadb10.11, ruby 3.0 mariadb10.6 — fail identically on master's last three runs (e.g. [run 32064696205](https://github.com/brianmario/mysql2/actions/runs/32064696205) on a8b6c4d): the #1572 spec `maps sslverify: true onto ssl_mode: :verify_identity` hits #1570's unenforceable-build refusal on Connector/C 3.3 legs, where `Klient.new(sslverify: true)` now raises at initialize. Pre-existing on master, untouched by this PR (whose code is entirely compiled out on Connector/C 3.3 — no `MARIADB_OPT_TLS_VERIFICATION_CALLBACK`). Happy to fix that in a separate PR.
* ubuntu-24.04 ruby 3.4 mariadb11.4 was cancelled 30 minutes into `apt update` (azure.archive.ubuntu.com mirror hang) before ever compiling — infra flake; its ruby 4.0 twin and both mariadb12.3 legs passed. I can't rerun jobs on this repo — a maintainer rerun of that one job should go green.

Everything this PR actually touches is green: all four macOS legs (mariadb@11.8 exercises the new install-name-derived OpenSSL linking, the dlfcn probe, and the runtime check against the real connector), the modern MariaDB Linux legs (linkage check compiled in via the `_GNU_SOURCE` probe, refusal spec runs), Fedora, and both Windows paths (check compiled out, constant false, spec skips).

jeremy:
That fix landed as #1578 (merged as 6bf79ff); rebased this PR onto it, so the fresh run should cover all 34 legs green.

PR #1580, opened 2026-08-18
Fix prepared-statement FLOAT casting to honor the field's declared precision

### LINKED ISSUE #1276: Prepared statement float value converted to 64bits ?
Hi,

When trying to switch to `prepared_statements`, our float columns had different values in ruby (2.6 series).

For instance, 1.1 is now 1.100000023841858

I know that 1.100000023841858 is a [limited floating point representation using 32bits](https://en.wikipedia.org/wiki/Single-precision_floating-point_format).

I'm far from being an expert in C but when looking at the code, [I've figured that a T_FLOAT is converted to as MYSQL_TYPE_DOUBLE in a case of a prepared statement](https://github.com/brianmario/mysql2/blob/33ee746f5af293a58c060a057a6bf94e4deb2051/ext/mysql2/statement.c#L319-L323)
```ruby
case T_FLOAT:
  bind_buffers[i].buffer_type = MYSQL_TYPE_DOUBLE;
  bind_buffers[i].buffer = xmalloc(sizeof(double));
  *(double*)(bind_buffers[i].buffer) = NUM2DBL(argv[i]);
  break;
 ```        

But it's not handle the same way within a simple [MySQL result](https://github.com/brianmario/mysql2/blob/33ee746f5af293a58c060a057a6bf94e4deb2051/ext/mysql2/result.c#L576-L581)
```ruby
case MYSQL_TYPE_FLOAT:        // float
  val = rb_float_new((double)(*((float*)result_buffer->buffer)));
  break;
case MYSQL_TYPE_DOUBLE:       // double
  val = rb_float_new((double)(*((double*)result_buffer->buffer)));
  break;
```

At least, the buffer seems to be handled differently.

Also, I did a small change locally by switching my `float` column to `double` column and 1.1 stays 1.1.

Thanks for your time :)

sodabrew:
## Summary

Fixes #1276. MySQL's binary protocol returns `FLOAT` columns as raw float32 bits; naively widening that to a Ruby `Float` (a `double`) surfaces digits float32 never held (`1.1` -> `1.100000023841858`). This reformats the value to match what the field's own declared precision would show via the text protocol: `%.6g` (6 significant digits) when no precision is declared (`decimals == 31`, MySQL's `NOT_FIXED_DEC` sentinel), or `%.*f` with the field's exact decimal count for `FLOAT(M,D)`.

The formatted string is parsed with `rb_cstr_to_dbl()` -- the same locale-independent C function `Kernel#Float()` calls internally, called directly to skip a String allocation and a method dispatch. The `snprintf()` doing the formatting is Ruby's own locale-independent `ruby_snprintf` (`ruby/subst.h` `#define`s `snprintf` to it inside any Ruby C extension), not libc's locale-sensitive `snprintf`.

During review, @jeremy caught a real buffer-overflow bug in this same fix: `float_buf`'s original 32 bytes silently truncated wide `FLOAT(M,D)` values (`D` up to 30, combined with float32's max magnitude of 39 integer digits, needs up to 72 bytes) -- producing a *different, wrong* number rather than a crash. Fixed by sizing the buffer for the true worst case, and by checking `snprintf()`'s return value at every FLOAT/DOUBLE call site: if a formatted value ever exceeds its buffer, mysql2 now raises `Mysql2::Error` instead of silently parsing a truncated string as the wrong value.

The same silent-truncation risk existed in the **text protocol** paths too (plain `client.query`, and the `cast: false`/`cast: :fast` prepared-statement string-bind path). Those used `Kernel#Float()` on an unbounded Ruby String, so they were already truncation-safe, but paid for a String allocation and a method dispatch per FLOAT/DOUBLE cell. Converted both to the same `snprintf` + `rb_cstr_to_dbl()` pattern -- buffer sized at 512 bytes, comfortably above `DOUBLE(255,30)`'s actual worst case of 257 bytes (confirmed empirically against a live server) -- microbenchmarked at roughly 2x faster than `Kernel#Float()` on typical short values.

## Test plan

- [x] FLOAT/DOUBLE parity spec (prepared statement vs. plain query) widened to `FLOAT(255,30)` near `FLT_MAX`, proven to catch the buffer-overflow bug class by temporarily shrinking the buffer and confirming the spec fails, then restoring the fix.
- [x] Two new specs (`does not truncate a wide DOUBLE(M,D) value`) covering both text-protocol paths, verified the same way.
- [x] Locale-independence specs (comma-decimal `de_DE.UTF-8`) across all three FLOAT/DOUBLE code paths.
- [x] `bundle exec rspec` -- 633 examples, 0 failures.
- [x] RuboCop clean.

### DIFF
--- a/ext/mysql2/result.c
+++ b/ext/mysql2/result.c
@@ -98,7 +98,7 @@ static VALUE cMysql2Result, cDateTime, cDate;
 static VALUE opt_decimal_zero, opt_float_zero, opt_time_year, opt_time_month, opt_time_day, opt_utc_offset;
 static VALUE opt_time_anchor_utc;
 static ID intern_new, intern_utc, intern_local, intern_localtime, intern_local_offset,
-  intern_civil, intern_new_offset, intern_merge, intern_BigDecimal, intern_Float,
+  intern_civil, intern_new_offset, intern_merge, intern_BigDecimal,
   intern_query_options, intern_plus;
 static VALUE sym_symbolize_keys, sym_as, sym_array, sym_database_timezone,
   sym_application_timezone, sym_local, sym_utc, sym_cast_booleans,
@@ -2687,7 +2687,6 @@ void init_mysql2_result(void) {
   intern_civil        = rb_intern("civil");
   intern_new_offset   = rb_intern("new_offset");
   intern_BigDecimal   = rb_intern("BigDecimal");
-  intern_Float        = rb_intern("Float");
   intern_query_options = rb_intern("@query_options");
   intern_plus         = rb_intern("+");
 
--- a/ext/mysql2/result.c
+++ b/ext/mysql2/result.c
@@ -1494,21 +1494,23 @@ static VALUE rb_mysql_result_fetch_row_stmt(VALUE self, MYSQL_FIELD * fields, co
         case MYSQL_TYPE_FLOAT: {      // float
           /* Honor the field's declared precision instead of upconverting
            * raw float32 to a double directly, which surfaces digits
-           * float32 can't distinguish. MySQL formats FLOAT with 6
-           * significant digits by default, or exactly D decimal places
-           * for FLOAT(M,D); decimals == 31 is MySQL's NOT_FIXED_DEC
-           * sentinel for "no explicit precision" (not in the public
-           * headers). Parsed via Kernel#Float(), not strtod(): only
-           * strtod() honors LC_NUMERIC's decimal separator, so pairing
-           * it with snprintf() here would misparse under a non-'.'
-           * locale. */
+           * float32 can't distinguish: 6 significant digits by default,
+           * or exactly D decimal places for FLOAT(M,D); decimals == 31
+           * is MySQL's NOT_FIXED_DEC sentinel for "no explicit
+           * precision" (not in the public headers). Format to a
+           * string, then parse it with rb_cstr_to_dbl() -- Kernel#
+           * Float()'s own locale-independent parser, called directly
+           * to skip a String allocation and a method dispatch.
+           * strtod() would read LC_NUMERIC's decimal separator
+           * instead, misparsing this always-'.' string under a
+           * non-'.' locale. */
           char float_buf[32];
           double float_as_double = (double)(*((float*)result_buffer->buffer));
           if (fields[i].decimals == 31)
             snprintf(float_buf, sizeof(float_buf), "%.6g", float_as_double);
           else
             snprintf(float_buf, sizeof(float_buf), "%.*f", fields[i].decimals, float_as_double);
-          val = rb_funcall(rb_mKernel, intern_Float, 1, rb_str_new2(float_buf));
+          val = rb_float_new(rb_cstr_to_dbl(float_buf, 0));
           break;
         }
         case MYSQL_TYPE_DOUBLE:       // double
--- a/ext/mysql2/result.c
+++ b/ext/mysql2/result.c
@@ -1492,27 +1492,20 @@ static VALUE rb_mysql_result_fetch_row_stmt(VALUE self, MYSQL_FIELD * fields, co
           }
           break;
         case MYSQL_TYPE_FLOAT: {      // float
-          /* A direct float->double cast preserves the exact float32 value
-           * but not what the text protocol already shows for the same
-           * stored value (#1276), confirmed against a live server: a
-           * FLOAT column with no explicit precision serializes with 6
-           * significant digits (1.1 -> "1.1", but 1.0/3 -> "0.333333"
-           * and 123456.789 -> "123457", both losing digits float32
-           * actually stores), while FLOAT(M,D) serializes with exactly D
-           * decimal places. fields[i].decimals reports which: 31 is
-           * MySQL's NOT_FIXED_DEC sentinel for "no explicit precision"
-           * (not exposed in the public client headers, hence the literal
-           * here) or the declared D otherwise. Match either convention
-           * instead of upconverting the raw bits directly, which
-           * surfaces every digit float32 stores (1.100000023841858,
-           * 0.3333333432674408) that a plain SELECT of the same column
-           * would never show. snprintf() here does not reliably honor
-           * LC_NUMERIC the way strtod() does (confirmed under a comma
-           * locale: snprintf() still emits '.', so pairing it with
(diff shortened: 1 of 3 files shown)


ext/mysql2/result.c
```diff
@@ -1459,9 +1459,26 @@ static VALUE rb_mysql_result_fetch_row_stmt(VALUE self, MYSQL_FIELD * fields, co
...
+           * float32 can't distinguish. MySQL formats FLOAT with 6
+           * significant digits by default, or exactly D decimal places
+           * for FLOAT(M,D); decimals == 31 is MySQL's NOT_FIXED_DEC
+           * sentinel for "no explicit precision" (not in the public
+           * headers). Parsed via Kernel#Float(), not strtod(): only
+           * strtod() honors LC_NUMERIC's decimal separator, so pairing
+           * it with snprintf() here would misparse under a non-'.'
+           * locale. */
+          char float_buf[32];
+          double float_as_double = (double)(*((float*)result_buffer->buffer));
```
jeremy, line 1524: `%.*f` overflows this buffer for wide `FLOAT(M,D)` columns: a float32 carries up to 39 integer digits and D goes up to 30, so the fixed-notation string reaches 72 bytes. `snprintf` truncates silently and `Float()` then parses the prefix as a different number. Reproduces on MySQL 9.6:

```sql
CREATE TABLE t (a FLOAT(60,20)); INSERT INTO t VALUES (1e35);
```
```
query:   1.0000000409184788e+35
prepare: 1.0000000409184787e+30
```

The text path sizes its buffer at 512 for the same my_fcvt worst case (result.c:1061-1063); 80 covers float32 here. The parity spec only exercises `FLOAT(10,2)`, so a wide column with a large value there would have caught this.

ext/mysql2/result.c
```diff
@@ -1459,9 +1459,26 @@ static VALUE rb_mysql_result_fetch_row_stmt(VALUE self, MYSQL_FIELD * fields, co
...
+           * raw float32 to a double directly, which surfaces digits
+           * float32 can't distinguish. MySQL formats FLOAT with 6
+           * significant digits by default, or exactly D decimal places
+           * for FLOAT(M,D); decimals == 31 is MySQL's NOT_FIXED_DEC
+           * sentinel for "no explicit precision" (not in the public
+           * headers). Parsed via Kernel#Float(), not strtod(): only
+           * strtod() honors LC_NUMERIC's decimal separator, so pairing
+           * it with snprintf() here would misparse under a non-'.'
+           * locale. */
+          char float_buf[32];
```
jeremy: The locale rationale is inverted: C's `snprintf()` honors `LC_NUMERIC` exactly as `strtod()` does. What makes this pairing safe is that `ruby.h` (via `ruby/subst.h`) redefines `snprintf` to `ruby_snprintf`, which always emits `.`. `Kernel#Float()` is still a correct parse, though `rb_cstr_to_dbl(float_buf, 0)` does the same locale-independent parse without a String allocation and method dispatch per cell.

Also, `NOT_FIXED_DEC` is in 5.7's public `mysql_com.h`; 8.0+ is what dropped it.

jeremy (changes requested):
Approach is right and the parity holds everywhere I swept it (zero, subnormal, 1e-10, 1e30, 3.4e38, values needing >6 digits, FLOAT(10,2)/(7,0)/(30,25)), CI is green on MySQL and MariaDB, and the specs pass locally. One real bug in the fixed-notation branch, plus a comment correction.

ext/mysql2/result.c
jeremy: Still open on 2afb858: `float_buf` is 32 bytes and the parity spec still only covers `FLOAT(10,2)`. Same repro against this head on MySQL 9.6.0:

```sql
CREATE TABLE t (a FLOAT(60,20)); INSERT INTO t VALUES (1e35);
```
```
query:   1.0000000409184788e+35
prepare: 1.0000000409184787e+30
```

`%.*f` with D=20 on 1e35 is a 57-byte string; `snprintf` truncates to 31 chars and `rb_cstr_to_dbl` parses the prefix as a different number. 80 bytes covers the float32 worst case (39 integer digits + '.' + 30 decimals + sign + NUL); a `FLOAT(60,20)` column with a value like `1e35` in the parity spec pins it.

ext/mysql2/result.c
jeremy: Addressed in 2afb858. One leftover: the comment still says `NOT_FIXED_DEC` is "not in the public headers"; it's in 5.7's `mysql_com.h`, 8.0+ dropped it.

jeremy (changes requested):
2afb858 switches to `rb_cstr_to_dbl` and straightens out the locale comment, but `float_buf[32]` still truncates wide `FLOAT(M,D)` values: on this head, `FLOAT(60,20)` holding `1e35` reads `1.0000000409184788e+35` via query and `1.0000000409184787e+30` via prepare. Sizing the buffer to 80 and adding a wide column to the parity spec closes it.

ext/mysql2/result.c
jeremy: Addressed in 5dab863. Re-ran the repro on bade4c4: `FLOAT(60,20)` with `1e35` now returns `1.0000000409184788e+35` from both the query and prepared paths, byte-identical. Worst case is 39 integer digits + 30 decimals + sign + point + NUL = 72, so 80 covers it, and the parity spec now exercises FLOAT(255,30) near FLT_MAX. statement_spec + result_spec pass locally (411 examples, 0 failures).

ext/mysql2/result.c
jeremy: Comment fixed in 5dab863.

jeremy (approved):
Text/binary FLOAT parity now holds across the sweep, including wide FLOAT(M,D) near FLT_MAX — repro and both spec files pass locally against bade4c4.

jeremy (approved):
Re-approving at 4567d67: the three FLOAT/DOUBLE parse sites now share mysql2_float_from_str / mysql2_snprintf_to_dbl; text and prepared reads agree across a zero/subnormal/1e-10..1e38/FLOAT(M,D)/DOUBLE(255,30)/DBL_MAX sweep, and the binary FLOAT arm still formats to the field's declared precision before parsing.

jeremy (approved):
Still good at 9629fb1 — the if/else is a straight rewrite of the ternary, same two snprintf calls.

PR #1581, opened 2026-08-18
Add :downcase_keys result option

### LINKED ISSUE #1241: Request for :downcase_keys option
Hey there and thanks for a great library!

I'm working with a legacy database where there's a mixture of uppercase and lowercase column names, but I need all the result keys symbolized AND in lower case...

It's easy to convert them in the application code, but seems like a useful feature for the gem...

Thanks for considering! And if there's interest in this feature, I'll try to scrape together some free time to implement it.

sodabrew:
## Summary

Fixes #1241. Adds a `:downcase_keys` option, downcasing result field names (String or Symbol, composing with `:symbolize_keys`) -- for connecting to a legacy database with inconsistently-cased column names without transforming every row in application code.

```ruby
client.query("SELECT * FROM legacy_table", downcase_keys: true).first
# => {"foo_bar" => 1}  -- regardless of whether the column is named foo_bar, FOO_BAR, or FooBar

client.query("SELECT * FROM legacy_table", downcase_keys: true, symbolize_keys: true).first
# => {foo_bar: 1}
```

Implementation mirrors `:symbolize_keys` at every layer it touches:
- The per-`#each` options cache (`mysql2_each_opts_cache`).
- The prepared-statement metadata cache -- a statement's cached field-name array is only adopted when both `:symbolize_keys` *and* `:downcase_keys` match the execute adopting it.
- `Result#fields` reading the option straight from `@query_options`.

Downcasing operates on the ASCII `a-z`/`A-Z` range only, byte by byte: safe on a UTF-8 column name, since every continuation byte in a multibyte sequence is `>= 0x80`, outside that range, so it can never corrupt one -- it just doesn't case-fold non-ASCII letters, which unambiguously covers real-world column/alias names.

## Test plan

- [x] Basic behavior: `:downcase_keys` alone, composed with `:symbolize_keys`, unset (no-op), and a non-ASCII column name (verifies only ASCII bytes are touched).
- [x] `Result#fields` honors per-`#each` `:downcase_keys`, including after a force-freed abandoned stream.
- [x] Prepared-statement metadata cache invalidation: alternating `:downcase_keys` across three consecutive executes of the same statement produces the correct field names each time, not a stale cached copy.
- [x] `bundle exec rspec spec/` -- 633 examples, 0 failures.
- [x] RuboCop clean.

### DIFF
--- a/ext/mysql2/result.c
+++ b/ext/mysql2/result.c
@@ -1,6 +1,5 @@
 #include <mysql2_ext.h>
 
-#include <ctype.h>
 #include <limits.h>
 #include <string.h>
 
@@ -444,16 +443,28 @@ static VALUE rb_mysql_result_fetch_field(VALUE self, unsigned int idx, int symbo
     name_length = name_end ? (size_t)(name_end - field->name) : field->name_length;
     field_name = field->name;
 
-    /* :downcase_keys lowercases the ASCII a-z/A-Z range only, byte by byte.
-     * Safe on a UTF-8 name: every continuation byte in a multibyte sequence
-     * is >= 0x80, outside that range, so this can never corrupt one -- it
-     * just doesn't case-fold non-ASCII letters, which is fine since column
-     * and alias names are effectively always ASCII in practice. */
+    /* :downcase_keys lowercases the ASCII A-Z range only, byte by byte, by
+     * hand rather than via tolower(3): tolower() is locale-dependent, and
+     * MySQL's wire format is not (the same class of bug this project has
+     * hit before with strtod() -- see result.c's DECIMAL/FLOAT parsing).
+     * Under an ISO8859-1-family locale, tolower(0xC3) returns 0xE3,
+     * corrupting the first byte of 'À' (U+00C0) instead of leaving it
+     * alone -- confirmed directly, not just suspected: a locale-independent
+     * A-Z check on a UTF-8 name is safe, because every continuation byte in
+     * a multibyte sequence is >= 0x80, outside that range, so this can
+     * never corrupt one -- it just doesn't case-fold non-ASCII letters.
+     * MySQL identifiers are not restricted to ASCII (a UTF-8 alias like
+     * `fooÀbar` is legal and exercised by this file's specs), so that's a
+     * deliberate scope limit, not a "rare in practice" shortcut:
+     * :downcase_keys folds the ASCII range only, on purpose, rather than
+     * taking on full Unicode case-folding. */
     if (downcase_keys && name_length > 0) {
       size_t i;
       char *buf = ALLOCV_N(char, downcase_holder, name_length);
-      for (i = 0; i < name_length; i++)
-        buf[i] = (char)tolower((unsigned char)field_name[i]);
+      for (i = 0; i < name_length; i++) {
+        unsigned char c = (unsigned char)field_name[i];
+        buf[i] = (c >= 'A' && c <= 'Z') ? (char)(c + ('a' - 'A')) : (char)c;
+      }
       field_name = buf;
     }
 
--- a/ext/mysql2/result.c
+++ b/ext/mysql2/result.c
@@ -443,21 +443,12 @@ static VALUE rb_mysql_result_fetch_field(VALUE self, unsigned int idx, int symbo
     name_length = name_end ? (size_t)(name_end - field->name) : field->name_length;
     field_name = field->name;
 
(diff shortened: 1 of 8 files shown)


ext/mysql2/result.c
```diff
@@ -435,33 +441,54 @@ static VALUE rb_mysql_result_fetch_field(VALUE self, unsigned int idx, int symbo
...
+     * functions. MySQL identifiers may be any Basic Multilingual Plane
+     * Unicode character (up to 3-byte UTF-8; 4-byte supplementary
+     * characters are not permitted), but for this feature, only A-Z is
+     * downcased to a-z. */
+    if (downcase_keys && name_length > 0) {
+      size_t i;
+      char *buf = ALLOCV_N(char, downcase_holder, name_length);
+      for (i = 0; i < name_length; i++) {
+        unsigned char c = (unsigned char)field_name[i];
+        buf[i] = (c >= 'A' && c <= 'Z') ? (char)(c + ('a' - 'A')) : (char)c;
```
jeremy, line 457: This byte-wise loop is safe for UTF-8 -- trail bytes are all >= 0x80 -- but not for the SJIS-class connection encodings mysql2 supports: cp932/Shift_JIS, GBK, and Big5 all use trail bytes in 0x40-0x7E, which includes A-Z. Those bytes get case-flipped here, corrupting the character. Repro on this branch against a cp932 connection:

```ruby
sj = Mysql2::Client.new(encoding: "cp932", ...)
q = (+"SELECT 1 AS `\x83\x41B`").force_encoding("Windows-31J")  # alias \u30A2B
sj.query(q).first.keys.first
#  => "\u30A2B" (83 41 42)
sj.query(q, downcase_keys: true).first.keys.first
#  => "\u30C2b" (83 61 62) -- \u30A2 corrupted to \u30C2
```

The key is tagged conn_enc, so this isn't a mis-tag; the bytes themselves are wrong. The general fix is to walk characters instead of bytes -- step with `rb_enc_mbclen(p, end, conn_enc)` (conn_enc is already in scope) and downcase only when the char length is 1. Pure-ASCII and UTF-8 names cost the same as today. The comment above and the PR body's safety argument want the same adjustment: continuation bytes >= 0x80 is a UTF-8/EUC property, not a general one.

jeremy (changes requested):
The mechanism is right where it needs to be: downcasing at field-name materialization, once per field, on both the query and prepared paths, with the statement metadata cache keyed on the new flag so cross-execute reuse can't serve stale casing. I ran result_spec + statement_spec locally on fd3d63d (423 examples, 0 failures) and probed both paths with mixed-case and non-ASCII names; symbolize composition and the alternating-execute cache invalidation behave.

One real bug, inline: the byte-wise A-Z loop corrupts multibyte names on SJIS-class connection encodings (cp932/Shift_JIS, GBK, Big5), whose multibyte trail bytes land in the A-Z range. The UTF-8 safety argument doesn't extend to those charsets, and the string keys are tagged with the connection encoding, so the corrupted bytes are user-visible.

Note: no CI ran on this push -- the [skip ci] tip suppressed the whole chain, so there's no rollup on either head commit.

ext/mysql2/result.c
sodabrew: Great finding! I'm not going to land this PR. It was a good exercise to go through the implementation nonetheless.

sodabrew:
I'm satisfied with the state of this change, but am going to close it out as rejected. The PR remains if there's any new call for this feature, or if other database drivers have something similar and we want to copy it for consistency.

Here's the proposed squash-merge text, written before the finding that SJIS encodings do get corrupted by the naive algorithm.

```
Downcases result field names for connecting to legacy databases with
inconsistently-cased column names without needing to transform every
row in application code.

Mirrors :symbolize_keys at every layer it touches: the per-#each
options cache, the statement metadata cache (a prepared statement's
cached field-name array is only adopted when both :symbolize_keys and
:downcase_keys match the execute that's adopting it) and Result#fields
reading the option from @query_options directly.

MySQL identifiers may be any Basic Multilingual Plane Unicode character,
up to 3-byte UTF-8, but for this feature, only A-Z is downcased to a-z.
Case folding in Unicode is a tricky subject because different languages
fold differently and a substantial table of knowledge is needed for global
interoperability. Therefore an intentionally naive algorithm is used: add
0x20 to any byte in the range 0x41 - 0x5A; since Unicode continutation
bytes are always greater than 0x80, this does not break a multi-byte
character mid-sequence.

Fixes #1241

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>
```

PR #1583, opened 2026-08-18
Drop :database, :default_file, :default_group from query_options after connecting (#437, #493)

sodabrew:
## Summary

`:database` is a connect-time-only option: libmysqlclient reads it once at connect and never consults it again. But mysql2 kept a copy sitting in `query_options`, where every other key (`:as`, `:symbolize_keys`, `:downcase_keys`, `:cast`, ...) genuinely is live and re-consulted per query. That made `client.query_options[:database] = "prod"` look exactly like a working option change, when it silently did nothing (#437).

`#database` (added in #1245) already reads the live database directly off the connection and stays correct through `#select_db` and session-tracked `USE` statements -- it just wasn't cross-referenced from anywhere `:database` in `query_options` was discussed. This PR:

- Drops `:database` (and its deprecated `:dbname`/`:db` aliases) from `query_options` after connecting, the same way `:password`/`:pass` are already dropped -- no more decoy key.
- Cross-references `#select_db` and `#database` in the rdoc and README so the supported way to read/change the current database is discoverable.
- Adds specs covering: the key's absence right after connect (with and without an initial `:database`), and its continued absence after `#select_db`.

**Also drops `:default_file`/`:default_group` (#493), same underlying confusion:** these are read once from the `opts` hash passed to `Client.new` (never from `@query_options`), so `Client.default_query_options.merge!(default_file: ...)` looks like it configures future connections and silently doesn't. Unlike `:database` there's no live value to read back afterward -- once connected they're nothing but a decoy, so they're dropped the same way, with a matching spec.

Supersedes #507, which predates the current `check_and_clean_query_options`/`#database` machinery and no longer applies cleanly.

This intentionally does *not* do the fuller `query_options` vs. `connect_options` split brianmario floated in the original #437 thread -- it fixes the specific keys these two reports were actually about, without touching the keys (`:as`, `:cast`, `:symbolize_keys`, `:downcase_keys`) that are genuinely reactive and already work correctly today.

## Test plan
- [x] `bundle exec rspec spec/mysql2/client_spec.rb` -- 214 examples, 0 failures
- [x] `bundle exec rspec` (full suite) -- 630 examples, 0 failures
- [x] `bundle exec rubocop lib/mysql2/client.rb spec/mysql2/client_spec.rb` -- no offenses

### DIFF
--- a/lib/mysql2/client.rb
+++ b/lib/mysql2/client.rb
@@ -317,6 +317,10 @@ module Mysql2
       @query_options.delete(:database)
       @query_options.delete(:dbname)
       @query_options.delete(:db)
+
+      # :default_file/:default_group: same story as :database above, but with
+      # no live value to read back -- nothing but a decoy (#493).
+      %i[default_file default_group].each { |k| @query_options.delete(k) }
     end
 
     class << self
--- a/lib/mysql2/client.rb
+++ b/lib/mysql2/client.rb
@@ -1,5 +1,5 @@
 module Mysql2
-  class Client
+  class Client # rubocop:disable Metrics/ClassLength
     attr_reader :query_options, :read_timeout
 
     def self.default_query_options
--- a/lib/mysql2/client.rb
+++ b/lib/mysql2/client.rb
@@ -307,6 +307,16 @@ module Mysql2
       # avoid logging sensitive data via #inspect
       @query_options.delete(:password)
       @query_options.delete(:pass)
+
+      # :database is a connect-time-only setting: libmysqlclient never
+      # re-reads it, so leaving it in @query_options would let callers set
+      # query_options[:database] = "x" and see it silently do nothing
+      # (#437). #database always reflects the live, current database
+      # (updated by #select_db too), so it's the correct place to look --
+      # drop the inert copy here rather than let it linger as a decoy.
+      @query_options.delete(:database)
+      @query_options.delete(:dbname)
+      @query_options.delete(:db)
     end
 
     class << self
--- a/lib/mysql2/client.rb
+++ b/lib/mysql2/client.rb
@@ -54,15 +54,49 @@ module Mysql2
     }.freeze
     private_constant :TLS_OPTION_ALIASES
 
+    # Keys Client.new(...) recognizes as per-query-reactive -- these end up
+    # in @query_options, inherited by every #query call. Anything not
+    # listed here or in SUPPORTED_CONNECT_OPTIONS is silently ignored
+    # rather than raised on (#65's "TODO: stricter validation"), matching
+    # existing behavior for an unrecognized option.
+    SUPPORTED_QUERY_OPTIONS = %i[
+      as async cast cast_booleans symbolize_keys database_timezone
+      application_timezone cache_rows rows_per_gvl_yield stream
+      force_encoding
+    ].freeze
+    private_constant :SUPPORTED_QUERY_OPTIONS
+
+    # Keys Client.new(...) recognizes as connect-time-only -- read once
+    # here, in @connect_options, and never re-consulted afterward. A key
+    # in this list must never also be added to SUPPORTED_QUERY_OPTIONS:
+    # that's exactly the shape of bug #437/#493 fixed.
+    SUPPORTED_CONNECT_OPTIONS = %i[
+      host hostname username user password pass port database dbname db
+      socket sock tls_sni_name reconnect connect_timeout local_infile
+      read_timeout write_timeout default_file default_group secure_auth
+      init_command automatic_close enable_cleartext_plugin default_auth
+      get_server_public_key tls_version encoding ssl_mode sslkey sslcert
+      sslca sslcapath sslcipher sslverify tls_key tls_cert tls_ca
+      tls_capath tls_cipher tls_mode tls_passphrase tls_peer_fingerprint
+      tls_peer_fingerprint_list connect_attrs flags connect_flags
+    ].freeze
+    private_constant :SUPPORTED_CONNECT_OPTIONS
+
     def initialize(opts = {})
       raise Mysql2::Error, "Options parameter must be a Hash" unless opts.is_a? Hash
 
       opts = Mysql2::Util.key_hash_as_symbols(opts)
       @read_timeout = nil
+
+      # Must run before apply_tls_option_aliases: that method overwrites a
+      # legacy :ssl* key's value with its :tls_* equivalent in place, so
+      # @connect_options has to capture both spellings' original values
+      # first, or warn_incoherent_options's alias-mismatch check below
+      # would never see them differ.
       @query_options = self.class.default_query_options.dup
-      @query_options.merge! opts
+      @query_options.merge!(opts.select { |k, _| SUPPORTED_QUERY_OPTIONS.include?(k) })
       @connect_options = self.class.default_connect_options.dup
-      @connect_options.merge!(opts.slice(*@connect_options.keys))
+      @connect_options.merge!(opts.select { |k, _| SUPPORTED_CONNECT_OPTIONS.include?(k) })
 
       apply_tls_option_aliases(opts)
 
@@ -74,6 +108,12 @@ module Mysql2
       # Set an explicit default false for LOCAL INFILE support: some client libraries
       # enable it by default, letting a malicious server read files off the client.
       opts[:local_infile] = false unless opts.key?(:local_infile)
+
+      # Both were resolved after @connect_options was already sliced from
+      # opts above; propagate the resolved values so connect_options
+      # reflects what was actually used, not just what the caller passed.
+      @connect_options[:connect_timeout] = opts[:connect_timeout]
+      @connect_options[:local_infile] = opts[:local_infile]
 
       # TODO: stricter validation rather than silent massaging
       %i[reconnect connect_timeout local_infile read_timeout write_timeout default_file default_group secure_auth init_command automatic_close enable_cleartext_plugin default_auth get_server_public_key tls_version].each do |key|
@@ -297,8 +337,8 @@ module Mysql2
       # that never names the real problem. Either option may be given under
       # its legacy :ssl* name or its :tls_* alias.
       # https://dev.mysql.com/doc/refman/en/using-encrypted-connections.html
-      effective_key = @query_options[:tls_key] || @query_options[:sslkey]
-      effective_cert = @query_options[:tls_cert] || @query_options[:sslcert]
+      effective_key = @connect_options[:tls_key] || @connect_options[:sslkey]
+      effective_cert = @connect_options[:tls_cert] || @connect_options[:sslcert]
       warn ":sslkey and :sslcert only take effect together; alone, libmysqlclient sends no client certificate and MariaDB Connector/C fails to connect" \
         if effective_key.nil? != effective_cert.nil?
 
@@ -306,8 +346,8 @@ module Mysql2
       # different values, the :tls_* value silently wins (see
       # TLS_OPTION_ALIASES); warn so the conflict isn't invisible.
       TLS_OPTION_ALIASES.each do |tls_key, legacy_key|
-        next unless @query_options.key?(tls_key) && @query_options.key?(legacy_key)
-        next if @query_options[tls_key] == @query_options[legacy_key]
+        next unless @connect_options.key?(tls_key) && @connect_options.key?(legacy_key)
+        next if @connect_options[tls_key] == @connect_options[legacy_key]
 
         warn ":#{legacy_key} and :#{tls_key} were both given with different values; :#{tls_key} wins"
       end
@@ -322,34 +362,19 @@ module Mysql2
     def check_and_clean_query_options
       warn_incoherent_options
 
-      if %i[user pass hostname dbname db sock].any? { |k| @query_options.key?(k) }
+      if %i[user pass hostname dbname db sock].any? { |k| @connect_options.key?(k) }
         warn "============= WARNING FROM mysql2 ============="
         warn "The options :user, :pass, :hostname, :dbname, :db, and :sock are deprecated and will be removed at some point in the future."
         warn "Instead, please use :username, :password, :host, :port, :database, :socket, :flags for the options."
         warn "============= END WARNING FROM mysql2 ========="
       end
 
-      # avoid logging sensitive data via #inspect
-      @query_options.delete(:password)
-      @query_options.delete(:pass)
-
-      # :database is a connect-time-only setting: libmysqlclient never
-      # re-reads it, so leaving it in @query_options would let callers set
-      # query_options[:database] = "x" and see it silently do nothing
-      # (#437). #database always reflects the live, current database
-      # (updated by #select_db too), so it's the correct place to look --
-      # drop the inert copy here rather than let it linger as a decoy.
-      @query_options.delete(:database)
-      @query_options.delete(:dbname)
-      @query_options.delete(:db)
-
-      # :default_file/:default_group: same story as :database above, but with
-      # no live value to read back -- nothing but a decoy (#493).
-      %i[default_file default_group].each { |k| @query_options.delete(k) }
-
-      # :connect_flags now lives in @connect_options; drop the blanket-merge
-      # copy here so it doesn't linger as a second, stale source of truth.
-      @query_options.delete(:connect_flags)
+      # avoid logging sensitive data via #inspect. :password/:pass never
+      # entered @query_options in the first place (not in
+      # SUPPORTED_QUERY_OPTIONS), but @connect_options is where they
+      # genuinely live now, so the redaction moves here with them.
+      @connect_options.delete(:password)
+      @connect_options.delete(:pass)
     end
 
     class << self
--- a/lib/mysql2/client.rb
+++ b/lib/mysql2/client.rb
@@ -1,7 +1,13 @@
 module Mysql2
   class Client # rubocop:disable Metrics/ClassLength
-    attr_reader :query_options, :read_timeout
+    attr_reader :query_options, :connect_options, :read_timeout
 
+    # Options genuinely re-consulted on every #query call (directly, or via
+    # a per-instance default set here). Connect-time-only settings -- read
+    # once at Client.new and never again -- belong in
+    # .default_connect_options instead, not here: a key that only takes
+    # effect at connect time but lives in query_options looks like a live,
+    # per-query option and silently isn't one (#437, #493).
     def self.default_query_options
       @default_query_options ||= {
         as: :hash,                   # the type of object you want each row back as; also supports :array (an array of values)
@@ -12,10 +18,16 @@ module Mysql2
         application_timezone: nil,   # timezone Mysql2 will convert to before handing the object back to the caller
         cache_rows: true,            # tells Mysql2 to use its internal row cache for results
         rows_per_gvl_yield: 8192,    # buffered rows to materialize between GVL yields; 0 disables yielding
-        connect_flags: REMEMBER_OPTIONS | LONG_PASSWORD | LONG_FLAG | TRANSACTIONS | PROTOCOL_41 | SECURE_CONNECTION | CONNECT_ATTRS,
         cast: true,
-        default_file: nil,
-        default_group: nil,
+      }
+    end
+
+    # Options read once at connect time and never re-consulted afterward.
+    # Unlike .default_query_options, these have no live per-query meaning --
+    # changing them after connecting has no effect, by design.
+    def self.default_connect_options
+      @default_connect_options ||= {
+        connect_flags: REMEMBER_OPTIONS | LONG_PASSWORD | LONG_FLAG | TRANSACTIONS | PROTOCOL_41 | SECURE_CONNECTION | CONNECT_ATTRS,
       }
     end
 
@@ -49,6 +61,8 @@ module Mysql2
       @read_timeout = nil
       @query_options = self.class.default_query_options.dup
       @query_options.merge! opts
+      @connect_options = self.class.default_connect_options.dup
+      @connect_options.merge!(opts.slice(*@connect_options.keys))
 
       apply_tls_option_aliases(opts)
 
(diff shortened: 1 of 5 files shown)


jeremy (commented):
Ran this on macOS arm64 against Homebrew MySQL: client specs 211 examples / 0 failures, full suite 630 / 0 (the 25 pending are the usual environment-conditional ones, same as master here). Also verified the claims behind the change directly:

- Nothing reads `:database`/`:dbname`/`:db`/`:default_file`/`:default_group` out of `@query_options` anywhere — `initialize` pulls them from `opts` (`lib/mysql2/client.rb:36` for the default-file pair, `:81` for database), and nothing in ext/ touches those keys either. The deletes remove genuinely dead copies.
- `@query_options` is a `.dup` of the class hash, so the instance-level deletes don't leak into `Client.default_query_options` — confirmed the class hash is intact after connecting.
- The `:dbname`/`:db` deprecation warning still fires, since the deletes land after the warn block in `check_and_clean_query_options`.
- `select_db` doesn't resurrect the key, and `#database` tracks it correctly — matches your new spec, confirmed interactively.

Two things worth a thought, neither blocking:

1. **`Client.default_query_options` still advertises the #493 decoy.** `default_file: nil, default_group: nil` remain in the defaults hash (`lib/mysql2/client.rb:17-18`), and `Client.default_query_options.merge!(default_file: '/nonexistent/my.cnf')` still connects fine, silently ignored — I verified against this branch. The instance-level drop makes the no-op *discoverable after the fact*, but the bait that invited the `merge!` in the first place is those two keys sitting in the defaults listing. Since nothing reads them from `@query_options`, they could come out of the defaults hash too, which would finish the #493 story at the layer it was reported. Your call whether that belongs here or as a follow-up.

2. **Release-notes flag: the `query_options` round-trip loses the database now.** Code doing `Mysql2::Client.new(old_client.query_options.merge(password: ...))` as a reconnect pattern previously landed back in the same database; on this branch it connects with no default DB and fails later with "No database selected" on unqualified queries — a quieter failure mode than the missing password, which at least fails at auth time. AR and Sequel keep their own config hashes so the big adapters are unaffected, but it's the one observable break and deserves a line in the release notes; the migration is `merge(database: old_client.database)`.

The scoping is right: fixing the two reported decoy keys with the `:password` precedent, without dragging in the full `query_options`/`connect_options` split from the #437 thread, keeps this reviewable and keeps the genuinely reactive keys untouched. The spec covering `database: nil` (key absent even when never set) is a nice touch — it pins the invariant, not just the cleanup.

jeremy (commented):
Re-reviewed at 0e5b0c3 (the rework). Rebuilt and reran locally on macOS arm64 / Homebrew MySQL: client spec 212 examples, 0 failures; the four new #437/#493 examples all pass. Probed interactively as well:

- `:database` absent from `query_options` after connect, `#database` reads back correctly, `select_db` switches without resurrecting the key.
- `Client.default_query_options` class hash is untouched by the instance-level deletes.
- Deprecation warn for `:dbname`/`:db` still fires — the deletes land after the warn block.
- The `check_and_clean_query_options` placement is safe: it runs before `connect`, but the connect args and the `default_file=`/`default_group=` setters all read from `opts`, never `@query_options`, and nothing in lib/ or ext/ reads any of the five dropped keys. The client.c hunk is docstring-only.
- The ClassLength disable is earned: without it the cop fires on this branch, with it `rubocop lib/mysql2/client.rb` is clean, and the inline scoping is the right grain.

From my earlier review: the README note covers the `query_options` round-trip flag nicely (confirmed the failure mode is still "No database selected", now documented as intended behavior). The one piece still open is the #493 decoy at the *defaults* layer — `default_file: nil, default_group: nil` still sit in the `Client.default_query_options` listing, and `merge!(default_file: ...)` into it is still silently ignored (re-verified on this branch). Fine as a follow-up; the instance-level drop at least makes the no-op discoverable.

One blocker before merge, and it's not the code: this push produced **no CI at all** — the `[skip ci]` tip suppressed the whole run, so there are zero check suites on 0e5b0c3 *and* on d710c0f. Same pattern as #1581. Needs a CI-triggering push (even an empty amend without the tag) or a manual re-run so a rollup exists on the merge commit's ancestry; happy to approve once one is green.

lib/mysql2/client.rb
```diff
@@ -49,6 +61,8 @@ def initialize(opts = {})
       @read_timeout = nil
       @query_options = self.class.default_query_options.dup
       @query_options.merge! opts
+      @connect_options = self.class.default_connect_options.dup
+      @connect_options.merge!(opts.slice(*@connect_options.keys))
```
jeremy: This is where the old global idiom falls through. `Mysql2::Client.default_query_options[:connect_flags] |= Mysql2::Client::MULTI_STATEMENTS` (peatio et al., ~36 code-search hits) now evaluates `nil | MULTI_STATEMENTS`, which is `true`, not an error -- so `connect_flags: true` sits in `default_query_options`, gets merged into `@query_options` on line 63, and `check_and_clean_query_options` deletes it without it ever reaching `@connect_options`. Repro on 70db47c:

```ruby
Mysql2::Client.default_query_options[:connect_flags] |= Mysql2::Client::MULTI_STATEMENTS
c = Mysql2::Client.new(...)
c.connect_options[:connect_flags] & Mysql2::Client::MULTI_STATEMENTS  # => 0
c.query("SELECT 1; SELECT 2")
# => Mysql2::Error: You have an error in your SQL syntax ... near 'SELECT 2' at line 1
```

An app that works on 0e5b0c3 boots clean here and fails on its first multi-statement query. Two ways out, your call: keep `default_query_options[:connect_flags]` as a fallback base (warn once, point at `default_connect_options`), or make the stale spelling fail loudly at `Client.new` instead of silently. Either way a README/changelog line redirecting the idiom to `Mysql2::Client.default_connect_options[:connect_flags]` -- the README only documents `:flags` today, so there's no existing note to update.

jeremy (changes requested):
The split is the right shape and it closes the #493 decoy properly: `default_file`/`default_group` are gone from the defaults hash rather than deleted after the fact. Flags parity checks out against this head: Array/String/Integer/nil `:flags` and `:sslverify` all resolve into `connect_options[:connect_flags]`, that resolved value is the `flags` local handed to the extension's `connect`, passing `connect_flags:` directly to `Client.new` still overrides the base via the `slice`, and the write-back lands in the per-instance dup (`default_connect_options` stays untouched). Nothing in ext/ or elsewhere in lib/ reads `query_options[:connect_flags]`. client_spec green locally, CI 40/40.

One real regression, which is the only reason this is a request-changes rather than an approve -- see inline. Short version: `Mysql2::Client.default_query_options[:connect_flags] |= MULTI_STATEMENTS` was the global hook for connect flags and it's in the wild (GitHub code search: ~36 hits, peatio and its forks' `config/initializers/mysql2.rb`). On this branch it neither works nor fails, it just drops the flag.

lib/mysql2/client.rb
jeremy: Still open on c6c9476. The whitelist only filters `opts`; `default_query_options.dup` carries the idiom's `connect_flags: true` straight into `@query_options`, nothing routes it to `@connect_options`, and nothing warns:

```
Mysql2::Client.default_query_options[:connect_flags] |= Mysql2::Client::MULTI_STATEMENTS
c = Mysql2::Client.new(...)
c.query_options[:connect_flags]                                   # => true
c.connect_options[:connect_flags] & Mysql2::Client::MULTI_STATEMENTS  # => 0
c.query("SELECT 1; SELECT 2")   # => Mysql2::Error: You have an error in your SQL syntax ... near 'SELECT 2'
```

Same two options as before: read a stray `default_query_options[:connect_flags]` into the connect side with a deprecation warning, or fail loudly on it and point the README at `default_connect_options`.

Rest of c6c9476 checks out: every key the extension reads from `@query_options` (`as async cast cast_booleans symbolize_keys database_timezone application_timezone cache_rows rows_per_gvl_yield stream force_encoding`) is in `SUPPORTED_QUERY_OPTIONS`, per-call `#query` opts still merge unfiltered, the `:sslkey`/`:tls_key` mismatch warning still fires off `@connect_options`, and client/result/statement specs pass locally.

sodabrew:
**Working draft of the fallout/trade-offs/release notes for this PR** -- posting now for visibility while the one open item below gets resolved, not as a final writeup.

## Breaking changes

**`:database` no longer round-trips through `query_options`.** A reconnect pattern like `Mysql2::Client.new(old_client.query_options.merge(password: ...))` used to silently preserve the database; on this branch it connects with no default DB and fails later with "No database selected" on the first unqualified query -- a quieter failure than a missing password (which fails immediately at auth). Migration: `merge(database: old_client.database)`. ActiveRecord and Sequel keep their own config hashes, so the major adapters are unaffected.

**`Client.default_query_options`'s key set changed.** `:connect_flags`, `:default_file`, `:default_group` are no longer present (they never had any effect there -- see below). Anything introspecting the full key set (`.keys`, `.key?`, iteration) will observe the difference. No spec in this repo pins the full key set, so this is a real but likely low-impact API-surface change.

**Open, not yet resolved: `Client.default_query_options[:connect_flags] |= SOME_FLAG` silently breaks.** Caught by @jeremy, with real-world precedent (~36 GitHub code search hits, including peatio and forks, using this exact idiom to set global connect flags). Since `:connect_flags` no longer has a default entry in `default_query_options`, `nil | SOME_FLAG` evaluates to `true` (Ruby's `NilClass#|` coerces to boolean), not an integer -- the flag is silently dropped, not applied, and the failure surfaces far from the cause (e.g. a `MULTI_STATEMENTS`-dependent multi-statement query failing with a generic SQL syntax error). Two ways to close this, not yet decided:
- Keep `default_query_options[:connect_flags]` as a working fallback base (restore its default value so the old idiom stays a valid integer operation), read it at `Client.new` time if it differs from baseline, warn once, and route it into `connect_options` -- backward-compatible, but keeps a legacy escape hatch alive indefinitely.
- Make the stale spelling fail loudly at `Client.new` instead of silently -- cleaner long-term, but is itself a breaking change for the same ~36 call sites, just a loud one instead of a silent one.

Either way, the README currently only documents `:flags`, never `:connect_flags` -- there's no existing doc line to update, only a new one to add pointing at `Client.default_connect_options[:connect_flags]`.

## Trade-offs

**Decoy keys are now closed structurally (whitelist), not by individual patch.** `SUPPORTED_QUERY_OPTIONS`/`SUPPORTED_CONNECT_OPTIONS` replace the old one-key-at-a-time deletion list in `check_and_clean_query_options`. This closes the #493 gap properly (`default_file`/`default_group` are gone from the defaults hash entirely, not just pruned post-hoc) and generalizes the fix beyond the two originally-reported keys -- every connect-time-only option (`host`, `username`, all TLS/SSL settings, `reconnect`, timeouts, ...) gets the same protection now, most of which were never reported as bugs but had the identical decoy shape.

**The whitelist only filters `Client.new`'s `opts` argument -- not the class-level default hashes themselves.** This is exactly the gap the open `connect_flags` item above falls through: `@query_options = self.class.default_query_options.dup` copies the *entire* class hash unfiltered before the whitelist ever applies, so a stray key placed directly onto `Client.default_query_options` (rather than passed to `Client.new`) bypasses the protection entirely. Worth deciding as part of the fix above: filter the class-level `.dup` through the same whitelist too, so this failure mode is closed for any key, not just `:connect_flags` specifically, and not just for this one already-found case.

**Scope: this doesn't migrate every connect-time-only option out of `opts`-only handling.** `SUPPORTED_CONNECT_OPTIONS` is the authoritative list of what's recognized, but most of those keys (host, username, TLS settings, etc.) are still read directly from `opts` in `initialize`, not from `@connect_options` -- only `:connect_flags`, `:connect_timeout`, and `:local_infile` currently get their *resolved* value written back into `@connect_options` for introspection. The rest are present in `connect_options` as whatever was passed in, not necessarily what was actually used (e.g. `:reconnect` isn't coerced the way `send(:reconnect=, !!opts[:reconnect])` coerces it internally). Consistent enough for this PR's purpose, but `connect_options` isn't yet a fully "what was actually used to connect" mirror the way `connect_flags` specifically is.

## Test plan (unchanged from earlier, still accurate)
- `bundle exec rspec` -- full suite passing locally and in CI.
- RuboCop clean.
- Manually verified against a live server: zero decoy keys remain in `query_options` after connecting, `connect_options` is fully populated including resolved `connect_flags`/`connect_timeout`/`local_infile`, password redaction works on the new structure, and the TLS-alias-mismatch warning still fires correctly (this took two tries -- initial ordering had `apply_tls_option_aliases` running before the whitelist split, which silently broke the mismatch detection; two specs caught it).

jeremy:
Weighing in on the open item since it's my finding, after looking at what the 36 hits actually are.

**What the idiom is in the wild.** All 36 are the same line: `Mysql2::Client.default_query_options[:connect_flags] |= Mysql2::Client::MULTI_STATEMENTS` (the peatio family's `config/initializers/mysql2.rb`, ~28 of them) or `|= LOCAL_FILES` (a handful of Rails apps). Zero assignments, zero `merge!`, zero uses of the class hash for `:default_file`/`:default_group`/`:local_infile`, and every hit is an application initializer, not a library. Three things follow: additive is the only semantics to honor; the shim needs to cover `:connect_flags` and nothing else — the class-level filter can strip the rest silently; and because `nil | FLAG` is `true`, the flag bits are unrecoverable once the key is absent, so "detect and raise" is the only option that doesn't keep the key, and "honor" requires keeping an Integer there.

**Pick: warn-and-route, plus filter the class-level `.dup` through the whitelist.**

- The finding was about *silent* breakage. Fail-loud turns 36 silent-wrong into 36 loud-broken at `Client.new` of a 0.5.x release — a second deliberate break in the same PR (after `:database`), landing in app boot. The realistic response is a gem pin, not a one-line fix. Warn-and-route costs about ten lines.
- The steelman for fail-loud is "restoring the key re-creates the decoy this PR exists to remove," and it's wrong in a useful way: a decoy is a key that *looks* live and is inert. Honored and warned, it's a deprecated alias — exactly the category `check_and_clean_query_options` already handles for `:user`/`:pass`/`:hostname`, in the same method. Same policy, applied to a location instead of a spelling. The spec at 166–171 flips from "not present" to "present only as a deprecated alias: mutating it warns once and is honored."
- "Keeps a legacy escape hatch alive indefinitely" — true of every deprecation in this gem (the `:user`/`:pass` warning is a decade old), so it's a choice, not a property of this option. Tag it `# Deprecated alias for default_connect_options[:connect_flags]` with a target version and it's the same promise as the others.

**Shape I'd build:**

- One constant (`DEFAULT_CONNECT_FLAGS`) referenced by both `default_query_options[:connect_flags]` and `default_connect_options[:connect_flags]`, so "did they touch the legacy slot" is an equality against the constant rather than against the other hash.
- `@query_options = default_query_options.select { |k, _| SUPPORTED_QUERY_OPTIONS.include?(k) }` — the instance hash never carries `connect_flags` (spec 174 holds), and any other stray class-level key is dropped silently, matching the documented silent-ignore on the `opts` path.
- In `resolve_connect_flags`, which is already "what base do `:flags` build on": `legacy = self.class.default_query_options[:connect_flags]`; `base = @connect_options[:connect_flags] | legacy` (guarded on `legacy.is_a?(Integer)`, since `Integer | true` raises); warn iff `legacy != DEFAULT_CONNECT_FLAGS`. OR rather than replace: every observed use is additive, OR is the identity when the slot is untouched, and it's the only rule that never loses a flag when an app is half-migrated (old path sets `MULTI_STATEMENTS`, new path `FOUND_ROWS`). Two shapes OR doesn't reproduce, both worth a line in the comment beside it: an assignment through the old location that *clears* a baseline bit (zero in the wild), and an explicit `Client.new(connect_flags: X)` — not `:flags` — which used to override the global and would now combine with it (needs both a mutated global and the undocumented per-connection spelling; also none found).
- Warn once per process, not per `Client.new` — a pool would emit it N times at boot, and a warning that repeats 25 times is the one that gets ignored. The text should name the replacement literally: `Mysql2::Client.default_connect_options[:connect_flags] |= ...`.
- README: `default_connect_options` gets its own paragraph under "Cascading config" beside `default_query_options`, and that's where the `connect_flags` sentence goes. The README never documented `:connect_flags` anywhere — the idiom is copy-paste folklore — so documenting the new location is the actual fix for the next 36.

With that in, the release note's open item closes as "honored with a one-time deprecation warning; move to `default_connect_options`," and the key-set change shrinks to `:default_file`/`:default_group` only.

jeremy:
Weighing in on the open item since the finding was mine. Recommendation: the first option — keep the legacy key as a working fallback base, warn once, route into `connect_options` — plus the class-level `.dup` filter, and schedule the escape hatch rather than leaving it open-ended: warn in this release, raise in the next minor, remove after. Reasoning, and the mechanics that make it hold up:

**Why not fail-loud now.** It's the cleaner hash — it doesn't even need the key restored, since after a `|=` against a missing key the hash holds `true`, and `key?(:connect_flags)` at `Client.new` is enough to raise on. But it delivers the same information as the warning in the form of a boot failure, for every one of those ~36 sites, on a gem bump. The `:database` round-trip break in the notes is a different kind of break: it fails on a pattern that was already subtly wrong. This one would fail on a pattern that worked exactly as documented by the code for a decade. Fail-loud is the right second rung, not the first.

**Mechanics for the fallback, since they're where it goes wrong.**
- Restore `connect_flags:` in `default_query_options` with its original value and a comment saying it's a deprecated fallback base consulted only at `Client.new`. The key has to be present for `|=` to produce an integer — that's the asymmetry with fail-loud.
- "Modified" = value differs from that original constant. When modified, use the legacy value as the *whole* base, not OR it in — master used it as the base, so `&= ~FLAG` worked too, not just `|=`. If both the legacy key and `default_connect_options[:connect_flags]` were modified, the new spelling wins outright and the warning says the legacy value was ignored — the same precedence rule this branch already applies to `:tls_*` over `:ssl*`.
- Warn once per process, not per connection (pools would spam), and with plain `warn`, not `category: :deprecated` — `Warning[:deprecated]` has defaulted to false since 2.7.2, so a categorized warning is invisible in exactly the situation it exists for. `uplevel:` doesn't help either; the stale line ran at boot. The message should name both spellings so it greps.
- The `.dup` filter keeps instances clean: `client.query_options` never shows `:connect_flags` even though the class hash carries it. Your "does not advertise connect-time-only keys in default_query_options" spec then needs one explicit carve-out for `:connect_flags` with the deprecation as the reason.

**Specs that pin it:** legacy `|= MULTI_STATEMENTS` still reaches the server (a two-statement query succeeds); the warning fires once across two `Client.new` calls; both spellings modified → new wins; `&= ~FLAG` honored; a stray key set directly on `Client.default_query_options` is absent from the instance — each with an `ensure` restoring the class hashes.

**README:** one line under Connection options pointing global flags at `Client.default_connect_options[:connect_flags]`, noting the old spelling warns.

With those in, the review item is closed from my side — ping when it's pushed and I'll re-run the idiom probe against the new head.

sodabrew:
Fixed in e1b974a, went with the backward-compatible fallback option.

`:connect_flags` is back in `default_query_options` as a real `Integer` default (shared with `default_connect_options` via a `DEFAULT_CONNECT_FLAGS` constant, so the two can't drift), so `Client.default_query_options[:connect_flags] |= SOME_FLAG` keeps computing a valid bitmask instead of corrupting into `true`. `Client.new` reads it as a fallback (with a one-time warning pointing at `default_connect_options`) only when it differs from the untouched default and the caller didn't pass `connect_flags:` directly -- explicit per-instance always wins.

Also closed the more general gap you pointed at: both class-level defaults hashes now go through the same `SUPPORTED_QUERY_OPTIONS`/`SUPPORTED_CONNECT_OPTIONS` whitelist as `opts`, not just `.dup`, so a stray key set directly on either class hash (bypassing `Client.new` entirely) can't slip through either -- not just for `:connect_flags`.

Reran your exact repro against e1b974a:

```ruby
Mysql2::Client.default_query_options[:connect_flags] |= Mysql2::Client::MULTI_STATEMENTS
c = Mysql2::Client.new(...)
c.connect_options[:connect_flags] & Mysql2::Client::MULTI_STATEMENTS  # => 65536 (nonzero)
c.query("SELECT 1; SELECT 2")  # => succeeds, both statements execute
```

warns once:
```
Client.default_query_options[:connect_flags] is deprecated and will stop being read in a future version; use Client.default_connect_options[:connect_flags] instead
```

README now documents the migration path under "Flags option parsing" -- there was no existing note for the global-default idiom to redirect, since it was never actually documented before this.

Three new specs cover it: the fallback applying with a warning, explicit `connect_flags:` overriding the fallback, and `default_query_options[:connect_flags]` staying a real Integer. Full suite green (646 examples), RuboCop clean.

jeremy:
e1b974a verified against the thread's contract: fallback base restored with the shared `DEFAULT_CONNECT_FLAGS` constant, whole-value use of the legacy base (so flag-clearing works, not just `|=`), explicit `connect_flags:` winning, plain greppable warning, and both class hashes filtered through the whitelists — that last one is better than the `.dup` filter I asked for, since it closes the stray-key path for every key on either hash. README and the spec carve-out read right, CI green at 1c5f28d. Two mechanics from the thread didn't make it in, both small:

1. **The warning fires per `Client.new`, not once per process** — `apply_legacy_connect_flags_fallback` has no guard, so a connection pool emits it on every checkout-connect. One class-level flag (`@legacy_connect_flags_warned ||= ...`) and a spec asserting a single emission across two `new_client` calls.

2. **A modified `default_connect_options[:connect_flags]` loses to the legacy key.** `build_query_and_connect_options` seeds `@connect_options` from `default_connect_options`, then the fallback overwrites it whenever the legacy value differs from baseline — so an app mid-migration (new spelling set, old `|=` line not yet deleted) gets the legacy value silently winning over the spelling we're telling them to use, and the warning text points them at the very option that just lost. Same one-line shape as the explicit-opts guard: also `return if self.class.default_connect_options[:connect_flags] != DEFAULT_CONNECT_FLAGS` — the newer spelling wins outright, matching this branch's own `:tls_*`-over-`:ssl*` precedence — plus a spec with both modified asserting the new one lands.

With those two, this closes out my review.

Issue #1127, opened 2020-07-07
MySQL2 LOCAL INFILE should be disabled by default.


thypon:
MySQL LOCAL INFILE has been leveraged to perform file exfiltration in the [past][1].

Official [documentation][2] advise to disable this capability unless it is explicitly enabled.
```
To avoid LOAD DATA issues, clients should avoid using LOCAL unless proper client-side precautions have been taken.
```

May you consider disabling `LOCAL INFILE` by default in future `mysql2` releases?

Following the affected code that in null-y case will default to `1` (enabled)
```
    case MYSQL_OPT_LOCAL_INFILE:
      intval = (value == Qfalse ? 0 : 1);
      retval = &intval;
      break;
```

[1]: https://w00tsec.blogspot.com/2018/04/abusing-mysql-local-infile-to-read.html
[2]: https://dev.mysql.com/doc/refman/8.0/en/load-data-local-security.html

sodabrew:
Seems reasonable. I'll target this for the next major since it would be a breaking change to current users.

PR #1585, opened 2026-08-18
Disable LOCAL INFILE by default (#1127)

sodabrew:
## Summary

`:local_infile` was only ever set explicitly -- if a caller didn't pass it, mysql2 never called `mysql_options(MYSQL_OPT_LOCAL_INFILE, ...)` at all, leaving the setting at whatever the underlying client library defaults to (typically enabled). That means `LOAD DATA LOCAL INFILE` was reachable by default, which a malicious or compromised server can [abuse to read arbitrary files off the client](https://dev.mysql.com/doc/refman/8.0/en/load-data-local-security.html).

This defaults `:local_infile` to `false` (mirroring how `:connect_timeout` already gets a default filled into `opts` before the option-application loop), unless a caller explicitly opts in with `:local_infile => true`.

## Test plan
- [x] `bundle exec rspec spec/mysql2/client_spec.rb` -- 208 examples, 0 failures (added a spec asserting LOCAL INFILE is rejected when `:local_infile` isn't given at all)
- [x] `bundle exec rspec` (full suite) -- 627 examples, 0 failures
- [x] `bundle exec rubocop lib/mysql2/client.rb spec/mysql2/client_spec.rb` -- no offenses

### DIFF
--- a/spec/mysql2/client_spec.rb
+++ b/spec/mysql2/client_spec.rb
@@ -1081,7 +1081,7 @@ RSpec.describe Mysql2::Client do # rubocop:disable Metrics/BlockLength
       end.to raise_error(Mysql2::Error, /command is not allowed/)
     end
 
-    it "is disabled by default when :local_infile is not given (#1127)" do
+    it "is disabled by default when :local_infile is not given" do
       client = new_client
       expect do
         client.query "LOAD DATA LOCAL INFILE 'spec/test_data' INTO TABLE infileTest"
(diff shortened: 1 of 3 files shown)


spec/mysql2/client_spec.rb
```diff
@@ -1081,6 +1081,13 @@ def run_gc
       end.to raise_error(Mysql2::Error, /command is not allowed/)
     end
 
+    it "is disabled by default when :local_infile is not given (#1127)" do
```
sodabrew:
```suggestion
    it "is disabled by default when :local_infile is not given" do
```

Issue #719, opened 2016-01-07
MySQL time is not compatible with Ruby time


okke-formsma:
Time fields in MySQL may contain times longer than 24 hours. This can not always be parsed by the ruby Time type. See the following 2 examples:

```
#this succeeds
result = client.query(%(
select time('24:00:00') as succeeds
)).each do |row|
    p row
end

#this fails          
result = client.query(%(
select time('24:00:01') as failure
)).each do |row|
    p row
end
```

I'm not sure what the best solution would be. For example, using the Time from DateTime instead of the standard library, or casting to another type if the cast to Time is not working.

sodabrew:
What is a MySQL time value beyond 24 hours supposed to represent?

sodabrew:
Oh, you're right, this could be a problem! http://dev.mysql.com/doc/refman/5.5/en/time.html

> MySQL retrieves and displays TIME values in 'HH:MM:SS' format (or 'HHH:MM:SS' format for large hours values). TIME values may range from '-838:59:59' to '838:59:59'. The hours part may be so large because the TIME type can be used not only to represent a time of day (which must be less than 24 hours), but also elapsed time or a time interval between two events (which may be much greater than 24 hours, or even negative).

sodabrew:
Fixed in #1589: TIME is now represented as a Time offset from the existing placeholder-date anchor, adding the signed duration as an exact Rational seconds offset instead of constructing the Time directly with raw (and potentially out-of-range) hour/min/sec components. Covers the full -838:59:59 to 838:59:59 range, including the binary-protocol sign bug (MYSQL_TIME.neg was never read).

Also spun off #1590: a general `:cast_except` option to exclude specific columns from casting, which came up while considering (and ultimately not needing) a per-value String fallback for this issue.

Issue #1552, opened 2026-08-13
Proposal: mktime() fast path for :local DATETIME decoding — the gem's default timezone


jeremy:
Part of the roadmap in #1510 (round-2 addition).
**Category: low-hanging fruit** · `effort:M · risk:med · ready:now · layer:C`

Sibling of the merged :utc fast path (#1461), covering the timezone that #1461 deliberately left on the slow path.

## Problem

#1461 builds `:utc` DATETIME values via `rb_time_timespec_new` instead of a 7-argument `rb_funcall`, but the gate is hard on `:utc`: `MYSQL2_UTC_FAST_PATH_OK` requires `tz == intern_utc` (`result.c:710-711`). Meanwhile `:database_timezone` **defaults to `:local`** — both the unset case and the invalid-value fallthrough map to `intern_local` (`result.c:1791`, `result.c:1796`).

So the gem's own default configuration decodes every DATETIME cell through `rb_funcall(rb_cTime, :local, 7 boxed args)` (`result.c:1423`). Profiling the rails-shaped benchmark (hash rows, two datetime columns) under the default tz config, the Time cluster is ~4–5% of the whole process: `time_s_mktime` 2.56% tot + `time_arg` 1.76 + `find_time_t` 1.62 + `localtimew` 2.34 + `time_localtime` 2.29 (overlapping). The expensive part is Ruby's `Time.local` doing a `find_time_t` search that calls `localtime_r` repeatedly per cell — work `mktime` does once.

Honest scoping up front: Rails sets `database_timezone = :utc`, so Rails apps already ride #1461. The beneficiaries are non-Rails and legacy-config users — which is nonetheless everyone running the gem's defaults.

## Proposal

A `mysql2_local_time` sibling of `mysql2_utc_time` (`result.c:691-706`):

1. Fill a `struct tm` from the already-parsed components, `tm_isdst = -1`.
2. `mktime()` — one libc call resolving the wall time against the local zone.
3. Round-trip-verify: `localtime_r` the resulting `time_t` and compare components. A mismatch means the wall time was nonexistent (DST spring-forward gap) or otherwise normalized — return Qnil and fall back to the existing funcall, so edge cases keep byte-identical behavior with `Time.local`.
4. On match, `rb_time_timespec_new(&ts, INT_MAX)` — `INT_MAX` is the documented "local time" sentinel, per the binding's own comment from #1461 (`result.c:684-685`).

[...]

PR #1589, opened 2026-08-18
Represent TIME as a duration-anchored Time, not a broken time-of-day (#719)

sodabrew:
## Summary

MySQL's `TIME` is a signed duration (`-838:59:59` to `838:59:59`, per its own docs: "elapsed time or a time interval between two events"), not a time-of-day, but mysql2 built it with `Time.utc/local(..., hour, min, sec, usec)` directly using the raw components. Ruby's `Time` constructor rejects out-of-range components, so:

- Any `TIME` with a 3-digit hour or a negative value silently returned `nil` (the text-protocol `sscanf` couldn't parse a 3-digit hour or a leading `-`).
- Some in-range-looking values (e.g. `hour=24` with a nonzero second) raised `ArgumentError` outright, uncaught, mid-iteration.
- The binary protocol (prepared statements) was worse: it never read `MYSQL_TIME.neg` at all, so a negative `TIME` silently came back **positive** -- wrong data, not just missing data.

All three confirmed live before this fix, against a real `TIME` column and via prepared statements.

## Fix

`TIME` is now represented as a `Time` offset from the existing `2000-01-01` placeholder anchor this code already used, adding the signed duration as an exact `Rational` number of seconds via `Time#+` instead of constructing the `Time` directly with the raw (and potentially out-of-range) components.

`Time#+` operates on the underlying continuous instant, not calendar fields, so this is immune to leap-year layout (a day is always 86400 seconds) and exact to the microsecond (`Rational`, never a float -- no precision loss). Under `:local`, a duration crossing a DST boundary is displayed with whatever wall-clock shift DST implies, which is the same behavior any `Time + seconds` arithmetic in Ruby already has, not something this introduces.

The common in-range case (`0 <= hour < 24`) is unchanged: it produces byte-identical output to today's direct construction.

Both protocols now agree exactly, including the previously-broken sign:

```
-838:59:59  => 1999-11-27 01:00:01
-01:00:00   => 1999-12-31 23:00:00
00:00:00    => 2000-01-01 00:00:00
12:34:56.25 => 2000-01-01 12:34:56.25
24:00:00    => 2000-01-02 00:00:00
24:00:01    => 2000-01-02 00:00:01
838:59:59   => 2000-02-04 22:59:59
```

## Behavior change (release-note candidate)

Text-protocol callers that used to get `nil` for an out-of-range `TIME` (3-digit hour, or negative) now get a real `Time` on a rolled placeholder date instead. This is broken -> correct, not a new regression, but it's an observable change worth a release-note line: any code that was relying on (or silently tolerating) `nil` for these values will now see a `Time` instead.

## Test plan
- [x] Live-verified the original bug (nil/ArgumentError on text protocol, silent sign-flip on binary protocol) against a real `TIME` column before the fix, and the corrected output after.
- [x] New specs in `result_spec.rb` (text protocol, full value matrix) and `statement_spec.rb` (binary-protocol parity against the text protocol, specifically covering the sign bug).
- [x] Removed two now-stale spec comments in `result_spec.rb` that referenced this as a known pre-existing limitation being deliberately avoided.
- [x] `bundle exec rspec` (full suite) -- 628 examples, 0 failures
- [x] `bundle exec rubocop spec/mysql2/result_spec.rb spec/mysql2/statement_spec.rb` -- no offenses

## Review

@jeremy independently re-verified every claim above against fresh builds of master and this branch, including cases beyond the specs (in-range byte-identity via `eql?`, sub-microsecond precision on negative fractional values, `:application_timezone` composition, and that `cast: false`/`:fast` still bypass the new path entirely), plus a perf comparison (no measurable regression -- within noise). Two non-blocking suggestions from that review are addressed: `mysql2_time_from_duration` now has its own `opt_time_day` instead of reusing `opt_time_month` for both month and day, and the behavior-change note above.

### DIFF
--- a/README.md
+++ b/README.md
@@ -717,6 +717,15 @@ Then, if `:application_timezone` is set to say - `:local` - Mysql2 will then con
 
 Both options only allow two values - `:local` or `:utc` - with the exception that `:application_timezone` can be [and defaults to] nil
 
+A `TIME` column is a signed duration in MySQL, not a time-of-day -- it ranges
+from `-838:59:59` to `838:59:59` (elapsed time, or an interval between two
+events). Mysql2 represents it as a `Time` offset from a fixed placeholder
+date (`2000-01-01`, in whichever timezone `:database_timezone` names),
+rolling that date backward or forward as the duration requires: `24:00:01`
+becomes `2000-01-02 00:00:01`, `-01:00:00` becomes `1999-12-31 23:00:00`.
+Values within a single day (`00:00:00` to `23:59:59`) land on the
+placeholder date unchanged.
+
 ### Casting "boolean" columns
 
 You can now tell Mysql2 to cast `tinyint(1)` fields to boolean values in Ruby with the `:cast_booleans` option.
--- a/README.md
+++ b/README.md
@@ -726,6 +726,9 @@ becomes `2000-01-02 00:00:01`, `-01:00:00` becomes `1999-12-31 23:00:00`.
 Values within a single day (`00:00:00` to `23:59:59`) land on the
 placeholder date unchanged.
 
+Both plain `TIME` (whole-second precision) and `TIME(N)` up to `TIME(6)`
+(microsecond precision) are supported.
+
 ### Casting "boolean" columns
 
 You can now tell Mysql2 to cast `tinyint(1)` fields to boolean values in Ruby with the `:cast_booleans` option.
--- a/README.md
+++ b/README.md
@@ -717,14 +717,11 @@ Then, if `:application_timezone` is set to say - `:local` - Mysql2 will then con
 
 Both options only allow two values - `:local` or `:utc` - with the exception that `:application_timezone` can be [and defaults to] nil
 
-A `TIME` column is a signed duration in MySQL, not a time-of-day -- it ranges
-from `-838:59:59` to `838:59:59` (elapsed time, or an interval between two
-events). Mysql2 represents it as a `Time` offset from a fixed placeholder
-date (`2000-01-01`, in whichever timezone `:database_timezone` names),
-rolling that date backward or forward as the duration requires: `24:00:01`
-becomes `2000-01-02 00:00:01`, `-01:00:00` becomes `1999-12-31 23:00:00`.
-Values within a single day (`00:00:00` to `23:59:59`) land on the
-placeholder date unchanged.
+A `TIME` column ranges from `-838:59:59` to `838:59:59`. Mysql2 represents
+it as a `Time` offset from a fixed placeholder date (`2000-01-01`, in
+`:utc` or `:local` per `:database_timezone`), rolling that date backward
+or forward as the duration requires: `24:00:01` becomes
+`2000-01-02 00:00:01`, `-01:00:00` becomes `1999-12-31 23:00:00`.
 
 Both plain `TIME` (whole-second precision) and `TIME(N)` up to `TIME(6)`
 (microsecond precision) are supported.
--- a/ext/mysql2/result.c
+++ b/ext/mysql2/result.c
@@ -889,11 +889,10 @@ static VALUE mysql2_utc_time(unsigned int year, unsigned int month, unsigned int
 /* MySQL TIME is a signed duration of hour, minute, second, and
  * microseconds, ranging from -838:59:59.999999 to 838:59:59.999999, but
  * Ruby's Time constructor rejects an hour beyond 24 or a negative value.
- * We solve this by taking a Time anchored at 2000-01-01 00:00:00 and
- * adding or subtracting the value converted to seconds -- a plain Integer
- * when there are no microseconds (the common case, and measurably cheaper
- * than a Rational), or an exact Rational when there are, so no microsecond
- * is lost to float rounding. */
+ * We solve this by adding the signed duration to a Time anchored at
+ * 2000-01-01 00:00:00. Duration is converted to an Integer in seconds
+ * or a Rational in microseconds, depending on the precision of the
+ * field. */
 static VALUE mysql2_time_from_duration(VALUE db_timezone, int negative,
                                        unsigned int hour, unsigned int min, unsigned int sec,
                                        unsigned long usec) {
--- a/ext/mysql2/result.c
+++ b/ext/mysql2/result.c
@@ -900,10 +900,8 @@ static VALUE mysql2_time_from_duration(VALUE db_timezone, int negative,
   int64_t total_sec = (int64_t)hour * 3600 + (int64_t)min * 60 + (int64_t)sec;
 
   /* :utc has no environment dependence, so the cached anchor is exact.
-   * :local isn't cacheable the same way: Time.local re-reads ENV['TZ'] on
-   * every call, so a cached anchor would freeze the process's timezone at
-   * whatever it was when this extension loaded, silently going wrong after
-   * any later ENV['TZ'] change. */
+   * :local can't be cached the same way: Time.local re-reads ENV['TZ']
+   * on every call. */
   anchor = (db_timezone == intern_utc) ? opt_time_anchor_utc
          : rb_funcall(rb_cTime, intern_local, 7, opt_time_year, opt_time_month, opt_time_day,
                       INT2FIX(0), INT2FIX(0), INT2FIX(0), INT2FIX(0));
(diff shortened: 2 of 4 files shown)


jeremy (commented):
Reviewed with fresh builds of master and this branch against a local MySQL, since the claims here are all empirically checkable. Everything checked out.

**Before-state confirmed on clean master** (222ee2c), text vs. prepared, `database_timezone: :utc`:

```
-838:59:59       text=nil                          stmt=RAISED ArgumentError
-01:00:00        text=nil                          stmt=2000-01-01 01:00:00.000000   <- silent sign flip
24:00:00         text=2000-01-02 00:00:00.000000   stmt=2000-01-02 00:00:00.000000
24:00:01         text=RAISED ArgumentError         stmt=RAISED ArgumentError
838:59:59        text=nil                          stmt=RAISED ArgumentError
```

The `MYSQL_TIME.neg` claim holds: on master the binary fetch path (result.c:1207-1209) constructs from `ts->hour/minute/second/second_part` and never reads `ts->neg` — the only `neg` references in the tree are `t.neg = 0` on the param-*send* side (statement.c:489, 518). The `-01:00:00 => +01:00:00` row above is that bug live: wrong data returned silently, which makes this the strongest motivation of the three failure modes.

**After-state on this branch**: ran the full value matrix (`-838:59:59`, `-01:00:00`, `-00:00:00.5`, `00:00:00`, `12:34:56.25`, `23:59:59.999999`, `24:00:00`, `24:00:01`, `838:59:59`) through both protocols under both `:utc` and `:local`. Every value matches the PR description's table, and text/stmt agree exactly (`eql?`, not just `==`) on every row in both timezones.

Specific claims I probed beyond the specs:

- **In-range byte-identity**: `12:34:56.25` comes back `eql?` to `Time.utc(2000,1,1,12,34,56,250000)` and to `Time.local(...)` respectively — `subsec == (1/4r)` either way, so the Rational-offset construction is indistinguishable from the old direct construction for the values that worked before. The compat story for existing consumers is genuinely "no change".
- **Precision**: `23:59:59.999999` exact through both protocols; `-00:00:00.5` yields `1999-12-31 23:59:59.5` — the negative-with-fraction case the magnitude/sign split has to get right, and does (binary sends `neg=1` with fractional magnitude in `second_part`; text gets the leading `-` stripped before `%3u`).
- **`:application_timezone` interplay**: `838:59:59` with db `:utc` + app `:local` returns the same instant localtime'd (`2000-02-04 14:59:59 -0800` here) — the post-hoc conversion composes correctly with the new construction.
- **`cast: false` / `cast: :fast` untouched**: both still return the raw string (`"-838:59:59.000000"`) on both protocols, so the deferred-cast paths don't route through the new helper.
- **Suite**: `bundle exec rspec` — 628 examples, 0 failures, 25 pending, matching your run.
- **Perf** (since TIME full-cast is now anchor + `rb_rational_new` + `Time#+` instead of one `Time.utc` call): 50x1000 in-range TIME rows, full cast, min-of-10 — master 0.086s vs. branch 0.093s, medians overlapping (0.143 vs. 0.134). Difference is in the noise; the funcall into `Time.utc` dominates either way. No fast-path carve-out warranted.

Design-wise, anchoring the duration on the placeholder date this code already used is the right call — it fixes the sign and range bugs without inventing a new return type, and `Time#+` on the instant sidesteps every calendar-field validity question at once. The README paragraph and the removal of the two now-stale spec breadcrumbs are appreciated; the stmt spec asserting binary/text parity (rather than re-asserting the value table) is exactly the shape that would have caught the `.neg` bug in the first place.

Two non-blocking observations:

- `mysql2_time_from_duration` inherits the `opt_time_year, opt_time_month, opt_time_month` triple from the lines it replaces — day passed as the month constant, which works only because both are 1. It was a two-second stumble on master's two call sites and is now concentrated in the one place a future reader will study. An `opt_time_day` (or a comment) would be a kindness, entirely optional.
- Worth a release-note line at merge time (repo convention keeps CHANGELOG.md pointed at Releases): text-protocol callers that treated an out-of-range TIME as `nil` now receive a real `Time` on a rolled date. That's broken → correct, but it is an observable change.

Verified sign handling, range extremes, sub-second exactness, timezone/app-timezone composition, cast-mode isolation, and in-range compat; came up empty hunting for a failing input.

jeremy:
Checked the delta since my review: abd73d8 is the `opt_time_day` addition and nothing else (+3/−2 in result.c). Dedicated constant in the statics, `INT2NUM(1)` initialized alongside its siblings — a Fixnum immediate like `opt_time_year`/`opt_time_month`, so no `rb_global_variable` needed — and the anchor call now reads year/month/day. Exactly what the observation asked for; no behavior change.

Review stands. LGTM.

jeremy:
Delta since abd73d8: `mysql2_time_from_duration` branches on `usec == 0` — plain Integer seconds on the whole-second path, Rational only when there's a fraction — plus a TIME(4) partial-precision spec.

Checked the equivalence. `Time#+` operates on the instant for Integer exactly as for Rational, so `anchor + total_sec` and `anchor + Rational(total_sec * 10^6, 10^6)` produce the same value under both `:utc` and `:local` anchors; negation lands on the right operand in each branch; 838:59:59 (3,020,399 s / ~3.02e12 µs) is nowhere near int64 or Fixnum limits. The whole-second check holds on both protocols — binary passes `second_part` (already µs, 0 when whole), text leaves `msec_char` at `"000000"` when sscanf matches only three tokens.

Coverage splits cleanly across the branches too: the #719 spec's whole-second values (-838:59:59 through 838:59:59, asserted to %6N) now exercise the Integer path, and the fractional rows — 12:34:56.25 there, plus the new TIME(4) values with the text-vs-binary `eql` — hold the Rational path and the wire-precision expansion. An inverted condition or a mis-anchored fast path fails one side or the other.

LGTM, still.

jeremy:
Delta since 9f191c7: a626b06 caches the two anchors at init, a1ad03a trims comments (+12/−3). CI: a626b06 got its own full green run (40/40); the `[skip ci]` tip is comment-only, so nothing uncovered.

The `:utc` anchor cache is sound, checked: `Time.utc` has no environment dependence, `Time#+` allocates a new object so the shared anchor is never mutated, the anchor itself never escapes to Ruby land, and the assign-then-`rb_global_variable` ordering is safe (statics aren't stack-scanned, and no allocation intervenes). `db_timezone` is only ever `intern_utc` or `intern_local` (result.c:2287–2293 covers unset and invalid too), so the ternary covers the whole domain.

The `:local` anchor cache has a real behavior delta, though: it freezes the process timezone at extension-load time. `Time.local` consults `ENV['TZ']` on every call (tzset semantics), so the old per-call anchor tracked runtime TZ changes; the cached one pins the anchor's *instant* at whatever TZ was active at `require "mysql2"`. And since localtime-mode Times re-render under the current zone, the symptom is a wrong wall clock, not a wrong offset — verified live against this branch's built extension:

```ruby
# TZ=UTC at require time, then ENV['TZ'] = 'America/Chicago' at runtime:
CAST('01:00:00' AS TIME), database_timezone: :local
# => 1999-12-31 19:00:00 -0600   (per-call anchor gives 2000-01-01 01:00:00 -0600)
```

`:local` is the gem's default `database_timezone` (unset and invalid both map to it), and runtime TZ flips are exactly the spec pattern the TZ-matrix style in #1552 relies on, so I'd rather not freeze it silently. DST within a zone is *not* affected — the anchor date is fixed at 2000-01-01 either way — so no ghost to chase there.

Suggestion: keep `opt_time_anchor_utc` cached (that arm keeps the full 2.4x, and it's the arm Rails-configured apps hit) and rebuild the local anchor per call:

```c
anchor = (db_timezone == intern_utc) ? opt_time_anchor_utc \
       : rb_funcall(rb_cTime, intern_local, 7, opt_time_year, opt_time_month, opt_time_day,
                    INT2FIX(0), INT2FIX(0), INT2FIX(0), INT2FIX(0));
```

That drops `opt_time_anchor_local` entirely and restores exact pre-a626b06 behavior for `:local` while keeping the Integer-offset win from 9f191c7 on both arms.

jeremy:
ac87426 resolves it exactly as hoped: :utc keeps the cached anchor, :local rebuilds per call, and the comment pins down why the asymmetry exists. Verified the cache removal is complete (static, init construction, and GC registration all gone) and that ac87426 carries its own full 40/40 green run, with the [skip ci] tips comment/README-only on top. LGTM stands.

Issue #1517, opened 2026-08-13
Proposal: report the stored result's libc-heap copy to Ruby's GC


jeremy:
Part of the roadmap in #1510.
**Category: gap filler** · `effort:M · risk:low · ready:now · layer:C`

## Problem

A buffered query holds the entire result set twice outside Ruby's view. `mysql_store_result` (client.c:858, via `nogvl_store_result` at client.c:869; the stmt analog `mysql_stmt_store_result` at statement.c:597) copies every row into the client library's own heap allocations, and on the stmt path the binding additionally retains `result_buffers` — a bind array plus per-column buffers that grow to largest-value-seen. Ruby's GC sees none of it:

- There is no `rb_gc_adjust_memory_usage` call anywhere in ext/, so `malloc_increase`-driven GC pacing never learns that a Result is pinning tens of MB. A process that holds large results grows RSS silently instead of provoking a major GC.
- `rb_mysql_result_memsize` (result.c:235-245) reports three struct `sizeof`s and nothing else, so `ObjectSpace.memsize_of(result)` and heap dumps understate by the entire payload — the number a person debugging memory actually needs is the one that's missing.

## Proposal

Two independent fixes, same PR:

1. **GC pacing**: at store time, estimate the client-library allocation (row count × per-row pointer/length overhead + data lengths — `mysql_fetch_lengths` is available per row on a stored result, summed directly or sampled log-scaled as pg does), then `rb_gc_adjust_memory_usage(+estimate)`. On `free_result`, `rb_gc_adjust_memory_usage(-estimate)`.
2. **Honest memsize**: add the retained `result_buffers` buffer_length sums and the stored estimate to `rb_mysql_result_memsize`, so heap dumps stop lying.

Bookkeeping discipline: the estimate is stored on the result wrapper and the `-adjust` reads it back, so plus and minus balance exactly no matter how the estimator evolves — drift between paired calls is the one real correctness hazard, and this closes it by construction. Streaming results are skipped (rows aren't client-buffered; there's nothing material to report). Guarded by `have_func("rb_gc_adjust_memory_usage")` in extconf.rb (Ruby 2.4+).

## Evidence and precedent

[...]

PR #1597, opened 2026-08-19
Report the stored result's client-heap copy to Ruby's GC

jeremy:
A buffered query holds its entire result set in the client library's own heap — `mysql_store_result` on the text path, `mysql_stmt_store_result` on the binary path — where Ruby's GC never sees it. Malloc-driven GC pacing never learns that a Result is pinning tens of MB, so a process holding large results grows its RSS unbounded without ever provoking a collection, and `ObjectSpace.memsize_of` reports a few hundred bytes for a Result pinning megabytes.

Proposed in #1517.

## Approach

**GC pacing.** Estimate the client-side copy when the Result adopts it and report it with `rb_gc_adjust_memory_usage(+N)`. The estimate is stashed on the wrapper and handed back, negated, when the underlying result is freed, so the paired adjustments balance by construction however the estimator evolves. Streaming results are exempt: their rows are never client-buffered. pg has shipped this mechanism since 2017 (`pg_result.c`), including its boundary lesson: only memory Ruby cannot otherwise see is reported — the statement bind buffers are xmalloc'd and already flow through Ruby's allocator, so they are deliberately not part of the adjustment.

The estimator differs per protocol because the client libraries record different things:

- **Text**: `mysql_fetch_lengths` over a log-scaled sample of rows (16 per power of two of row count; a million-row result probes ~320), extrapolated to the full row count and added to the per-row node/pointer-array/NUL overheads both client libraries allocate. The pass advances the stored cursor with `mysql_fetch_row` — local pointer chasing, no I/O, no error path — and rewinds with `mysql_data_seek`. (libmysqlclient leaves `MYSQL_ROWS.length` unpopulated for text rows, so the exact walk below isn't available here.)
- **Binary**: walk the statement's `MYSQL_ROWS` list, whose per-row packet length both libmysqlclient (`add_binary_row`) and MariaDB Connector/C (`mthd_stmt_read_all_rows`) record as they buffer it — exact, no sampling. `MYSQL_STMT` and its `result` member are public in both clients' headers; the free path already reaches into the struct the same way for `stmt->bind_result_done`.

Both adjustments are safe from every caller: `rb_gc_adjust_memory_usage` feeds `malloc_increase` as a `MEMOP_TYPE_REALLOC`, which never starts a collection itself (the next real malloc paces off it), so the negative side may run from the GC dfree path, exactly as pg's does. Guarded by `have_func("rb_gc_adjust_memory_usage")` (Ruby 2.4+).

**Honest memsize.** `rb_mysql_result_memsize` gains the stored estimate and the statement result bind buffers (the `MYSQL_BIND` array, its side arrays, and each field's value buffer), so heap dumps stop understating a Result by its entire payload. Accuracy in practice: 1,045,576 reported for 512 rows × 2,000 bytes (+2.1%) on the sampled text path; 1,038,976 (+1.5%) on the exact binary path.

## Evidence

Held-result workload — 40 × 10MB stored results dropped unreferenced (no explicit `#free`), light allocation churn, 7 interleaved runs per arm, medians (ranges), arm64 macOS, Ruby 4.0.6, MySQL 9.6, base 222ee2c:

|  | base | branch |
|---|---|---|
| GC runs | 0 (0..0) | 13 (13..13) |
| … major | 0 (0..0) | 4 (4..4) |
| peak RSS | 283MB (190..467) | 120MB (119..126) |

**The flip side is the point**: workloads that hold big buffered results now GC when they should, where before they silently grew until an OOM. In allocation-heavy microbenchmarks that hold large results, this can read as a throughput loss — that's the corrected pacing, not a regression.

Store-time cost of the estimator is noise: with GC disabled to isolate it, query+free of a 50,000-row result overlaps across arms (branch medians 26–38ms vs base 24–43ms across three interleaved runs).

## Specs

Eight new examples: memsize covers the stored copy (single-row and multi-row/extrapolated, with upper-bound sanity), memsize returns to baseline on free and on materialization auto-free, statement bind buffers counted, `malloc_increase_bytes` adjusts up at store and back down at free, streaming exempt, and no drift across repeated store/free cycles nor across statement executes. The four assertions that pin new behavior fail against base exactly as expected (memsize ≈ 450 bytes vs ≥ payload; `malloc_increase_bytes` delta 1,592 vs ≥ 1,000,000); the other four pass on base by design as regression guards.

### DIFF
--- a/ext/mysql2/result.c
+++ b/ext/mysql2/result.c
@@ -263,6 +263,20 @@ static void rb_mysql_result_free_result(mysql2_result_wrapper * wrapper, int fro
     } else {
       mysql_free_result(wrapper->result);
     }
+
+#ifdef HAVE_RB_GC_ADJUST_MEMORY_USAGE
+    /* Hand back exactly the bytes reported to the GC at store time --
+     * reading the stashed value keeps the negative adjustment the mirror
+     * image of the positive one, whatever the estimator said. Safe from the
+     * GC dfree path: a negative adjustment only decrements the GC's malloc
+     * bookkeeping (atomically, floored at zero) and can never itself start
+     * a collection. */
+    if (wrapper->reported_size > 0) {
+      rb_gc_adjust_memory_usage(-(ssize_t)wrapper->reported_size);
+      wrapper->reported_size = 0;
+    }
+#endif
+
     wrapper->resultFreed = 1;
   }
 }
@@ -301,6 +315,19 @@ static void rb_mysql_result_free(void *ptr) {
 static size_t rb_mysql_result_memsize(const void * wrapper) {
   const mysql2_result_wrapper * w = wrapper;
   size_t memsize = sizeof(*w);
+  /* The client library's heap copy of the stored result: zero while
+   * streaming and once the underlying result has been freed. */
+  memsize += w->reported_size;
+  /* Statement result bind buffers: the per-field MYSQL_BIND array, its
+   * is_null/error/length side arrays, and each field's value buffer. */
+  if (w->result_buffers) {
+    unsigned int i;
+    memsize += (size_t)w->numberOfFields
+             * (sizeof(MYSQL_BIND) + 2 * sizeof(my_bool) + sizeof(unsigned long));
+    for (i = 0; i < w->numberOfFields; i++) {
+      memsize += w->result_buffers[i].buffer_length;
+    }
+  }
   if (w->stmt_wrapper) {
     memsize += sizeof(*w->stmt_wrapper);
   }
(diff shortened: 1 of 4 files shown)


sodabrew:
This is really neat!

ext/mysql2/result.c
```diff
@@ -2579,6 +2691,24 @@ VALUE rb_mysql_result_to_obj(VALUE client, VALUE encoding, VALUE options, MYSQL_
...
+   * Ruby's GC, which otherwise never sees it: malloc-driven GC pacing would
+   * let a process holding large results grow its RSS unbounded without ever
+   * provoking a major collection. The estimate is stashed on the wrapper
+   * and handed back, negated, when the underlying result is freed
+   * (rb_mysql_result_free_result), so the two adjustments balance exactly.
+   * Streaming results are exempt: their rows are never client-buffered.
+   * Estimation and adjustment cannot raise and cannot themselves trigger a
+   * collection, keeping this function's no-raise contract. */
+  if (!wrapper->is_streaming) {
+    wrapper->reported_size = wrapper->stmt_wrapper != NULL
```
sodabrew: I think this will read better as a block if rather than a ternary. It took me two reads to realize we're just checking if this is a prepared statement or an ordinary query.


```suggestion
  if (!wrapper->is_streaming) {
    if (wrapper->stmt_wrapper) {
      wrapper->reported_size = rb_mysql_stmt_estimate_heap_size(wrapper->stmt_wrapper->stmt, mysql_num_fields(r));
    } else {
      wrapper->reported_size = rb_mysql_result_estimate_heap_size(r);
    }
  }
```

ext/mysql2/result.c
```diff
@@ -262,6 +262,20 @@ static void rb_mysql_result_free_result(mysql2_result_wrapper * wrapper, int fro
...
     }
+
+    /* Hand back exactly the bytes reported to the GC at store time --
+     * reading the stashed value keeps the negative adjustment the mirror
+     * image of the positive one, whatever the estimator said. Safe from the
+     * GC dfree path: a negative adjustment only decrements the GC's malloc
+     * bookkeeping (atomically, floored at zero) and can never itself start
+     * a collection. */
+    if (wrapper->reported_size > 0) {
+#ifdef HAVE_RB_GC_ADJUST_MEMORY_USAGE
```
sodabrew: This could wrap the entire if block, since we don't do anything with the values if we don't have the function to call.

ext/mysql2/result.c
```diff
@@ -2579,6 +2691,24 @@ VALUE rb_mysql_result_to_obj(VALUE client, VALUE encoding, VALUE options, MYSQL_
...
+   * and handed back, negated, when the underlying result is freed
+   * (rb_mysql_result_free_result), so the two adjustments balance exactly.
+   * Streaming results are exempt: their rows are never client-buffered.
+   * Estimation and adjustment cannot raise and cannot themselves trigger a
+   * collection, keeping this function's no-raise contract. */
+  if (!wrapper->is_streaming) {
+    wrapper->reported_size = wrapper->stmt_wrapper != NULL
+      ? rb_mysql_stmt_estimate_heap_size(wrapper->stmt_wrapper->stmt, mysql_num_fields(r))
+      : rb_mysql_result_estimate_heap_size(r);
+#ifdef HAVE_RB_GC_ADJUST_MEMORY_USAGE
```
sodabrew: This ifdef could wrap the entire if block, since we don't do anything with the computed value if we don't have the gc function to call.

ext/mysql2/result.c
jeremy: Done in b119994 — took your block-if shape verbatim, with the adjust call folded in now that the whole block sits under the ifdef (next thread). Thanks for the close read!

ext/mysql2/result.c
jeremy: Done in b119994 — the ifdef wraps the whole block on the free path now.

ext/mysql2/result.c
jeremy: Done in b119994 — the ifdef wraps the whole store-time block, plus the two estimator functions themselves (call sites gone means -Wunused-function otherwise; verified the guardless path compiles warning-clean). One knock-on worth noting: reported_size also feeds rb_mysql_result_memsize, so builds without rb_gc_adjust_memory_usage (pre-2.4 CRuby) now skip the estimate in ObjectSpace.memsize_of too — a fair trade, since the sampling walk is pure overhead when there is no GC to inform.

Issue #1602, opened 2026-08-25
Proposal: one-shot async-result delegation for Mysql2::EM / em-synchrony


jeremy:
Part of roadmap #1510 (event-driven addition).

Category: gap filler  
Facets: `effort:M · risk:med · ready:design-first · layer:lib+C`

## Problem

Mysql2's Fiber ownership model from #1284 prevents unrelated Fibers from operating on the same client. Cleanup work in #1508 and guard expansion in #1563 reinforce that model.

The shipped EventMachine adapter has one intentional cross-Fiber handoff, however:

1. the application Fiber starts an asynchronous query;
2. the EventMachine watcher later calls `async_result` from the reactor Fiber;
3. em-synchrony resumes the initiating Fiber with that result.

That watcher is completing the existing query, not issuing a second command. Requiring it to be the initiating Fiber breaks this adapter topology; allowing any Fiber to call `Client#async_result` weakens ownership too broadly.

## Proposal

Provide an opt-in, opaque, one-shot result receiver bound to the particular asynchronous query:

```ruby
receiver = client.query(sql, async: :delegate) # illustrative spelling
result = receiver.async_result
```

`Mysql2::EM::Client` would pass that receiver to its watcher:

```ruby
receiver = super(sql, opts.merge(async: :delegate))
@watch = EM.watch(socket, Watcher, receiver, deferrable)
```

The existing watcher protocol remains `@client.async_result`, so em-synchrony can keep treating the watched object structurally.

Required invariants:

- The receiver is bound to one client and one query generation.
- It may be redeemed once, from another Fiber on the same Ruby Thread.
- It can only complete that query's first result; it grants no general client authority.
- Wrong-client, stale, closed, or replayed receivers fail before socket access.
- Completion, server error, interruption, or close revokes the receiver.
- The initiating Fiber remains the client's active owner until completion.
- Existing `async: true` behavior and the current owner-Fiber path remain compatible.
- Cross-thread delegation is out of scope.

The public API spelling is negotiable. An adapter-private receiver constructor would also work; the important part is explicit, per-query, one-shot authority rather than a broad exception in `Client#async_result`.

## Adjacent work kept separate

1. **Statement guard PR #1603:** an independent one-command ownership change. It should contain no EventMachine or delegation API work.
2. **Compatibility-spec PR #1604:** tests only, documenting the same-thread/different-Fiber Mysql2::EM/em-synchrony handoff. It should not change production behavior.
[...]

PR #1603, opened 2026-08-25
Guard prepared-statement operations against concurrent client use

jeremy:
## Summary

- claim the client across statement prepare, execute, and close operations
- reject client disposal, deferred statement reaping, and native `Statement`-handle access while a command or streaming result owns the connection
- reserve the internal `QUERYING` state for the text-query / `async_result` lifecycle and mark prepared-statement and next-result handle phases as `COMMAND`
- release claims after completed replies and validation errors, while invalidating a connection after an interruption leaves protocol completion uncertain
- keep encoded bind bytes in call-owned memory through the native `mysql_stmt_execute` call

There are no new Ruby methods or signature changes; concurrent misuse now fails fast. The existing EventMachine cross-Fiber completion path remains supported. Result streaming/lifetime behavior is unchanged and remains separate work.

## Why

A client connection can process one protocol command at a time. `Client#query` and the command methods covered by #1563 already use the active-Fiber claim to reject overlap; prepared-statement operations did not consistently participate in it. #1508 also called out statement execution's missing unwind protection.

This extends the existing fail-fast connection-ownership model to prepared-statement operations. Explicit one-shot async completion delegation remains a separate design discussion in #1602.

## Tests

- full suite: 649 examples, 0 failures, 33 environment-only pendings
- focused ownership, interruption, completed-server-error reuse, bind-buffer lifetime, callback-teardown, and live-stream reaping coverage
- the separate test-only EventMachine compatibility branch passed against this build: 8 examples, 0 failures, including application-Fiber to reactor-Fiber result completion
- the GC-stress/auto-compaction example is skipped only on Ruby 3.0 for Windows, whose VM hits [Ruby #19320](https://bugs.ruby-lang.org/issues/19320) under that combination; it passes on Linux Ruby 3.0 and Windows Ruby 3.4
- `bundle exec rubocop spec/mysql2/client_spec.rb spec/mysql2/statement_spec.rb`

### DIFF
--- a/ext/mysql2/client.c
+++ b/ext/mysql2/client.c
@@ -908,10 +908,11 @@ static void rb_mysql_client_set_active_fiber(VALUE self, bool closing) {
     // mark this connection active
     wrapper->active_fiber = fiber_current;
   } else if (wrapper->active_fiber == fiber_current) {
-    /* Closing an async text query from its owner is supported. A generic
-     * command may be inside a user conversion callback, though, and closing
-     * it here would free the MYSQL/MYSQL_STMT still on that same C stack. */
-    if (!closing || wrapper->state == MYSQL2_CLIENT_COMMAND) {
+    /* Closing an async text query from its owner is supported. Every other
+     * claimed phase may be inside a user callback -- including prepared
+     * result construction after a streaming cursor has opened -- so closing
+     * there would free the MYSQL/MYSQL_STMT still on that same C stack. */
+    if (!closing || wrapper->state != MYSQL2_CLIENT_QUERYING) {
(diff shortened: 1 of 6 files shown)


jeremy:
CI note: the RuboCop job is failing on three pre-existing `Layout/ExtraSpacing` offenses in `ext/mysql2/extconf.rb:301,315,321`; this PR does not touch that file. The last green `master` run resolved RuboCop 1.89.0, while current PR jobs resolve 1.90.0, whose [#15567](https://github.com/rubocop/rubocop/pull/15567) fixed false negatives in that cop. I have kept the unrelated formatting cleanup out of this PR.

Issue #456, opened 2013-11-30
Sometimes dangling connections test  fails


yui-knk:
Sometimes, on  `client_spec.rb Line.120`  test fails.

https://travis-ci.org/brianmario/mysql2/jobs/14707577
https://travis-ci.org/sodabrew/mysql2/jobs/14707571

Always this test is pass. But sometimes fails.
May be query `SHOW STATUS LIKE 'Threads_connected'` is not correctly much number of Ruby objects.

PR #1604, opened 2026-08-25
Test EventMachine async completion across Fibers

jeremy:
## Summary

- make the Fiber context used by ordinary EventMachine query completion explicit
- cover the Synchrony-style handoff where an application Fiber issues a query, yields, and is resumed after the reactor Fiber consumes the result
- keep the coverage self-contained without adding an em-synchrony dependency

## Why

The README points asynchronous Active Record users to em-synchrony. That integration issues a query from an application Fiber, then relies on EventMachine's watcher to collect the result on the reactor Fiber before resuming the application Fiber.

The existing EventMachine specs cover ordinary callbacks, where query issue and result consumption both happen on the reactor Fiber. This adds the complementary cross-Fiber compatibility case so ownership changes do not inadvertently break the README-documented em-synchrony topology.

This PR changes no production behavior. It is independent of the prepared-statement guard in #1603 and of the explicit delegation API design discussion in #1602.

## Tests

- `bundle exec rspec spec/em/em_spec.rb`
- `bundle exec rubocop spec/em/em_spec.rb`

### DIFF
--- a/ext/mysql2/extconf.rb
+++ b/ext/mysql2/extconf.rb
@@ -298,7 +298,7 @@ wishlist = [
 ]
 
 usable_flags = wishlist.select do |flag|
-  try_link('int main() {return 0;}',  "-Werror #{flag}")
+  try_link('int main() {return 0;}', "-Werror #{flag}")
 end
 
 $CFLAGS << ' ' << usable_flags.join(' ')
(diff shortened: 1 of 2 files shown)


jeremy:
CI note: the RuboCop job is failing on three pre-existing `Layout/ExtraSpacing` offenses in `ext/mysql2/extconf.rb:301,315,321`; this PR does not touch that file. The last green `master` run resolved RuboCop 1.89.0, while current PR jobs resolve 1.90.0, whose [#15567](https://github.com/rubocop/rubocop/pull/15567) fixed false negatives in that cop.

The failed Linux matrix jobs are the existing nondeterministic dangling-connection GC example discussed in #456; both new EventMachine examples passed in those jobs. The Windows 2025 failure occurred in MSYS2 package setup on an invalid PGP signature, before this branch built.

